PR #2, opened 2021-12-05
[canonical-facts] Display output as pretty formatted JSON

strider:
This patch outputs the canonical facts with an appropriate indentation
for better readability.

Expected results:

```
$ rhc canonical-facts
{
    "insights_id": "08eca62e-ee30-49c6-a96c-d8bf59a570e5",
    "machine_id": "e64a070b-fdb9-139e-2586-f6ee138068b5",
    "bios_uuid": "2a362d74-d64e-4be5-aaf8-daa758d052f4",
    "subscription_manager_id": "a5c5f1e1-c117-4a55-93c3-bc361353e9b6",
    "ip_addresses": [
        "192.168.0.28"
    ],
    "mac_addresses": [
        "52:54:00:78:d4:02",
        "00:00:00:00:00:00"
    ],
    "fqdn": "rhel84.strider.local"
}
```

Signed-off-by: Gael Chamoulaud (Strider) <gchamoul@redhat.com>

PR #5, opened 2021-12-06
Fix rendering bug where the spinner doesn't erase its loading message

strider:
Summary

Downgrade the spinner library to fix a rendering bug. See briandowns/spinner#123.

With spinner v1.16.0:
image

With spinner v.1.13.0:
image

Signed-off-by: Gael Chamoulaud (Strider) <gchamoul@redhat.com>

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,7 +4,7 @@ go 1.16
 
 require (
 	git.sr.ht/~spc/go-log v0.0.0-20210611184941-ce2f05edb627
-	github.com/briandowns/spinner v1.16.0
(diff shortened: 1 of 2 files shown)


strider:
FYI @subpop

strider:
https://github.com/RedHatInsights/rhc/pull/2 should be merged first!

subpop:
Funny. It looks like when I split all this code out from RedHatInsights/yggdrasil, I did not take care to copy the go.mod correctly. `go` rebuilt it, using the newer package versions of the dependencies that it detected, so it pulled down spinner-0.16.0.

PR #7, opened 2021-12-08
Include insights-client during "connect" and "disconnect"

subpop:
Introduce `insights-client` as an explicit step during the `connect` and `disconnect` subcommands. During `connect`, the process shells out to, and waits for, `/usr/bin/insights-client --register` to execute successfully. Conversely, during `disconnect`, the process shells out to and waits for `/usr/bin/insights-client --unregister` to execute successfully. In addition, the `insights-client --status` exit code is checking during the `status` subcommand execution.

### DIFF
--- a/insights.go
+++ b/insights.go
@@ -1,0 +1,25 @@
+package main
+
+import (
+	"os/exec"
+)
+
+func registerInsights() error {
+	cmd := exec.Command("/usr/bin/insights-client", "--register")
+
+	return cmd.Run()
+}
+
+func unregisterInsights() error {
+	cmd := exec.Command("/usr/bin/insights-client", "--unregister")
+
+	return cmd.Run()
+}
+
+func insightsIsRegistered() bool {
+	cmd := exec.Command("/usr/bin/insights-client", "--status")
+
+	_ = cmd.Run()
+
+	return cmd.ProcessState.Success()
+}
--- a/main.go
+++ b/main.go
@@ -82,7 +82,7 @@ func main() {
 			},
 			Usage:       "Connects the system to " + Provider,
 			UsageText:   fmt.Sprintf("%v connect [command options]", app.Name),
-			Description: fmt.Sprintf("The connect command connects the system to Red Hat Subscription Management and %v and activates the %v daemon that enables %v to interact with the system. For details visit: https://red.ht/connector", Provider, BrandName, Provider),
+			Description: fmt.Sprintf("The connect command connects the system to Red Hat Subscription Management, Red Hat Insights and %v and activates the %v daemon that enables %v to interact with the system. For details visit: https://red.ht/connector", Provider, BrandName, Provider),
 			Action: func(c *cli.Context) error {
 				hostname, err := os.Hostname()
 				if err != nil {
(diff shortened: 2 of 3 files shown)


subpop:
@strider, do you mind giving me a sanity check on this before I merge it?

strider:
It is forcing the customer to register their machine on the console.redhat.com as well. I think this should be optional for the customers who don't want to use insights. Wdyt?

strider:
Otherwise, it looks gods to me. I will test it later in the night.

subpop:
This is intentional. The vision for `rhc` is to be an "enable all the things in one step" activation utility; this includes `insights-client`. For customers who do not want to use insights, they still have the option of using `subscription-manager` and, if they want to use the console.redhat.com services (such as config-manager and direct-attached remediations), can simply activate `yggd` directly with systemctl (`systemctl enable --now rhcd`).

Note the vision for `rhc` and `yggdrasil` differ in an important way: `rhc` is a front-end to `com.redhat.RHSM1` and other Red Hat specific services. `yggdrasil` is a general-purpose remote-messaging service that happens to be a component of "Red Hat connector" (the product).

The reason I forked `rhc` out of `yggdrasil` and its a standalone project is to support this exact vision, without unintentionally affecting the `yggdrasil` project.

strider:
Ack it makes sense now and it is clearer on the vision. However, this new feature won't appear in the `--help` and in the man page. I think it should be mentioned there. 

Moreover I think this kind of information you gave me in comment should be mentioned as well in the readme file.

strider:
image

image

image

image

It works pretty well, let's merge it! 👍

strider:
@subpop just noticed the description of the **ESSNTL-1634** Jira card. And it is written:

"`/usr/bin/rhc` needs to be modified to execute `insights-client --register` after it activates the `rhcd systemd` unit when executing the `connect` command. Conversely, `/usr/bin/rhc` needs to be modified to execute `insights-client --unregister` before it deactivates the `rhcd systemd` unit when executing the `disconnect` command." 

Is this still accurate? because your patch doesn't seem to respect that description.

subpop:
> Is this still accurate? because your patch doesn't seem to respect that description.

Are you observing that it runs things in this order instead?

1. Register RHSM
2. Register Insights
3. Activate rhcd

1. Deactivate rhcd
2. Unregister Insights
3. Unregister RHSM

strider:
> > Is this still accurate? because your patch doesn't seem to respect that description.
> 
> Are you observing that it runs things in this order instead?
> 
>     1. Register RHSM
> 
>     2. Register Insights
> 
>     3. Activate rhcd
> 
>     4. Deactivate rhcd
> 
>     5. Unregister Insights
> 
>     6. Unregister RHSM

According to the screenshots I've put in a previous comment, yes.

subpop:
> > ```
> > 
> > ```
> 
> 
> According to the screenshots I've put in a previous comment, yes.

I view "activating the rhc daemon" as the last thing that `rhc` should do, so any other operations it should perform should happen before that. That's why the JIRA description differs from the actual implementation.

strider:
Feel free to merge it then.

subpop:
> this new feature won't appear in the `--help` and in the man page. I think it should be mentioned there.

Good point. I'll update the description strings to mention Red Hat Insights too.

PR #9, opened 2022-02-10
Bundle yggdrasil and yggdrasil-worker-package-manager

subpop:
As part of the decoupling of `rhc` and `yggdrasil`, we need to stitch things back together when building the `rhc` RPM. This PR modifies the spec file that is built in COPR to include `yggdrasil` and `yggdrasil-worker-package-manager` as secondary sources, builds them and includes them in the resulting RPM.

### DIFF
--- a/.copr/Makefile
+++ b/.copr/Makefile
@@ -1,5 +1,8 @@
 include Makefile
 
+outdir ?= $(PWD)/outdir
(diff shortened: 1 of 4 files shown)


strider:
I've just rebased your PR on top of main.

PR #10, opened 2022-02-14
Collect error messages during disconnect and log them and the end

subpop:
If an error occurs during disconnect, the error is collected and reported at the end of the operation instead of aborting the operation part-way through.

Fixes: ESSNTL-2281


main.go
```diff
@@ -180,29 +183,49 @@ func main() {
...
-				s.Stop()
-				fmt.Print(failPrefix + " Disconnected from Red Hat Insights\n")
 
 				s.Suffix = " Disconnecting from Red Hat Subscription Management..."
 				s.Start()
 				if err := unregister(); err != nil {
-					return cli.Exit(err, 1)
+					errorMessages["rhsm"] = fmt.Errorf("cannot disconnect from Red Hat Subscription Management: %w", err)
+					s.Stop()
+					fmt.Printf(failPrefix + " Cannot disconnect from Red Hat Subscription Management\n")
```
strider, line 210: Here it should be:

`fmt.Printf(errorPrefix + " Cannot disconnect from Red Hat Subscription Management\n")`

strider (changes requested):
Looks good overall and works pretty well.

image

main.go
subpop: Good catch. Thank you!

PR #11, opened 2022-02-14
feat: report duration of steps during operations

subpop:
The duration of a step (RHSM, insights-client, rhcd) is now timed and printed if the log-level is >= debug.


strider (changes requested):
Instead of having this:

image

It might be better to use `text/tabwriter` here as well:

```[cloud-user@test-client-test-vm-rhel90-gchamoul rhc]$ sudo ./rhc --log-level trace connect
Connecting test-client-test-vm-rhel90-gchamoul to Red Hat.

Username: rh-gs-gchamoul
Password:

● Activated the rhc daemon

STEP     DURATION
rhsm     18.988s
insights 24.735s
rhc      17ms
```

subpop:
This is funny. That's how I originally had it, and changed it to print out the duration after performing each step. I do like the `text/tabwriter` approach more too. I'll rework this patch to use a tabwriter.

strider (approved):
Perfect!

PR #15, opened 2022-09-19
Bug fix: check result of running command

jirihnidek:
* When insights-client is not installed, then running command is finished with `*ExitError` and it is not possible to check state of the process

### DIFF
--- a/insights.go
+++ b/insights.go
@@ -16,10 +16,20 @@ func unregisterInsights() error {
 	return cmd.Run()
 }
 
-func insightsIsRegistered() bool {
+func insightsIsRegistered() (bool, error) {
 	cmd := exec.Command("/usr/bin/insights-client", "--status")
 
-	_ = cmd.Run()
+	err := cmd.Run()
 
-	return cmd.ProcessState.Success()
+	if err != nil {
+		// When the error is ExitError, then we know that insights-client only returned
+		// some error code not equal to zero. We do not care about error number.
+		if _, ok := err.(*exec.ExitError); ok {
+			return false, nil
+		} else {
+			return false, err
+		}
+	}
+
+	return cmd.ProcessState.Success(), err
 }
--- a/main.go
+++ b/main.go
@@ -17,9 +17,13 @@ import (
 	"golang.org/x/term"
 )
 
-const successPrefix = "\033[32m●\033[0m"
-const failPrefix = "\033[31m●\033[0m"
-const errorPrefix = "\033[31m!\033[0m"
+const redColor = "\u001B[31m"
+const greenColor = "\u001B[32m"
+const endColor = "\u001B[0m"
+
+const successPrefix = greenColor + "●" + endColor
+const failPrefix = redColor + "●" + endColor
+const errorPrefix = redColor + "!" + endColor
 
 func main() {
 	app := cli.NewApp()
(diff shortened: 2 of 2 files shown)


insights.go
```diff
@@ -19,7 +19,11 @@ func unregisterInsights() error {
 func insightsIsRegistered() bool {
 	cmd := exec.Command("/usr/bin/insights-client", "--status")
 
-	_ = cmd.Run()
+	res := cmd.Run()
 
-	return cmd.ProcessState.Success()
+	if res == nil {
```
subpop, line 24: Could we log the error if the returned error from `Run` is not nil? Something like:

```go
if err := cmd.Run(); err != nil {
    log.Errorf("cannot execute insights-client: %v", err)
}
return cmd.ProcessState.Success()
```

subpop (changes requested):

insights.go
jirihnidek: Sure, good idea.

jirihnidek: Hmm, when it is not possible to execute insights-client, then the log is printed to `stderr`. Is it expected behavior? Shouldn't `rhc` write logs to some file in e.g. `/var/log/`?

subpop: Yea, I'm not sure if printing the error would clobber the pretty-printed status output. There is no logging-to-a-file mechanism at all in `rhc` today, so that would have to be added. The program isn't really meant to be a logging-to-a-file type program. It's a front-end wrapper around the registration utilities in RHEL.

If `insights-client` is not present in the path, I think that should be considered a fatal error. The application has a hard execution dependency on `insights-client`.

jirihnidek: Yeah, printing the error clobber the pretty-printed status output using your proposal.

When `insights-client` is not present, then it is some corner case (e.g. you run built `rhc` on Fedora, where insights-client is not available), but I think that `rhc` should recover from such situation gracefully...

I just commit update. The output looks like this in this case:

```
[jhnidek@thinkpad-t580 rhc]$ ./rhc status
Connection status for thinkpad-t580:

● Cannot execute insights-client: fork/exec /usr/bin/insights-client: no such file or directory
● The rhc daemon is inactive

```

Thoughts? Maybe we could emphasis this error more with red/bold text.

subpop: Oh yea I like this. It doesn't mangle the output but does include the error message.

main.go
```diff
@@ -312,13 +312,17 @@ func main() {
 
 				s.Suffix = " Checking Red Hat Insights..."
 				s.Start()
-				isRegistered := insightsIsRegistered()
+				isRegistered, error := insightsIsRegistered()
```
subpop, line 315: `error` is a type name. If the type system doesn't throw a syntax error for naming a variable `error`, it's certainly against convention. Conventionally, the error variable is named `err`.

insights.go
```diff
@@ -16,10 +16,14 @@ func unregisterInsights() error {
...
 
-func insightsIsRegistered() bool {
+func insightsIsRegistered() (bool, error) {
 	cmd := exec.Command("/usr/bin/insights-client", "--status")
 
-	_ = cmd.Run()
+	err := cmd.Run()
 
-	return cmd.ProcessState.Success()
+	if err == nil {
```
subpop, line 24: `ProcessState.Success()` returns true or false depending on the exit code of the command run. Couldn't we just return `cmd.ProcessState.Success()` and `err` here without needing to check if `err` is nil?

```go
err := cmd.Run()
return cmd.ProcessState.Success(), err
```

Then callers checking for whether err is nil or not will be able to act on the error or exit success state.

subpop (changes requested):

main.go
jirihnidek: You are right. I didn't noticed that GoLand marks it as type error. It uses different convention as PyCharm.

insights.go
jirihnidek: Yeah, you are right. We can simplify the code a little bit.

subpop: Why would calling `Success` cause a crash? I looked through the code for the method, and it appears to ultimately end up checking whether the return value of this method is zero or non-zero.

```go
func (w WaitStatus) ExitStatus() int {
	if !w.Exited() {
		return -1
	}
	return int(w>>shift) & 0xFF
}
```

subpop: Ahhh yes, I see. In that case it makes sense that we need to check the error returned from `Run`. If it's an ExitError, that we know that it ran (unsuccessfully). If it's another kind of error, then we shouldn't access the `ProcessState` struct field, since it's likely nil.

jirihnidek: Good point, when `ExitError` is returned, then we should only report that system is not connected to Red Hat Insights. PR is updated.

subpop (approved):

PR #16, opened 2022-09-20
fix: Do not use deprecated io/ioutil

jirihnidek:
The package io/ioutil was deprecated and the CI complained about
that. More detail could be found here: https://go.dev/doc/go1.16#ioutil

### DIFF
--- a/canonical_facts.go
+++ b/canonical_facts.go
@@ -4,7 +4,6 @@ import (
 	"crypto/x509"
 	"encoding/pem"
 	"fmt"
-	"io/ioutil"
 	"net"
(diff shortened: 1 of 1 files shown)


jirihnidek:
Hmm, interesting. Why do you still care about go version `1.15`? It is even not used by RHEL7.

jirihnidek:
If you actually do not care about 1.15 anymore, then I created PR #17.

subpop (approved):
LGTM. Thank you for fixing this. These deprecations have been nagging at the back of my mind for a while now.

PR #17, opened 2022-09-20
feat: Drop version of GO 1.15 from CI testing

jirihnidek:
Do not test Go version 1.15, because it is not used by any supported RHEL (7,8,9). The RHEL7 uses version 1.17.


jirihnidek:
Issue with lint is fixed in this PR: #16

subpop (approved):

PR #18, opened 2022-10-10
feat: Better support for non-interactive mode

### LINKED ISSUE #14: Better support for non-interactive shell environments
Need better support for running rhc in non-interactive shell environments. This would need to do things like disable the ASCII color escape sequences and disable the "spinners". The output needs to be suitable for piping into text files to log the execution behavior.

jirihnidek:
* When `stdout` is not TTY, then do not display spinner and status text and the output of `rhc` should not be colorful, because output is redirected to some file. When temporary status messages and escape sequences were added to log file, then viewing such file could be complicated.
* When `NO_COLOR` environment variable is set, then do not display color nor animation too.
* This commit closes #14 issue

### DIFF
--- a/main.go
+++ b/main.go
@@ -21,9 +21,15 @@ const redColor = "\u001B[31m"
 const greenColor = "\u001B[32m"
 const endColor = "\u001B[0m"
 
-const successPrefix = greenColor + "●" + endColor
-const failPrefix = redColor + "●" + endColor
-const errorPrefix = redColor + "!" + endColor
+// Colorful prefixes
+const ttySuccessPrefix = greenColor + "●" + endColor
+const ttyFailPrefix = redColor + "●" + endColor
+const ttyErrorPrefix = redColor + "!" + endColor
+
+// Black & white prefixes. Unicode characters
+const bwSuccessPrefix = "✓"
+const bwFailPrefix = "𐄂"
+const bwErrorPrefix = "!"
 
 func main() {
 	app := cli.NewApp()
@@ -42,6 +48,12 @@ func main() {
 	log.SetFlags(0)
 	log.SetPrefix("")
 
+	isColorful := true
+
+	successPrefix := ttySuccessPrefix
+	failPrefix := ttyFailPrefix
+	errorPrefix := ttyErrorPrefix
+
 	app.Flags = []cli.Flag{
 		&cli.BoolFlag{
 			Name:   "generate-man-page",
@@ -55,6 +67,12 @@ func main() {
 			Name:   "log-level",
 			Hidden: true,
 			Value:  "error",
+		},
+		&cli.BoolFlag{
+			Name:   "no-color",
+			Hidden: false,
+			Value:  false,
+			EnvVars: []string{"NO_COLOR"},
 		},
 	}
 	app.Commands = []*cli.Command{
@@ -128,20 +146,24 @@ func main() {
 						}
 					}
 
-					s := spinner.New(spinner.CharSets[9], 100*time.Millisecond)
-					s.Suffix = " Connecting to Red Hat Subscription Management..."
-					s.Start()
+					var s *spinner.Spinner
+					if isColorful {
+						s = spinner.New(spinner.CharSets[9], 100*time.Millisecond)
+						s.Suffix = " Connecting to Red Hat Subscription Management..."
+						s.Start()
+					}
 					var err error
 					if c.String("organization") != "" {
 						err = registerActivationKey(c.String("organization"), c.StringSlice("activation-key"), c.String("server"))
 					} else {
 						err = registerPassword(username, password, c.String("server"))
 					}
-					if err != nil {
+					if isColorful {
 						s.Stop()
+					}
+					if err != nil {
 						return cli.Exit(err, 1)
 					}
-					s.Stop()
 					fmt.Printf(successPrefix + " Connected to Red Hat Subscription Management\n")
 				} else {
 					fmt.Printf(successPrefix + " This system is already connected to Red Hat Subscription Management\n")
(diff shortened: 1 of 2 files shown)


jirihnidek:
@subpop Hi, can I ask for PR review? Thanks.

jirihnidek:
BTW: It would be probably good to refactor the code a little bit, because some code is duplicated. There is also one huge method in main function, but I think it would be good to do refactoring in another PR.

main.go
```diff
@@ -42,6 +49,28 @@ func main() {
 	log.SetFlags(0)
 	log.SetPrefix("")
 
+	// Detect if the output goes to TTY or some file
+	var isTTY bool
+	_, err := unix.IoctlGetWinsize(int(os.Stdout.Fd()), unix.TIOCGWINSZ)
+	if err != nil {
```
subpop, line 55: I believe this can be simplified to

```go
_, err := unix.IoctlGetWinsize(int(os.Stdout.Fd()), unix.TIOCGWINSZ)
isTTY := err == nil
```

You'll also probably want to report this error, if it is not nil.

subpop (changes requested):
Looks great overall. I realize I left part of my plan for this feature out of the reporting issue. I want to include support for the `NO_COLOR` environment variable. See the [Output section of the CLIG](https://clig.dev/#output) for more best practices around supporting non-TTY outputs.

main.go
jirihnidek: Yeah, it can be simplified like this. I don't think that I want to report the error (`err`) in this case, because it is absolutely normal, when somebody redirects `stdout` to some file. Or do I miss anything?

jirihnidek:
> Looks great overall. I realize I left part of my plan for this feature out of the reporting issue. I want to include support for the `NO_COLOR` environment variable. See the [Output section of the CLIG](https://clig.dev/#output) for more best practices around supporting non-TTY outputs.

Do you want to support only `NO_COLOR` environment variable or do you want to support other suggestions from [Output section of the CLIG](https://clig.dev/#output) (like `--no-color`)?

main.go
subpop: What sorts of errors do `IoctlGetWinsize` return? Are they fatal to the operation of the program or useful to report to the user? We do appear to be using it exclusively to check whether we're running in a TTY or not, so I can see the reason for not reporting it.

jirihnidek: I was thinking about this more, and maybe we should use something more reliable... What about using this?

https://pkg.go.dev/github.com/mattn/go-isatty#section-readme
GitHub repository of this project https://github.com/mattn/go-isatty

subpop:
> > > Do you want to support only `NO_COLOR` environment variable or do you want to support other suggestions from [Output section of the CLIG](https://clig.dev/#output) (like `--no-color`)?
> 
> > Hm. Yea, I'm leaning towards yes. I like that it expands discoverability of that feature. It's a useful enough and generic enough flag that I'm okay bringing that into the CLI contract.
> 
> Does it mean that I should add support for `--no-color` or should I add support for anything else?

Sorry. Yes. Let's add support for a `--no-color` CLI flag.

main.go
jirihnidek: Maybe using `go-isatty` is overkill in our case, because we need support only for Linux (not for Plan9, Solaris, etc.) We can grab 3 lines of code from there and just use it.

```go
func IsTerminal(fd uintptr) bool {
	_, err := unix.IoctlGetTermios(int(fd), unix.TCGETS)
	return err == nil
}
```

subpop: Yea. I don't mean to diminish the value that package; under certain conditions, I can agree with abstracting that kind of functionality into a package. But for our use here, calling `unix.IoctlGetTermios` directly seems sufficient.

jirihnidek: Funny, I should reload page before replying to any comment and don't trust to automatic updating of GitHub webpage ... It seems we had same idea.

subpop (changes requested):
The `cli` package supports getting values for flags from environment variables. Using the `NO_COLOR` variable as a flag value eliminates some manual checking and consolidates the logic setting the no-color options a bit more.

```
diff --git a/main.go b/main.go
index 4d852d2..d6c97c0 100644
--- a/main.go
+++ b/main.go
@@ -69,9 +69,10 @@ func main() {
 			Value:  "error",
 		},
 		&cli.BoolFlag{
-			Name:   "no-color",
-			Hidden: false,
-			Value:  false,
+			Name:    "no-color",
+			Hidden:  false,
+			Value:   false,
+			EnvVars: []string{"NO_COLOR"},
 		},
 	}
 	app.Commands = []*cli.Command{
@@ -426,23 +427,11 @@ func main() {
 		}
 		log.SetLevel(level)
 
-		// Detect if the output goes to TTY or some file
-		isColorful = isTerminal(os.Stdout.Fd())
-
-		// When environment variable NO_COLOR is set, then do not display colors and animations too.
-		// We do not care about value of NO_COLOR variable
-		if _, isNoColorSet := os.LookupEnv("NO_COLOR"); isNoColorSet {
-			isColorful = false
-		}
-
-		if c.Bool("no-color") {
-			isColorful = false
-		}
-
-		if !isColorful {
+		if c.Bool("no-color") || !isTerminal(os.Stdout.Fd()) {
 			successPrefix = bwSuccessPrefix
 			failPrefix = bwFailPrefix
 			errorPrefix = bwErrorPrefix
+			isColorful = false
 		}
 
 		return nil
```

main.go
```diff
@@ -56,6 +68,11 @@ func main() {
 			Hidden: true,
 			Value:  "error",
 		},
+		&cli.BoolFlag{
+			Name:   "no-color",
+			Hidden: false,
+			Value:  false,
```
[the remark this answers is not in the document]
jirihnidek: Good point.

BTW urfave/cli contains bug. The position of global flag like `--no-color` is not relaxed. You have to use it after main command (not after sub-command). It means, that you can use this:

```
rhc --no-color status
```

but you cannot use this, because it is terminated with error of incorrect usage:

```
rhc status --no-color
```

There is issue about this: https://github.com/urfave/cli/issues/1113

main.go
```diff
@@ -381,6 +426,25 @@ func main() {
...
+		// We do not care about value of NO_COLOR variable
+		if _, isNoColorSet := os.LookupEnv("NO_COLOR"); isNoColorSet {
+			isColorful = false
+		}
+
+		if c.Bool("no-color") {
+			isColorful = false
+		}
+
+		if !isColorful {
```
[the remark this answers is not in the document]
jirihnidek: OK, we can simplify that. BTW: don't you plan to add support for something like `--json` to `status` sub-command?

main.go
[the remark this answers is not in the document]
jirihnidek: BTW: the cli package tries to parse boolean value of environment variable in this case. If the environment variable is "1", "true", "True", "0", "false", "False", then it works as expected. When environment variable is anything else, then the rhc will be terminated with following error:

```
could not parse "OK" as bool value for flag no-color: strconv.ParseBool: parsing "OK": invalid syntax
```

main.go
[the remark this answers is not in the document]
subpop: No plan yet, but output that is easily parsed by machines is something I want to add.

main.go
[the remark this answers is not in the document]
subpop: Yea. I'm okay with that. Those are the values accepted by ParseBool, and it seems reasonable to follow the format set by that function. If we wanted "OK" to be a valid boolean-like, we should submit that upstream in the strconv package. NO_COLOR is a convention, so I can't say this definitively, but it sounds like a boolean variable, so only accepting boolean-like strings is fine with me. Setting NO_COLOR=blue doesn't make sense, so exiting with an error is reasonable.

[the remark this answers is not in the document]
jirihnidek: I don't think that "OK" should be considered as valid boolean value. It was just example.

subpop (approved):
Looks great!

PR #19, opened 2022-10-14
fix: Simple fix of Makefile to not require user input

jirihnidek:
* When `make` is used for building `rhc`, then `gzip` is triggered to compress man page, but it required confirmation that old compressed man page would be rewritten. Adding -f option avoid the need of this confirmation.


subpop:
Good find. The same is true in [yggdrasil's Makefile](https://github.com/RedHatInsights/yggdrasil/blob/main/Makefile#L100) if you want to fix it there too, feel free.

PR #20, opened 2022-11-01
feat: Big refactoring of main.go

jirihnidek:
* Functions of sub-commands were moved to function outside of main function. Created several functions to not duplicate code. Side effect of this effort is better displaying of errors for "connect" sub-command.
* Functions for "main" and "before" actions were also move outside main function
* Fixed one small issue with registration to RHSM. It is possible to use --organization, when --username and --password are specified. The registration process could be improved more, but it should be part of another commit. Added only fix-me comment for this case.

### DIFF
--- a/main.go
+++ b/main.go
@@ -314,7 +314,7 @@ func canonicalFactAction(_ *cli.Context) error {
 // answer on following questions:
 // 1. Is system registered to Red Hat Subscription Management?
 // 2. Is system connected to red Hat Insights?
-// 3. Is rhsm.service running?
+// 3. Is rhcd.service running?
 func statusAction(ctx *cli.Context) error {
 	hostname, err := os.Hostname()
 	if err != nil {
--- a/main.go
+++ b/main.go
@@ -4,6 +4,7 @@ import (
 	"bufio"
 	"encoding/json"
 	"fmt"
+	"golang.org/x/term"
 	"os"
 	"strings"
 	"text/tabwriter"
@@ -14,7 +15,6 @@ import (
 	"github.com/briandowns/spinner"
 	systemd "github.com/coreos/go-systemd/v22/dbus"
 	"github.com/urfave/cli/v2"
-	"golang.org/x/term"
 )
 
 const redColor = "\u001B[31m"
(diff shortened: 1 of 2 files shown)


jirihnidek:
Note: connecting process could look like this:

```
[root@thinkpad-t580 rhc]# ./rhc connect --username admin --password admin --organization donaldduck

STEP  ERROR  
rhc   cannot activate daemon: Unit file rhcd.service does not exist.
```

main.go
```diff
@@ -31,6 +31,398 @@ const bwSuccessPrefix = "✓"
...
+		s.Start()
+	}
+
+	err := function()
+
+	if isColorful {
+		s.Stop()
+	}
+
+	return err
```
subpop, line 51: `defer` can be used to streamline this code a bit.

```suggestion
	defer func() {
	    if isColorful {
	        s.Stop()
	    }
	}
	return function()
```

main.go
```diff
@@ -31,6 +31,398 @@ const bwSuccessPrefix = "✓"
 const bwFailPrefix = "𐄂"
 const bwErrorPrefix = "!"
 
+/**
+ * Call function and, when it is possible display spinner with some progress message
+ */
```
subpop, line 36:
```suggestion
// showProgress calls function and, when it is possible display spinner with
// some progress message.
```

There's a [godoc convention](https://go.dev/blog/godoc) that stipulates that documentation begins with the name of the variable, type, or function/method its documenting. I don't think it's a strict requirement, and `go doc` should still parse the documentation, but it's a common convention.

I've also not seen much use of the `/* */` comment block syntax, though it's certainly supported. Most comment documentation tends to just use line after line of double-slash comments.

Here's what `go doc` outputs for `/* */` comments:

```
[link@thelio rhc]$ go doc ShowProgress
func ShowProgress(progressMessage string, isColorful bool, function func() error) error
    *
      - Call function and, when it is possible display spinner with some
        progress message
```

And here's what it does with double-slash syntax.
```
[link@thelio rhc]$ go doc ShowProgress
func ShowProgress(progressMessage string, isColorful bool, function func() error) error
    ShowProgress calls function and, when it is possible display spinner with
    some progress message.
```

NB: `go doc` only processes exported symbols, so for the sake of the example above, I temporarily changed `showProgress` to `ShowProgress` to demonstrate the doc comment parsing. It's still good practice to use the same convention when documenting unexported symbols.

main.go
```diff
@@ -31,6 +31,398 @@ const bwSuccessPrefix = "✓"
...
+		s.Stop()
+	}
+
+	return err
+}
+
+/**
+ * Try to get color preferences form context
+ */
+func getColorPreferences(ctx *cli.Context) (string, string, string, bool) {
```
subpop, line 57: This could be a good opportunity to use named return values.

```suggestion
func getColorPreferences(ctx *cli.Context) (successPrefix string, failPrefix string, errorPrefix string, isColorful bool) {
```

This way, you can remove the variable declarations in lines 60 through 63. After setting the values, line 75 can call an empty `return`. This removes ambiguity around the order of return values.

main.go
```diff
@@ -31,6 +31,398 @@ const bwSuccessPrefix = "✓"
...
+	}
+
+	fmt.Printf("Connecting %v to %v.\nThis might take a few seconds.\n\n", hostname, Provider)
+
+	successPrefix, _, errorPrefix, isColorful := getColorPreferences(ctx)
+
+	/* 1. Register to RHSM, because we need to get consumer certificate. This blocks following action */
+	start = time.Now()
+	/* Note: All successful messages are printed inside registerRHSM() */
+	err = registerRHSM(ctx)
```
subpop, line 201: This doesn't follow the same pattern as steps 2 and 3. Can it be refactored so that all 3 steps follow a similar code pattern?

subpop (changes requested):
Nicely executed! This is a great improvement, and sets us up for further refactoring as this program matures.

main.go
jirihnidek: Yeah, I also do not like that the pattern is different in this case. I will try to think about it more and change it to use same pattern in all cases.

main.go
jirihnidek: Yeah, I know that Go has this functionality, but old C coder tents to use old-fashion style.

main.go
jirihnidek: OK, no problem, I can use this pattern.

main.go
jirihnidek: Good idea, and I think that we can streamline it little bit more.

main.go
```diff
@@ -14,23 +15,405 @@ import (
...
+
+	connectedPrefix, _, errorPrefix, isColorful := getColorPreferences(ctx)
+
+	/* 1. Register to RHSM, because we need to get consumer certificate. This blocks following action */
+	start = time.Now()
+	var returnedMsg string
+	err, returnedMsg = registerRHSM(ctx)
+	if err != nil {
+		errorMessages["rhsm"] = fmt.Errorf("cannot connect to Red Hat Subscription Management: %w", err)
+		fmt.Printf(errorPrefix + " Cannot connect to Red Hat Subscription Management\n")
```
jirihnidek, line 187: BTW: When this action is not successful, then it would be better to skip following steps ("Connecting to Red Hat Insights" and "Starting rhcd"), because these steps could not be successful anyway.

subpop: Yea, that's a good point. This action is required for the subsequent two. Maybe that's reason enough to make it "different" than the other two. It must succeed; if it does not, the program should exit.

main.go
```diff
@@ -14,23 +15,405 @@ import (
...
+	} else {
+		fmt.Printf(connectedPrefix + " " + returnedMsg + "\n")
+	}
+	durations["rhsm"] = time.Since(start)
+
+	/* 2. Register insights-client */
+	start = time.Now()
+	err = showProgress(" Connecting to Red Hat Insights...", isColorful, registerInsights)
```
jirihnidek, line 195: @subpop If registration of insights-client failed, would it be reason to not to try start `rhcd` too?

subpop: No. `rhcd` does not require a functional `insights-client` in order to operate itself. If `insights-client` fails, we should still start `rhcd`.

jirihnidek:
The connection looks like this, when wrong password is provided:

```
[root@rhel9 rhc]# ./rhc connect --username REDACTED --password foo
Connecting rhel9 to Red Hat.

! Cannot connect to Red Hat Subscription Management
● Skipping connection to Red Hat Insights
● Skipping activation of rhc daemon

STEP  ERROR  
rhsm  cannot connect to Red Hat Subscription Management: error: Invalid username or password. To create a login, please visit https://www.redhat.com/wapps/ugc/register.html
```

main.go
jirihnidek: Yeah, I agree. It is already implemented in the way that you suggested. I only tried to explain, why I coded it originally in that way.

main.go
jirihnidek: Yeah, I agree. Done.

main.go
jirihnidek: OK, Done.

jirihnidek:
@subpop Hi, can you please review this PR once again? Thanks in advance.

jirihnidek:
BTW: created BZ for one issue fixed as part of this PR: https://bugzilla.redhat.com/show_bug.cgi?id=2142797

main.go
```diff
@@ -14,23 +15,413 @@ import (
...
+	}
+	fmt.Println(string(data))
+	return nil
+}
+
+// statusAction tries to print status of system. It means that it gives
+// answer on following questions:
+// 1. Is system registered to Red Hat Subscription Management?
+// 2. Is system connected to red Hat Insights?
+// 3. Is rhsm.service running?
```
subpop, line 317: I think this is a mental slip-up, but this should be rhcd.service, I think?

subpop (approved):
Looks great! I found what looked like one typo in a comment. Otherwise, I'm ready to merge this.

main.go
jirihnidek: Yeah, you are right. I'm sorry.

PR #21, opened 2022-11-01
Update-actions

subpop:
Update action versions (this removes the deprecation notices about node12 actions) and update the Go version matrix to drop 1.17 and include 1.18.


strider (approved):
LGTM

PR #24, opened 2022-11-16
feat: Ask for org during registration of system

jirihnidek:
* BZ: https://bugzilla.redhat.com/show_bug.cgi?id=2142797
* When organization was not specified using --ogranization, then
   RHSM registration was terminated with error and it wasn't
   possible to input organization like username/password.
* When user is member of more organizations and he/she forget
   to specify organization using --organization CLI option or
   the rhc connect is triggered without any CLI options, then
   rhc tries to get list of organizations. It prints this
   list as a hint and then it asks for organization
* Refactoring of code
  - Wrapped too long lines
  - Function `unpackError()` renamed to `unpackRHSMErrorMsg()`
  - Introduced new function `unpackRHSMError()`

### DIFF
--- a/main.go
+++ b/main.go
@@ -159,14 +159,17 @@ func registerRHSM(ctx *cli.Context) (string, error) {
 					}
 
 					// Ask for organization and display hint with list of organizations
-					fmt.Printf("Available Organizations:\n")
-					for _, org := range orgs {
-						fmt.Printf(" - %v\n", org)
-					}
-					fmt.Printf("\n")
-
 					scanner := bufio.NewScanner(os.Stdin)
-					fmt.Print("Organization: ")
+					fmt.Println("Available Organizations:")
+					writer := tabwriter.NewWriter(os.Stdout, 0, 2, 2, ' ', 0)
+					for i, org := range orgs {
+						_, _ = fmt.Fprintf(writer, "%v\t", org)
+						if (i+1)%4 == 0 {
+							_, _ = fmt.Fprint(writer, "\n")
+						}
+					}
+					_ = writer.Flush()
+					fmt.Print("\nOrganization: ")
 					_ = scanner.Scan()
 					organization = strings.TrimSpace(scanner.Text())
 					fmt.Printf("\n")


jirihnidek:
The registration process can look like this:

```
[root@thinkpad-t580 rhc]# ./rhc connect

Username: admin
Password: 

Hint: User "admin" is member of following organizations: admin, snowwhite, donaldduck
Organization: admin


STEP  ERROR  
rhc   cannot activate daemon: Unit file rhcd.service does not exist.
```

register.go
```diff
@@ -17,55 +17,148 @@ func getConsumerUUID() (string, error) {
...
-		return fmt.Errorf("warning: the system is already registered")
+		return orgs, fmt.Errorf("warning: the system is already registered")
 	}
 
 	registerServer := conn.Object("com.redhat.RHSM1", "/com/redhat/RHSM1/RegisterServer")
 
 	var privateDbusSocketURI string
-	if err := registerServer.Call("com.redhat.RHSM1.RegisterServer.Start", dbus.Flags(0), "").Store(&privateDbusSocketURI); err != nil {
-		return err
+	if err := registerServer.Call(
```
subpop, line 102: Are you making these formatting changes yourself? Or are you using an IDE that's formatting these long function call expressions automatically into new-lines-by-parameter? I don't believe that `go fmt` cares one way or another, but including them in a patch like this just adds unnecessary change to the patch set.

register.go
```diff
@@ -17,55 +17,148 @@ func getConsumerUUID() (string, error) {
...
+				username,
+				password,
+				map[string]string{},
+				"",
+			)
+			var s string
+			err = orgsCall.Store(&s)
+			if err == nil {
+				orgs, err = unpackOrgs(s)
+				if err == nil {
```
subpop, line 150: Same thing here.

register.go
```diff
@@ -17,55 +17,148 @@ func getConsumerUUID() (string, error) {
...
+				"com.redhat.RHSM1.Register.GetOrgs",
+				dbus.Flags(0),
+				username,
+				password,
+				map[string]string{},
+				"",
+			)
+			var s string
+			err = orgsCall.Store(&s)
+			if err == nil {
```
subpop, line 148: Can you flip the logic around this to `if err != nil`? The rest of the program uses that pattern for error checking, and I think maintaining a consistent pattern is important.

register.go
```diff
@@ -127,33 +235,51 @@ func unregister() error {
...
+type RHSMError struct {
+	Exception string `json:"exception"`
+	Severity  string `json:"severity"`
+	Message   string `json:"message"`
+}
+
+// unpackError tries to unpack JSON document (part of error) into the structure RHSMError. When it is
+// not possible to parse error into structure, then return empty structure and error. When it is
+// possible to parse error into structure, then returned structure contains valid data and err is nil.
+func unpackError(err error) (RHSMError, error) {
```
subpop, line 261: There's a convention I've discovered that can ensure your caller doesn't mistakenly attempt to use the zero value of the struct returned by your function. If you return a pointer to `RHSMError`, you can return `nil` when returning an error. This prevents a caller of your function from ignoring the error (intentionally or unintentionally) and proceeding as if the function returned successfully. Returning a zero value struct along with an error can lead to silent problems later on when a struct field is accessed (and returns *its* zero value). Returning `nil` will cause that field access to panic and abort the program.

subpop (changes requested):

register.go
jirihnidek: Yes, I'm making these changes myself. I consider too long lines as hard to read. Should I try to split it to another PR/commit? Or is mentioning this change in commit message enough?

register.go
jirihnidek: Sure, I can.

register.go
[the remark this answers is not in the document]
jirihnidek: Yeah, good point.

register.go
subpop: I'm fine moving to a function call style like this. Could you move them into a separate commit? The same PR is fine, but they are a separate change from the logic changes adding the organization checking code. Does `go fmt` change the number of tabs/spaces used when you format the function calls like this?

register.go
jirihnidek: Yeah, I like this pattern. I will change the code.

register.go
subpop: `go fmt` is the de facto standard, so we should use it. I'm surprised Goland does not run `go fmt` before saving. One of the Go proverbs:

> Gofmt's style is no one's favorite, yet gofmt is everyone's favorite.

https://www.youtube.com/watch?v=PAAkCSZUG1c&t=8m43s

jirihnidek: I fixed the indentation using `go fmt`. I tried to split commit into two commits, but it would require re-implementing two steps of changes again, because there were two changes tangled together. I really don't like being slave of nice git history... I tried to describe all changes in commit message.

register.go
```diff
@@ -127,33 +239,51 @@ func unregister() error {
...
+type RHSMError struct {
+	Exception string `json:"exception"`
+	Severity  string `json:"severity"`
+	Message   string `json:"message"`
+}
+
+// unpackRHSMError tries to unpack JSON document (part of error) into the structure RHSMError. When it is
+// not possible to parse error into structure, then return nil pointer and error. When it is
+// possible to parse error into structure, then returned pointer on structure contains valid data and err is nil.
+func unpackRHSMError(err error) (*RHSMError, error) {
```
subpop, line 265: I'm having trouble trying to figure out the difference between `unpackRHSMError` and `unpackRHSMErrorMsg`. They seem to be doing mostly the same thing. Maybe `unpackRHSMErrorMsg` needs to be moved to a method on the `RHSMError` struct?

```go
func (r RHSMError) Error() string {
    return fmt.Sprintf("%v: %v", r.Severity, r.Message)
}
```

This has the added advantage of implementing the Error interface, allowing an RHSMError value to be used as an `error` type.

register.go
```diff
@@ -127,33 +239,51 @@ func unregister() error {
 		return fmt.Errorf("warning: the system is already unregistered")
 	}
 
-	if err := conn.Object("com.redhat.RHSM1", "/com/redhat/RHSM1/Unregister").Call("com.redhat.RHSM1.Unregister.Unregister", dbus.Flags(0), map[string]string{}, "").Err; err != nil {
-		return unpackError(err)
+	if err := conn.Object(
```
subpop, line 242: This is purely a formatting difference, but it might make these longer function calls more legible. Rather than putting the err declaration inline within the `if` condition, move it to an explicit line:

```go
err := conn.Object(
    "com.redhat.RHSM1",
	"/com/redhat/RHSM1/Unregister").Call(
	"com.redhat.RHSM1.Unregister.Unregister",
	dbus.Flags(0),
	map[string]string{},
	"")
if err != nil {
    return unpackRHSMErrorMsg(err)
}
```

subpop (changes requested):

register.go
jirihnidek: Function `unpackRHSMErrorMsg()` is needed to get error message in the most cases, but it is necessary to unpack only the error (and use `unpackRHSMError()`) in `registerUsernamePassword()`, because it is necessary to compare `rhsmError.Exception` with `""OrgNotSpecifiedException"`.

I can move `unpackRHSMErrorMsg` as a method of `RHSMError` struct.

register.go
jirihnidek: OK, np.

register.go
subpop: Yea, I can see they both have a purpose, but it's not clear what each function's purpose is. Since `unpackRHSMErrorMsg` seems to always require an `RHSMError` value, it effectively is a method on that type. Structuring things that way should make each method's intended purpose more explicit.

subpop:
I see more `goimports` formatting changes that need to be applied to `main.go`

```diff
diff -u main.go.orig main.go
--- main.go.orig        2022-12-05 12:35:34.050258185 -0500
+++ main.go     2022-12-05 12:35:34.050258185 -0500
@@ -4,12 +4,13 @@
        "bufio"
        "encoding/json"
        "fmt"
-       "golang.org/x/term"
        "os"
        "strings"
        "text/tabwriter"
        "time"
 
+       "golang.org/x/term"
+
        "git.sr.ht/~spc/go-log"
 
        "github.com/briandowns/spinner"
diff -u register.go.orig register.go
--- register.go.orig    2022-12-05 12:35:34.055258244 -0500
+++ register.go 2022-12-05 12:35:34.055258244 -0500
@@ -134,7 +134,7 @@
 
                // Is unpacked error RHSMError
                rhsmError, ok := err.(RHSMError)
-               if ! ok {
+               if !ok {
                        return orgs, err
                }
```

main.go
```diff
@@ -143,28 +143,54 @@ func registerRHSM(ctx *cli.Context) (error, string) {
...
+					}
+
+					// Ask for organization and display hint with list of organizations
+					fmt.Printf(
+						"Hint: User \"%s\" is member of following organizations: %v\n",
+						username,
+						strings.Join(orgs, ", "),
+					)
+					scanner := bufio.NewScanner(os.Stdin)
+					fmt.Print("Organization: ")
```
subpop, line 167: I don't have an account to test it with, but would this work?

```suggestion
					if len(orgs) <= 3 {
						fmt.Printf("Organization (%v): ", strings.Join(orgs, ", "))
					} else {
						fmt.Println("Current Organizations")
						for _, org := range orgs {
							fmt.Printf("- %v", org)
						}
						fmt.Printf("Organization: ")
					}
```

subpop (changes requested):

main.go
jirihnidek: Why should we print different messages for two/three organizations and for more than 3 organizations? I think that we should be consistent regardless of number of organizations.

subpop: You are correct. A new-line separated list doesn't make sense if the number of items in the list is really big. But a comma-separated list is equally difficult to view content in that case too. Could we format the list into a tabwriter with 3 or 4 columns?

jirihnidek: It could probably work for long list of organizations. Question is, how does it look like for small number of organizations (<= 4). I would not print there "Current Organizations:", but I would print there: "Hint: User "ldupont@redhat.com" is member of following organizations:"

Do you want to introduce printing of columns as part of this PR or could it be implemented later in another PR?

subpop: How about "Available Organizations"?

jirihnidek: OK

jirihnidek:
Hi @subpop, I updated the hint and it is printed only: "Available Organizations:" as a hint. Do you plan to implement your tabwriter in another PR?

main.go
```diff
@@ -159,14 +159,17 @@ func registerRHSM(ctx *cli.Context) (string, error) {
...
-						fmt.Printf(" - %v\n", org)
-					}
-					fmt.Printf("\n")
-
 					scanner := bufio.NewScanner(os.Stdin)
-					fmt.Print("Organization: ")
+					fmt.Println("Available Organizations:")
+					writer := tabwriter.NewWriter(os.Stdout, 0, 2, 2, ' ', 0)
+					for i, org := range orgs {
+						fmt.Fprintf(writer, "%v\t", org)
```
jirihnidek, line 166: The `fmt.Fprintf()` returns `int` and `err`. I suppose that we do not care about error in this case, then please use:

```go
_, _ = fmt.Fprintf(writer, "%v\t", org)
```

main.go
```diff
@@ -159,14 +159,17 @@ func registerRHSM(ctx *cli.Context) (string, error) {
...
+						if (i+1)%4 == 0 {
+							fmt.Fprint(writer, "\n")
```
jirihnidek, line 168: Same issue here

main.go
```diff
@@ -159,14 +159,17 @@ func registerRHSM(ctx *cli.Context) (string, error) {
...
+						}
+					}
+					writer.Flush()
```
jirihnidek, line 171: The `writer.Flush()` can also return error. I would use this:

```go
_ = writer.Flush()
```

jirihnidek (commented):
Thanks for updates. I have never provided review in my own PR. :smiley: I have few small requests.

main.go
```diff
@@ -143,28 +144,58 @@ func registerRHSM(ctx *cli.Context) (error, string) {
...
+					writer := tabwriter.NewWriter(os.Stdout, 0, 2, 2, ' ', 0)
+					for i, org := range orgs {
+						_, _ = fmt.Fprintf(writer, "%v\t", org)
+						if (i+1)%4 == 0 {
+							_, _ = fmt.Fprint(writer, "\n")
+						}
+					}
+					_ = writer.Flush()
+					fmt.Print("\nOrganization: ")
+					_ = scanner.Scan()
```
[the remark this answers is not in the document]
jirihnidek: It is good point, but it is generic problem of `rhc`. When you try to register using `rhc` and you redirect output to e.g. some file like this `rhc connect > ./log.txt` (no username/password), then `rhc` blocks without any error written to `stderr`. I think we could create issue for this. I think that this could be something nice, useful and simple to fix for Go newbie. ;-)

subpop (approved):
LGTM!

PR #29, opened 2022-12-23
feat: retrieve profile and print it as part of the registration

ahitacat:
* This commit introduces a new output when connecting rhc.
* When using connect, and after registering in rhsm, make a call to the API to get the profile information and show it.
* Introduced a new module to create http client configure to use mTLS.
* Introduced a new struct to save the Profile.
* Create a JSON Marshaler for meaningful print of the Profile.

The registration process can look like:
```
[ahitacat@localhost rhc]# ./rhc connect  
Connecting localhost.localdomain to Red Hat.

● This system is already connected to Red Hat Subscription Management

System profile:
{
        "Remote Configuration": true,
        "Remediations": true,
        "Compliance": true
}

This system is successfully connected to Red Hat, there may be additional tasks in the background according to the RHC Manager settings.
Use the following command to check them:
        # journalctl -u rhcd -f 

```


internal/http/client.go
```diff
@@ -0,0 +1,77 @@
...
+
+type Client struct {
+	client *http.Client
+}
+
+func NewHTTPClient() *Client {
+
+	tlsConfig, err := CreateTLSClientConfig()
+	if err != nil {
+		return nil
```
subpop, line 19: It's good practice to propagate the error back up to the caller, letting it handle the error appropriately.

internal/http/client.go
```diff
@@ -0,0 +1,77 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("cannot copy system certificate pool: %w", err)
+	}
+
+	pool.AppendCertsFromPEM(rootCA)
+
+	return tlsConfig, nil
+}
+
+func (c *Client) Get(url string) (*http.Response, error) {
```
subpop, line 70: It's customary to define methods on a struct closer to the struct declaration. Move this above `CreateTLSClientConfig`, or move the declaration of `Client` to below the top-level package functions.

main.go
```diff
@@ -196,7 +248,25 @@ func connectAction(ctx *cli.Context) error {
...
 
-	/* 2. Register insights-client */
+	/* 2 Show Profile information */
+	// Generate a new http client with mtls configure
+	client := http.NewHTTPClient()
+	if client != nil {
+		// Get the user profile
+		profile, err := getUserProfile(client)
+		if err != nil {
+			errorMessages["rhsm"] = fmt.Errorf("Cannot get the user profile: %w", err)
```
subpop, line 258: "user profile" is misleading. It's not a profile of user settings, it's a profile of configured state. "system profile" or "configuration profile" is more accurate.

main.go
```diff
@@ -196,7 +248,25 @@ func connectAction(ctx *cli.Context) error {
...
+		}
+
+		fmt.Printf("\nSystem profile:\n")
+		showUserProfile(profile)
+	}
+
+	fmt.Printf("\nThis system is successfully connected to Red Hat, there may be additional tasks" +
```
subpop, line 265: I don't like the way this sentence is written, and it doesn't include the ● prefix like the rest of the commands. Could we rearrange the output so that all the bullet "connected" messages are printed, and the configuration profile comes last? Something like:

```
● This system is already connected to Red Hat Subscription Management
● Activated the rhc daemon

✓ Successfully connected to Red Hat!
🛈 Additional operations may be executing on the system to activate the following features:

FEATURE               ACTIVE  
Remote Configuration  true
Remediations          true
Compliance            true

Run the following command to monitor output: journalctl -u rhcd -f
```

subpop (changes requested):

main.go
jirihnidek: BTW: was it discussed with any UX designer?

Shouldn't be this output part of `status` sub-command? I think that it would be useful to know such information anytime (not only after connecting). The `connect` command should only connect the system with some services, but it looks like that calling `GET https://cert.console.redhat.com/api/config-manager/v2/profiles/current` only gather some status information and it doesn't connect system with any new service.

internal/http/client.go
```diff
@@ -0,0 +1,77 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("cannot read cert-file: %w", err)
+	}
+
+	keyData, err = os.ReadFile("/etc/pki/consumer/key.pem")
+	if err != nil {
+		return nil, fmt.Errorf("cannot read key-file: %w", err)
+	}
+
+	rootCA, err = os.ReadFile("/etc/rhsm/ca/redhat-uep.pem")
```
jirihnidek, line 48: Well, it could be more complicated. There could be other CA certificates (e.g. in case Satellite is used) and it is recommended to read all CA certificates from this directory. On the other hand I'm not sure if `https://cert.console.redhat.com/api/config-manager/v2/profiles/current` API could provided by Satellite server.

Anyway the `rhc` should recover from such situation (Satellite is used) gracefully and it should not print any error in such case.

jirihnidek (changes requested):
Interesting. :+1: I have few comments and requests in the code.

main.go
ahitacat: I would say this has not being discussed with any UX designer, did you consider it should?

I think this could be a great feature to add to the `status` sub-command.

main.go
```diff
@@ -15,6 +16,7 @@ import (
 
 	"github.com/briandowns/spinner"
 	systemd "github.com/coreos/go-systemd/v22/dbus"
+	"github.com/redhatinsights/rhc/internal/http"
```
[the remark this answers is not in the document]
ahitacat: This URL is created in the PR when added the http package. It is a way to import an internal package https://www.practical-go-lessons.com/chap-11-packages-and-imports#import-path

main.go
```diff
@@ -94,6 +96,55 @@ func showErrorMessages(action string, errorMessages map[string]error) error {
...
+	Active      bool `json:"active"`
+	Remediation bool `json:"remediations"`
+	Compliance  bool `json:"compliance"`
+}
+
+// getUserProfile will retrieve the profile the system is configured
+
+func getUserProfile(client *http.Client) (*userProfile, error) {
+	var profile userProfile
+	res, err := client.Get("https://cert.console.redhat.com/api/config-manager/v2/profiles/current")
```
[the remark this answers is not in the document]
subpop: It is. https://github.com/RedHatInsights/config-manager/blob/master/internal/http/v2/openapi.json

main.go
```diff
@@ -196,7 +248,25 @@ func connectAction(ctx *cli.Context) error {
 	}
 	durations["rhsm"] = time.Since(start)
 
-	/* 2. Register insights-client */
+	/* 2 Show Profile information */
+	// Generate a new http client with mtls configure
```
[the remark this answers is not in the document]
subpop: Isn't it, technically, mutual TLS authentication though? The client has both a public and  a private key.

main.go
subpop: We should include this in the `status` subcommand, yes. This PR is introducing the changes requested by RHIN-501. That RFE doesn't expand the output of such status to the `status` subcommand, but we could certainly do so.

internal/http/client.go
ahitacat: That's right, that URL didn't provide the information by Satellite. But the API exists for Satellite (it's other URL). `rhc` should recover gracefully for the situation but I think that it also need to print the error if it can not reach the endpoint.

main.go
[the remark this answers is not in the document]
jirihnidek: Yes, it is mutual authentication, but this "feature" has been ever integrated part of TLS specification and it is really unusual to use `mTLS` shortcut.

jirihnidek (changes requested):
When I tried to connect to my local candlepin server, then rhc was terminated with this:

```
[root@thinkpad-t580 rhc]# ./rhc connect --username admin --password admin --organization admin

System profile:
FEATURE  ACTIVE  
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x562d619bde59]

goroutine 1 [running]:
main.showUserProfile(0x0)
	/home/jhnidek/githup/redhat_insights/rhc/main.go:104 +0x119
main.connectAction(0xc000133c20?)
	/home/jhnidek/githup/redhat_insights/rhc/main.go:262 +0x605
github.com/urfave/cli/v2.(*Command).Run(0xc00015c480, 0xc000110bc0)
	/home/jhnidek/go/pkg/mod/github.com/urfave/cli/v2@v2.3.0/command.go:163 +0x5dc
github.com/urfave/cli/v2.(*App).RunContext(0xc0001036c0, {0x562d61bc8910?, 0xc00013a000}, {0xc000130000, 0x8, 0x8})
	/home/jhnidek/go/pkg/mod/github.com/urfave/cli/v2@v2.3.0/app.go:313 +0xb7d
github.com/urfave/cli/v2.(*App).Run(...)
	/home/jhnidek/go/pkg/mod/github.com/urfave/cli/v2@v2.3.0/app.go:224
main.main()
	/home/jhnidek/githup/redhat_insights/rhc/main.go:721 +0x11c5
```

main.go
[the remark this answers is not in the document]
jirihnidek: I see, but the package `http` is part of this repository. Shouldn't we import it using something like `import internal/http`? This looks like that `http` package is part of another repository, and it is not true.

Is there any benefit of using this approach?

ahitacat:
New changes pushed:

- better propagation of errors
- refactor name of the profile retrieve structure
- Change the output to be aligned
    -  Added a new color and icon for information messages
- Added the reference to the API information
- Avoid to print back-trace if an error occurs while getting the system profile information
     - it will exits gratefully without printing the information
     - it will print the error message at the end of the execution
 - Implemented the call for connections done via Satellite and stage
     - rhc now configures the URL getting the information through DBus
     - It also loads the configuration from the config file

main.go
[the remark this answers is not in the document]
jirihnidek: Yeah, I was wrong in the suggestion. I know that `import "internal/http"` would try to import package from `/usr/lib/golang/src/`. I wanted to suggest to use relative imports.

When you use `import "github.com/redhatinsights/rhc/internal/http"`, then you probably cannot share code on another remote repository e.g. on gitlab or dist-git. You actually can, but then it will not work as expected.

jirihnidek:
It looks like this ATM:

image

jirihnidek:
You can compare it with main branch:

image

main.go
```diff
@@ -227,10 +330,33 @@ func connectAction(ctx *cli.Context) error {
...
+			profile, err := getConfProfile(client)
+			if err != nil {
+				errorMessages[BrandName] = fmt.Errorf("Cannot get the user profile: %w", err)
+			} else {
+				showConfProfile(profile)
+			}
+
+		}
+		fmt.Printf("\n" + infoPrefix + " Run the following command to check them:\n\t# journalctl -u rhcd -f \n")
+	}
```
jirihnidek, line 354: Again, I'm not UX designer, but I feel that this PR adds unnecessary clutter to the output of `rhc connect` command. I believe that it could be simplified. Like this:

```
[root@rhel9 rhc]# ./rhc connect --username user@company.com
Connecting rhel9 to Red Hat.

Password: 

● Activated the rhc daemon
   ● Remote Configuration
   ● Remediations
   ● Compliance

```

I don't believe that it is good idea to show following hint: `Run the following command ... journalctl -u rhcd -f`, because it looks like that something went wrong.

Running this command (`journalctl -u rhcd -f`) is also not helpful too much... I'm adding example from my system:

```
[root@rhel9 rhc]# journalctl -u rhcd -f
Jan 24 11:47:00 rhel9 systemd[1]: Started Red Hat connector daemon.
Jan 24 11:59:50 rhel9 systemd[1]: Stopping Red Hat connector daemon...
Jan 24 11:59:50 rhel9 systemd[1]: rhcd.service: Deactivated successfully.
Jan 24 11:59:50 rhel9 systemd[1]: Stopped Red Hat connector daemon.
Jan 24 11:59:50 rhel9 systemd[1]: rhcd.service: Consumed 1.892s CPU time.
Jan 24 12:00:50 rhel9 systemd[1]: Started Red Hat connector daemon.
^C
```

I don't see there any useful information and it is pretty confusing.

garrett:
Hi! I'm a designer at Red Hat and working in the Cockpit team. I've been using Linux since 1996 and working in design (in one sub-field or another) since the mid-90s. And @jirihnidek asked me to look over this PR.

---

I don't know all the context, and haven't used `rhc` before, but my initial impressions are:

- This is _extremely verbose_ output.
   - A lot of it seems to be debugging output and half of that is almost apologizing for the tool being slow and connecting to the servers. Surely this can be shortened?
   - How much of this information is really necessary? For example: It important to show that the system is already registered? Or can you just show exceptional cases, like when a system isn't registered?
   - Default terminals will not be able to show all this information. It will not only require someone to scroll, but read a lot of text to understand what's going on.
- Does it take some time before someone can even enter a password? Can this be changed so a password prompt shows up first? Or is that message in the wrong place?
- Unless it's absolutely necessary (and it rarely is), don't show 'daemon' or 'demon' in text that people using your software will see.
- Generally, don't use a table in text.
   - Screen readers will say something like: "FEATURE ACTIVE Remote configuration true Remediations true".
   - In some cases, some screen readers read uppercase with emphasis... but others will sometimes even spell it out. "eff ee ay tee you are ee ay see tee eye vee ee remote configuration..." (I think most these days only behave like this when they see an unfamiliar word. But I am unsure about that. There are a lot of screen readers out there.)
   - It would be better to show it as "Remote configuration: active" (or inactive). 
   - Perhaps sort the list so active features are first? Then, optionally, inactive ones?
   - If they don't need to know which features are inactive, you could just show: "Active features: One, two, three" on one line.
   - Stepping back: Do people even need to know which features are active or inactive? How often will features be turned on and off? How much does it impact usage?

Can you provide a screencast so I can better understand what's happening (including the timing)? 

GitHub supports videos (in webm, mp4, etc.), so you don't have to upload it elsewhere. If you don't have a screen recording app, I can recommend Kooha (https://flathub.org/apps/details/io.github.seadve.Kooha). Or, if you're on GNOME, you can hit printscreen and use its built-in screen recording.

---

Additionally, there are a bunch of best practices for command line interfaces at https://clig.dev/ (it's a pretty long document, but it's split into sections and has a table of contents at the side, so you can jump to the part that's relevant to what you're wondering about... for example, output: https://clig.dev/#output).

jirihnidek (commented):
I'm just adding my comments and ideas, but it is only very subjective, because I'm not UX designer.

main.go
subpop: I like your output suggestion, @jirihnidek. Taking into consideration some of Garrett's suggestions, what about:

```
● Activated Remote Configuration
● Activated Remediations
● Activated Compliance
```

Starting `rhcd.service` is implied in this case, because one of the 3 "Activated" lines are present. If one of the profile options is not enabled, we could either omit it, or print "Not activating Compliance".

jirihnidek:
This is screencast from main branch:

[rhc-connect-main-branch.webm](https://user-images.githubusercontent.com/2057012/214309448-4942347c-9353-4d02-ac8b-eb40e315f785.webm)

subpop:
This is all excellent feedback. The current UX of `rhc` clearly has room for improvement. I'll capture it in a separate issue for us to work through and implement in a second round of improvements. For this PR specifically, the backing enhancement request states "Provide feedback to the user on what is happening behind he scene when registering via rhc client.". After running `rhc connect`, the unit `rhcd.service` starts; then it is possible, depending on how the user's Red Hat account is set up on https://console.redhat.com/settings/connector, that the system will receive an Ansible playbook, and will execute it. That possibility of additional "background operations" occurring is what we'd like to communicate to the user after they connect their system. There is a service in console.redhat.com called "config-manager" that holds the state of those operations and determines whether or not to send a playbook to the host. The current state of that service can be queried by the host. Our implementation here does just that. After connecting and activating `rhcd.service`, it queries config-manager for the current profile and gets a JSON response:

```json
{
  "insights": true,
  "compliance": true,
  "remediations": false
  "active": true
}
```

These booleans indicate whether or not the playbook to enable each service will be sent to the host for execution. In the example above, the host will be sent a playbook to install `openscap` (required for the Insights Compliance app).

We'd like to somehow telegraph to the user the value of those states and what may occur on the system as a result. After digesting this conversation more, I'm thinking we add a line after activating rhcd that prints the enabled services:

```
[root@rhel9 rhc]# ./rhc connect --username user@company.com
Connecting rhel9 to Red Hat.

Password: 

● Activated the rhc daemon
● Enabled console.redhat.com services: Insights, Compliance, Remedations

```

ahitacat:
New changes pushed!
@garrett thanks for this incredible feedback.
I have reduce the quantity of output printed, now is like the last suggestion of @subpop 
image

subpop:
I'm okay with this, but as it's my suggestion, I feel like I need to defer the approval to one other person.

main.go
```diff
@@ -15,22 +18,27 @@ import (
...
 	"github.com/briandowns/spinner"
 	systemd "github.com/coreos/go-systemd/v22/dbus"
+	"github.com/redhatinsights/rhc/internal/http"
 	"github.com/urfave/cli/v2"
+	"github.com/urfave/cli/v2/altsrc"
 )
 
 const redColor = "\u001B[31m"
 const greenColor = "\u001B[32m"
+const infoColor = "\u001b[36m"
```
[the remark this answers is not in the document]
subpop: Good point; I like the idea of keeping the constant names consistent.

main.go
```diff
@@ -94,6 +104,108 @@ func showErrorMessages(action string, errorMessages map[string]error) error {
...
+			baseURL, err = url.Parse("https://cert.console.redhat.com")
+			if err != nil {
+				return nil, fmt.Errorf("cannot parse URL: %w", err)
+			}
+		} else {
+			baseURL, err = url.Parse("https://cert.console.stage.redhat.com")
+			if err != nil {
+				return nil, fmt.Errorf("cannot parse URL: %w", err)
+			}
+		}
```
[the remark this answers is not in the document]
subpop: Yea. This is likely to cause problems later when attempting to use `rhc` against a Satellite server. `rhc` already has the `--server` flag. Can we use that? Or perhaps we should add a second flag: `--api-server` to set the base URL for API calls that require ConsoleDot services?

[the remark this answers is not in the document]
subpop: `rhc` already gets the base URL value by calling the D-Bus method on `com.redhat.RHSM1`. I'd much rather couple the two programs connected over an API like that than through a configuration file format. Maybe we deprecate `--server` and split it into two flags: `--registration-server` and `--api-server`?

main.go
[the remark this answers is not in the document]
jirihnidek: I still think that this is issue. Imagine that this code is shared for some reason on some gitlab (notgithub) or fork. It does not matter, but some feature branch is created. Then old code from "github.com/redhatinsights/rhc/internal/http" is still used despite there is new code in feature branch.

main.go
```diff
@@ -94,6 +104,108 @@ func showErrorMessages(action string, errorMessages map[string]error) error {
...
+// MarshalJSON will convert the ConfProfile to a more meaningful structure
+func (p ConfProfile) MarshalJSON() ([]byte, error) {
+	type P struct {
+		RemoteConfiguration bool `json:"Remote Configuration"`
+		Compliance          bool `json:"Compliance"`
+		Insights            bool `json:"Insights"`
+		Remediation         bool `json:"Remediations"`
+	}
+	return json.Marshal(P(p))
+}
```
jirihnidek, line 123: I don't believe that this function is necessary. BTW: The structure `ConfProfile` is actually defined twice.

main.go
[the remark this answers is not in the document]
jirihnidek: BTW:  I don't think that `rhc` should have `--server` CLI option. Users do not modify this URL at all in 99.9999%. When they use Satellite, then the hostname in `/etc/rhsm/rhsm.conf` is modified by bootstrap script. This value is IMHO modified mostly by red hatter during testing. Thus it should be possible to change the server URL only in `rhsm.conf` file.

main.go
```diff
@@ -94,6 +104,108 @@ func showErrorMessages(action string, errorMessages map[string]error) error {
...
+	}
+
+	var uString string
+	// Check connection for satellite to create the final URL
+	base, err := url.Parse("https://" + serverHost)
+	if err != nil {
+		return nil, fmt.Errorf("cannot get base URL: %w", err)
+	}
+
+	if serverHost != "cdn.redhat.com" && serverHost != "cdn.stage.redhat.com" {
```
jirihnidek, line 142: Why do check the URL of CDN? Why don't you check `server.hostname` (default value is `subscription.rhsm.redhat.com`). Is it good idea to expose addresses of internal testing servers?

main.go
```diff
@@ -94,6 +104,108 @@ func showErrorMessages(action string, errorMessages map[string]error) error {
...
+	var profile ConfProfile
+
+	// Get the server hostname where this host is connected
+	var serverHost, err = getBaseURL()
+	if err != nil {
+		return nil, fmt.Errorf("cannot get server hostname: %w", err)
+	}
+
+	var uString string
+	// Check connection for satellite to create the final URL
```
jirihnidek, line 136: This checking connection to satellite server is wild. I'm sorry I don't like it.

I would just try to connect to server providing the profile. If there are more possible instances of the server providing profile, then there has to be some configuration file, where we should be able to read URL of the server.

util.go
```diff
@@ -51,3 +53,13 @@ func BashComplete(c *cli.Context) {
 		PrintFlagNames(c.App.VisibleFlags(), c.App.Writer)
 	}
 }
+
+func ConfigPath() (string, error) {
```
jirihnidek, line 57: Can you please add doc string for this function? It is not obvious what configuration file is actually read?

main.go
```diff
@@ -567,18 +726,37 @@ func main() {
...
 			Hidden:  false,
 			Value:   false,
 			EnvVars: []string{"NO_COLOR"},
 		},
+		&cli.StringFlag{
+			Name:      "config",
+			Hidden:    true,
+			Value:     defaultConfigFilePath,
+			TakesFile: true,
+			Usage:     "Read config values from `FILE`",
```
jirihnidek, line 740: What configuration file could be specified here? Is it rhsm.conf? Is it configuration for `rhc` or `rhcd` or `insights-client`?

jirihnidek (changes requested):
It is better. :+1: I have make questions, comments and requests.

main.go
ahitacat: What if this list grows? We will have then more lines added to the output, isn't it better to have the list instead of new lines?

main.go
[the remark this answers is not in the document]
ahitacat: Good catch, I'm changing this

conf.go
```diff
@@ -0,0 +1,72 @@
...
+const (
+	cliLogLevel = "log-level"
+	cliCertFile = "cert-file"
+	cliKeyFile  = "key-file"
+)
+
+type Conf struct {
+	CertFile string
+	KeyFile  string
+	LogLevel string
```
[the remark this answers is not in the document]
ahitacat: I added it because it was part of the configuration file. I think it can re-write the LogLevel of rhc . But this change seemed out of the scope of this PR.

I'll remove and create a ticket with this as a feature.

main.go
```diff
@@ -567,18 +726,37 @@ func main() {
...
+		},
+		altsrc.NewStringFlag(&cli.StringFlag{
+			Name:   cliCertFile,
```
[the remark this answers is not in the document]
ahitacat: These CLI options are used to configure the cert and key file path. This path is read from the configuration file but I added as a hidden hidden option as it can be use for debugging.  Same idea as [yggd ](https://github.com/RedHatInsights/yggdrasil/blob/main/cmd/yggd/main.go#L57)

main.go
ahitacat: This is meant for rhc configuration file

util.go
ahitacat: sure thing

main.go
ahitacat: That's true, I add this to print out nicely as a JSON (change the Active parameter to one that had more meaning). This is no longer used as the output isn't print as JSON but could be useful to print it with the `--status`. I will remove it because it doesn't belong to this PR.

conf.go
```diff
@@ -0,0 +1,72 @@
...
+	"path/filepath"
+)
+
+const (
+	cliLogLevel = "log-level"
+	cliCertFile = "cert-file"
+	cliKeyFile  = "key-file"
+)
+
+type Conf struct {
```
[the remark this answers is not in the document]
ahitacat: I will say yes, this is the structure of the config file

conf.go
```diff
@@ -0,0 +1,72 @@
...
+	KeyFile  string
+	LogLevel string
+}
+
+func (conf *Conf) CreateTLSClientConfig() (*tls.Config, error) {
+	var certData, keyData []byte
+	var err error
+	rootCAs := make([][]byte, 0)
+
+	const CADir string = "/etc/rhsm/ca/"
```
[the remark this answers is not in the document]
ahitacat: I'm changing this to get this conf from the D-Bus

main.go
ahitacat: I write this in one of the previous comments https://github.com/RedHatInsights/rhc/pull/29#discussion_r1090764015

ahitacat:
New pushes:

- Only hardcode the default API Server
- rhc will read the insights-client.conf file to get the default API Server (yes, we wanted to avoid read other config file but I also wanted to duplicate this value among configuration files). If this value is set it will make a call to this server.
- Otherwise it will get the value from rhsm, if it is the default it will make a call to the default server (which is hardcoded)
- If neither of this things happens, it is assumes to be connected to Satellite and uses the value given by rhsm as the server name.
- Log level now is get from rhc configuration
- CADir is get from rhsm configuration

main.go
```diff
@@ -541,6 +655,8 @@ func beforeAction(c *cli.Context) error {
 	return nil
 }
 
+var DefaultConfig = Conf{}
```
jirihnidek, line 658: Why is it called `DefaultConfig`? It suggest that it contains default values.

util.go
```diff
@@ -51,3 +55,52 @@ func BashComplete(c *cli.Context) {
...
+
+	return filePath, nil
+}
+
+func GuessAPIURL() (string, error) {
+	var uString string
+	var baseURL *url.URL
+	// First check if the server api is set in conf
+	if DefaultConfig.APIServer != "" {
+		base, err := url.Parse("https://" + DefaultConfig.APIServer)
```
jirihnidek, line 75: I have tried following values of `base_url` in `/etc/insights-client/insights-client.conf`:

```
# Just un-commented default value that I found on RHEL9
base_url=cert-api.access.redhat.com:443/r/insights
```

```
# Valid URL
base_url=https://cert.console.redhat.com
```

But only following worked for me: `base_url=cert.console.redhat.com`.

jirihnidek (changes requested):
Thanks for updates. Much better. :+1: I'm sorry, but I still have some requests and comments :smiley: ... I know, I'm horrible ;-).

main.go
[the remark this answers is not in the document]
subpop: `cliCertFile` is not a string; it's a `const` named `cliCertFile` with a value of `"cert-file"`. This is a convention [introduced in `yggd`](https://github.com/RedHatInsights/yggdrasil/blob/main/cmd/yggd/config.go#L12) to consolidate all CLI flag names. The end result is that the flag name is "`--cert-file`", even though the `Name:` field is given a variable rather than a string literal.

main.go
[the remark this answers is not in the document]
subpop: While this is true, I'd rather preserve the current convention and fix it deliberately in a separate PR than complicate this PR further. The module name of the rhc project is `github.com/redhatinsights/rhc`. We *can* change it to `rhc` if we desire, but we should do this in a separate PR.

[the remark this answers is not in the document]
jirihnidek: OK, fixing this in separate PR sounds good.

main.go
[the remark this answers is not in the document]
jirihnidek: Oh, I'm sorry. I overlooked this. BTW: Is there way how to display help with all hidden options and commands?

jirihnidek:
BTW: when I try to build `rhc` with `make`, then `make` complains that it cannot open: `/etc/insights-client/insights-client.conf`, which is strange, because Makefile has not been changed in this PR.

```
[jhnidek@thinkpad-t580 rhc]$ make
go build -buildmode=pie -ldflags "-X 'main.Version=0.2.1' -X 'main.ShortName=rhc' -X 'main.LongName=rhc' -X 'main.BrandName=rhc' -X 'main.PrefixDir=/usr/local' -X 'main.BinDir=/usr/local/bin' -X 'main.SbinDir=/usr/local/sbin' -X 'main.LibexecDir=/usr/local/libexec' -X 'main.SysconfDir=/usr/local/etc' -X 'main.DataDir=/usr/local/share' -X 'main.DatarootDir=/usr/local/share' -X 'main.ManDir=/usr/local/share/man' -X 'main.DocDir=/usr/local/share/doc' -X 'main.LocalstateDir=/usr/local/var' -X 'main.TopicPrefix=redhat/insights' -X 'main.Provider=Red Hat'" -o rhc .
go run -buildmode=pie -ldflags "-X 'main.Version=0.2.1' -X 'main.ShortName=rhc' -X 'main.LongName=rhc' -X 'main.BrandName=rhc' -X 'main.PrefixDir=/usr/local' -X 'main.BinDir=/usr/local/bin' -X 'main.SbinDir=/usr/local/sbin' -X 'main.LibexecDir=/usr/local/libexec' -X 'main.SysconfDir=/usr/local/etc' -X 'main.DataDir=/usr/local/share' -X 'main.DatarootDir=/usr/local/share' -X 'main.ManDir=/usr/local/share/man' -X 'main.DocDir=/usr/local/share/doc' -X 'main.LocalstateDir=/usr/local/var' -X 'main.TopicPrefix=redhat/insights' -X 'main.Provider=Red Hat'" . --generate-bash-completion > rhc.bash
go run -buildmode=pie -ldflags "-X 'main.Version=0.2.1' -X 'main.ShortName=rhc' -X 'main.LongName=rhc' -X 'main.BrandName=rhc' -X 'main.PrefixDir=/usr/local' -X 'main.BinDir=/usr/local/bin' -X 'main.SbinDir=/usr/local/sbin' -X 'main.LibexecDir=/usr/local/libexec' -X 'main.SysconfDir=/usr/local/etc' -X 'main.DataDir=/usr/local/share' -X 'main.DatarootDir=/usr/local/share' -X 'main.ManDir=/usr/local/share/man' -X 'main.DocDir=/usr/local/share/doc' -X 'main.LocalstateDir=/usr/local/var' -X 'main.TopicPrefix=redhat/insights' -X 'main.Provider=Red Hat'" . --generate-man-page > rhc.1
Fail to read file: open /etc/insights-client/insights-client.conf: permission denied
gzip -k -f rhc.1
go run -buildmode=pie -ldflags "-X 'main.Version=0.2.1' -X 'main.ShortName=rhc' -X 'main.LongName=rhc' -X 'main.BrandName=rhc' -X 'main.PrefixDir=/usr/local' -X 'main.BinDir=/usr/local/bin' -X 'main.SbinDir=/usr/local/sbin' -X 'main.LibexecDir=/usr/local/libexec' -X 'main.SysconfDir=/usr/local/etc' -X 'main.DataDir=/usr/local/share' -X 'main.DatarootDir=/usr/local/share' -X 'main.ManDir=/usr/local/share/man' -X 'main.DocDir=/usr/local/share/doc' -X 'main.LocalstateDir=/usr/local/var' -X 'main.TopicPrefix=redhat/insights' -X 'main.Provider=Red Hat'" . --generate-markdown > USAGE.md
Fail to read file: open /etc/insights-client/insights-client.conf: permission denied
rm rhc.1
```

main.go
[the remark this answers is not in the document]
subpop: Not that I'm aware of. We're using `urfave/cli`, and the last time I looked, the package didn't offer any such feature.

subpop:
Oh wow. That is because the `Makefile` runs the package as part of the build process in order to generate the man pages. If you strip out all the build flags, the command becomes: `go run . --generate-man-page > rhc.1` Being dependent on `/etc/insights-client/insights-client.conf` at build time is not a good thing.

conf.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+
+	tlsConfig := &tls.Config{}
+	cert, err := tls.X509KeyPair(certData, keyData)
+	if err != nil {
+		return nil, fmt.Errorf("cannot create key pair: %w", err)
+	}
+	tlsConfig.Certificates = []tls.Certificate{cert}
+
+	pool, err := x509.SystemCertPool()
```
jirihnidek, line 65: This is actually not correct. We do not want to load system CA certs. We only want to load CA certs from `conf.CADir`. Here is snapshot of code how to simplify:

```go
	caFiles, err := os.ReadDir(*caDir)
	if err != nil {
		return nil, fmt.Errorf("cannot read PEM files from CA directory: %w", err)
	}
	caCertPool := x509.NewCertPool()
	for _, file := range caFiles {
		fPath := filepath.Join(*caDir, file.Name())
		data, err := os.ReadFile(fPath)
		if err != nil {
			return nil, fmt.Errorf("cannot read CA PEM file %s : %w", fPath, err)
		}
		caCertPool.AppendCertsFromPEM(data)
	}
```

util.go
ahitacat: The default value in `insights-client` is not update, so is expected to not work

main.go
```diff
@@ -94,6 +103,62 @@ func showErrorMessages(action string, errorMessages map[string]error) error {
...
+	}
+
+	if err := json.Unmarshal(body, &profile); err != nil {
+		return nil, fmt.Errorf("cannot marshal HTTP response body: %w", err)
+	}
+
+	return &profile, nil
+}
+
+// showConfProfile shows tabed user profile gathered from API
```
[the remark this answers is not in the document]
ahitacat: I forgot to change this description

ahitacat:
Oops didn't realized that the beforeAction was used during the build. I'll move the read of this file when it needs to be used, so avoid to used on building process.

main.go
```diff
@@ -520,7 +611,30 @@ func mainAction(c *cli.Context) error {
...
+
+	DefaultConfig = Conf{
+		LogLevel: c.String(cliLogLevel),
+		CertFile: c.String(cliCertFile),
+		KeyFile:  c.String(cliKeyFile),
+	}
+
+	err := DefaultConfig.configureAPIServer()
+	if err != nil {
+		return err
```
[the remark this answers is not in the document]
ahitacat: The default value should be provided by rhsm if this is not set. I will change the approach, instead of storing this information in the Conf struct, when it tries to select the API server hostname, first it will try to read this configuration. 

That way it would not cause any conflicts while building rhc or if the config file doesn't exists.

register.go
```diff
@@ -220,3 +220,34 @@ func configureRHSM(serverURL string) error {
 
 	return nil
 }
+
+func getServerHostname() (string, error) {
```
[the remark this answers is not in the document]
ahitacat: It is if this string could be an empty one `""`, but I do not see why deal with pointers if is not needed.

util.go
```diff
@@ -51,3 +55,52 @@ func BashComplete(c *cli.Context) {
...
+		uString = base.ResolveReference(u).String()
+	} else {
+		// Get the server hostname where this host is connected
+		var serverHost, err = getServerHostname()
+		if err != nil {
+			return "", fmt.Errorf("cannot get server hostname: %w", err)
+		}
+		// Get the final api server url to make the call
+		// Check if it is the default api server
+		if strings.Contains(serverHost, "subscription.rhsm") {
```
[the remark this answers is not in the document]
ahitacat: I wanted to avoid hard-code the url, but you are right

register.go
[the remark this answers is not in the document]
subpop: I would only suggest returning a pointer to a type in the case of a complex data type like a struct. For simple types like strings, the zero value of a string ("") is fine to return in this case. As long as the value of the *error* is used to determine whether the routine failed or not, the resulting return value can be nil or the zero value of the data type.

internal/http/client.go
```diff
@@ -0,0 +1,33 @@
...
+			TLSClientConfig: tlsConfig,
+		},
+	}
+
+	return &Client{
+		client: client,
+	}, nil
+}
+
+func (c *Client) Get(url string) (*http.Response, error) {
```
jirihnidek, line 26: Just note. I'm working on the code that reads consumer/ent. certs/key and set up connections, keep connections, allow to do all HTTP requests (not only Get), debug print of HTTP requests, etc. Thus we will be able to share this code in the future as Go package.

jirihnidek (approved):
LGTM. Thanks for all updates. :+1:

conf.go
```diff
@@ -0,0 +1,70 @@
...
+	"fmt"
+	"os"
+	"path/filepath"
+)
+
+const (
+	cliLogLevel  = "log-level"
+	cliCertFile  = "cert-file"
+	cliKeyFile   = "key-file"
+	cliAPIServer = "base_url"
```
subpop, line 15: Flag names should be hyphenated. "base-url"

main.go
```diff
@@ -94,6 +103,62 @@ func showErrorMessages(action string, errorMessages map[string]error) error {
...
+	url, err := GuessAPIURL()
+	if err != nil {
+		return nil, err
+	}
+
+	res, err := client.Get(url)
+	if err != nil {
+		return nil, err
+	}
+	// Read the response body
```
subpop, line 126: Technically, this doesn't read the response body. This ensures that the response body (an `io.Reader`) is closed after opening by calling its `Close` method in the defer block of the function.

main.go
```diff
@@ -94,6 +103,62 @@ func showErrorMessages(action string, errorMessages map[string]error) error {
...
+	}
+
+	return &profile, nil
+}
+
+// showConfProfile shows a list of the system profile enabled features gathered from API
+// https://github.com/RedHatInsights/config-manager/blob/master/internal/http/v2/openapi.json
+func showConfProfile(p *ConfProfile) {
+	if p != nil {
+		s := reflect.ValueOf(p).Elem()
```
subpop, line 148: Wow. This is a whole lot of `reflect` code to just get some field names and true/false values. Maybe marshalling the JSON response into a `map[string]bool` would be better than a struct type.

```go
var profile map[string]bool
if err := json.Unmarshal(body, &profile); err != nil {
    ...
}
```

Then calling `showConfProfile` could just enumerate the map:

```go
for service, status := range profile {
    if status {
        fmt.Printf("%v, ", service)
    }
}
```

main.go
```diff
@@ -257,10 +323,35 @@ func connectAction(ctx *cli.Context) error {
...
+	if len(errorMessages) == 0 {
+		// Update the configuration file
+		config.CADir, _ = getCADir()
+		// Generate a new http client configured with mutual TLS certificate
+		tlsConfig, err := config.CreateTLSClientConfig()
+		if err != nil {
+			errorMessages[BrandName] = fmt.Errorf("cannot configure TLS: %w", err)
+		}
+		client, err := http.NewHTTPClient(tlsConfig)
+		if client != nil && err == nil {
```
subpop, line 336: This is an unusual pattern. This could check `if err != nil` like most function calls. Then proceed to use the `client` without checking whether it's nil or not. It's a programmer error (and should cause a crash) if the function NewHTTPClient returns a nil client and a nil error, so it's OK to proceed assuming client is not nil after checking the value of err. That's a long-winded way of saying:

```suggestion
		if err != nil {
		    ...
		}
		profile, err := getConfProfile(client)
```

register.go
```diff
@@ -373,3 +373,48 @@ func configureRHSM(serverURL string) error {
...
+		"server.hostname",
+		"").Store(&sname); err != nil {
+		return "", unpackRHSMError(err)
+	}
+
+	return sname, nil
+}
+
+// Call D-bus to get the CA directory from rhsm
+func getCADir() (string, error) {
```
subpop, line 401: This function and `getServerHostname` feel like they could be combined into a generic function that takes a config name and return argument:

```go
func getRHSMConfigOption(name string, val interface{}) error {
    conn, err := dbus.SystemBus()
    obj := conn.Object("com.redhat.RHSM1", "/com/redhat/RHSM1/Config")
    obj.Call("com.redhat.RHSM1.Config.Get", dbus.Flags(0), name, "").Store(val)
    return nil
}
```

util.go
```diff
@@ -51,3 +56,61 @@ func BashComplete(c *cli.Context) {
...
+	return filePath, nil
+}
+
+// Get the API server URL based on, insights-client.conf and rhsm.conf
+// This URL may differ from prod, stage and Satellite
+func GuessAPIURL() (string, error) {
+	var uString string
+	var baseURL *url.URL
+	// First check if the server api is set in insights conf
+	cfg, err := ini.Load("/etc/insights-client/insights-client.conf")
```
subpop, line 77: FWIW, I've written an [INI parsing package](https://git.sr.ht/~spc/go-ini/) that follows the Go standard `Marshal`/`Unmarshal` pattern. I never liked how this ini package doesn't follow that structure.

subpop (changes requested):

util.go
jirihnidek: It's really nice package, but it is not on GitHub and probably nobody knows about it.

subpop: I do mirror it to Github: https://github.com/subpop/go-ini

main.go
ahitacat: You are right, I put the comment before adding the lines. I have move it to the place the body is read.

main.go
ahitacat: I have modified this a bit. Instead of printing by service, I added everything as an array of string and uses the `strings` module to separate them by commas and then printed as a title (with the first letter in uppercase). This way the output looks like:

```
● Enabled console.redhat.com services: Remote Configuration, Insights, Remediations, Compliance
```

Instead of:
```
● Enabled console.redhat.com services: remote configuration, insights, remediations, compliance,
```

main.go
ahitacat: You are totally right!

register.go
ahitacat: Done, that makes more sense.

util.go
ahitacat: I have changed the ini-parser to the one created by @subpop. I prefer to use our tools than external modules as long as they are simpler. 
It is not as straightforward to use, I needed to create some structs to read the parameters, and also added to the struct some tags to read the hyphen parameter names.

subpop: Yea, you do have to define the file you're parsing as structs. I don't think I added support for unmarshalling into a map. I think it (currently) has to be a struct with struct tags.

main.go
ahitacat: I had to change this a bit because the `strings.Title` was deprecated. I opted to no add more complexity to the function and the output of this lines is:

```
● Enabled console.redhat.com services: remote configuration, insights, remediations, compliance
```

subpop:
I'm running this branch through some functional testing in a VM of mine. When I attempt to connect my host to stage with:

```
sudo go run ./ --cert-file /etc/pki/consumer/cert.pem --key-file /etc/pki/consumer/key.pem connect -o ###### -a ######### --server subscription.rhsm.stage.redhat.com
```

It prints an error from the rhc step:

```
rhc   Cannot get the user profile: Get "https://subscription.rhsm.stage.redhat.com/redhat_access/r/insights/platform/config-manager/v2/profiles/current": x509: certificate signed by unknown authority
```

Is this an error in the URL guessing detection?

subpop:
Worked fine in production. This LGTM.

PR #30, opened 2023-01-26
Enable content when registering with username & password

subpop:
When registering with username & password, include "enable_content" in the options map to ensure content gets enabled.

Fixes: RHBZ#2141454


jirihnidek (approved):
LGTM. We can implement other features described in the BZ in other PRs.

PR #31, opened 2023-02-01
feat: rhc renaming

ahitacat:
RHC stands for Remote Host configuration, change all old mentions to this acronym to meet this name.

Resolves: ESSNTL-4228

### DIFF
--- a/main.go
+++ b/main.go
@@ -228,7 +228,7 @@ func connectAction(ctx *cli.Context) error {
 	}
 
 	/* 4. Show footer message */
-	fmt.Printf("\nManage your Red Hat connector systems: https://red.ht/connector\n")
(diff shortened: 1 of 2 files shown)


subpop (approved):
LGTM

main.go
```diff
@@ -228,7 +228,7 @@ func connectAction(ctx *cli.Context) error {
 	}
 
 	/* 4. Show footer message */
-	fmt.Printf("\nManage your Red Hat connector systems: https://red.ht/connector\n")
+	fmt.Printf("\nManage your Remote Host Configuration systems: https://red.ht/connector\n")
```
subpop, line 231: Actually, I'm wondering if we should change this string to "Manage your connected systems: https://red.ht/connector".

ahitacat: sound much better

PR #32, opened 2023-02-02
fix: stdout redirection blocking terminal

DuckBoss:
When rhc is invoked non-interactively and required arguments
are not provided, user prompts from rhc block/hang the terminal.
This commit aims to check for this occurrence and return an error instead.
This issue was originally discussed by Matyas and Jiri in #24 

- Introduces two utility methods: `stdIsRedirected` and `readInput`.
- When an rhc command is invoked with redirected output, an error with exit code 1 is returned instead of blocking the terminal.
- Moved user input prompts to a universal `readInput` utility method which can handle prompt messages and sensitive user input to prevent repeated usage of non-interactive context checks.

Useful and informative article I found about this while trying to find a solution: [Here](https://rderik.com/blog/identify-if-output-goes-to-the-terminal-or-is-being-redirected-in-golang/)

### DIFF
--- a/main.go
+++ b/main.go
@@ -1,15 +1,12 @@
 package main
 
 import (
-	"bufio"
 	"encoding/json"
 	"fmt"
 	"os"
 	"strings"
 	"text/tabwriter"
 	"time"
-
-	"golang.org/x/term"
 
 	"git.sr.ht/~spc/go-log"
 
@@ -113,18 +110,17 @@ func registerRHSM(ctx *cli.Context) (string, error) {
 		if len(activationKeys) == 0 {
 			if username == "" {
 				password = ""
-				scanner := bufio.NewScanner(os.Stdin)
-				fmt.Print("Username: ")
-				_ = scanner.Scan()
-				username = strings.TrimSpace(scanner.Text())
+				username, err = readInput("Username: ", false)
+				if err != nil {
+					return "Unable to read username", cli.Exit(err, 1)
+				}
+				username = strings.TrimSpace(username)
 			}
 			if password == "" {
-				fmt.Print("Password: ")
-				data, err := term.ReadPassword(int(os.Stdin.Fd()))
+				password, err = readInput("Password: ", true)
 				if err != nil {
 					return "Unable to read password", cli.Exit(err, 1)
 				}
-				password = string(data)
 				fmt.Printf("\n\n")
 			}
 		}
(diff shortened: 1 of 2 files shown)


jirihnidek:
Thanks for the PR. :+1: The behavior of `rhc` is much better now as we can see here:

```
[root@localhost rhc]# ./rhc connect --username admin --password admin > ./foo.log
[root@localhost rhc]# echo $?
1
```

However, it is necessary to display output of redirected file (`foo.log`) to be able to see the reason of failure:

```
[root@localhost rhc]# cat foo.log 
Connecting localhost to Red Hat.

Available Organizations:
admin  snowwhite  donaldduck! Cannot connect to Red Hat Subscription Management
𐄂 Skipping connection to Red Hat Insights
𐄂 Skipping activation of rhc daemon

STEP  ERROR  
rhsm  cannot connect to Red Hat Subscription Management: non-interactive usage of rhc does not support interactive user input
```

I think we can do it better here. When it is not possible to be interactive, because `stdout` is redirected, then we should print some error to `stderr` and add user guidance how to provide required data using CLI options. When both `stdout` and `stderr` is redirected, then redirected outputs to some file should make sense.

More information here: https://clig.dev/#interactivity

Note: there is no line break after printing list of available organizations... Maybe we could provide list of available organizations in the error message, when cannot be interactive?

Using table for printing errors does not help here too.

It is interesting, that `subscription-manager` does not work very well too, when `stdout` is redirected. It print "Password:" to `stderr` and it reads input, but it prints "Organization:" to `stdout`.

Open questions (not only to @subpop):
* What should we do, when only `stdout` is redirected?
* Should we print "Username:", "Password:" and "Organization" to `stderr`, when only `stdout` is redirected?
* Should we use table for printing errors?
* Shouldn't we stop using tables at all according feedback from UX designer?

subpop:
>  What should we do, when only stdout is redirected?
> Should we print "Username:", "Password:" and "Organization" to stderr, when only stdout is redirected or should we terminate rhc and print some error message to stderr?

The CLI HIG doesn't explicitly declare what one *should* do when stdout is redirected. To me, if `rhc` is invoked as `rhc connect > file.log`, this is implicitly the same thing as `rhc --no-prompt connect`. `rhc` should exit with an error message informing the user how to correctly use the program in a non-interactive way.

> Should we use table for printing 
> Shouldn't we stop using tables at all according feedback from UX designer?

I agree, we should stop using tables. They sound like they cause accessibility issues that can easily be avoided.

jirihnidek:
OK, it seems that we agree that we do not want to use tables for displaying error messages.

I think that we should implement displaying error messages in different way 1) in different PR 2) we should do it according some UX design 3) we should fix displaying error messages first and  then we could continue with work on this PR.

Do you agree?

subpop:
Sure, seems like a reasonable approach to me.

PR #35, opened 2023-03-06
fix: not end with an error when retrieving the system profile

ahitacat:
This commit fixes the exit status when `rhc` fails to retrieve the system profile. When `rhc` cannot retrieve the profile it will debug this failure, but the execution will fail without any error.

`rhc` should not end with a 1 status if it is connected to `rhsm` and `insights`.

old `rhc` output:
```
rhc connect
Connecting localhost.localdomain to Red Hat.

● Activated the Remote Host Configuration daemon

STEP                       ERROR  
Remote Host Configuration  Cannot get the user profile: Get "https://subscription.rhsm.stage.redhat.com/redhat_access/r/insights/platform/config-manager/v2/profiles/current": x509: certificate signed by unknown authority

echo $?
1
```

fixed `rhc` output, with log-level configured to `debug`:

```
rhc connect
Connecting localhost.localdomain to Red Hat.

● This system is already connected to Red Hat Subscription Management
● Activated the rhc daemon
Cannot configure TLS: cannot create key pair: tls: failed to find certificate PEM data in certificate input, but did find a private key; PEM inputs may have been switched
Cannot get the user profile: cannot get system profile: 401 Unauthorized

STEP      DURATION  
rhsm      3ms       
insights  34.035s   
rhc       23ms   

echo $?
0
```

### DIFF
--- a/main.go
+++ b/main.go
@@ -37,6 +37,11 @@ const bwDisconnectedPrefix = "𐄂"
 const bwErrorPrefix = "!"
 const bwInfoPrefix = "·"
 
+type LogMessage struct {
+	level   log.Level
+	message error
+}
+
 // showProgress calls function and, when it is possible display spinner with
 // some progress message.
 func showProgress(progressMessage string, isColorful bool, function func() error) error {
@@ -86,17 +91,21 @@ func showTimeDuration(durations map[string]time.Duration) {
 }
 
 // showErrorMessages shows table with all error messages gathered during action
-func showErrorMessages(action string, errorMessages map[string]error) error {
-	if len(errorMessages) > 0 {
+func showErrorMessages(action string, errorMessages map[string]LogMessage) error {
+	if hasPriorityErrors(errorMessages, log.CurrentLevel()) {
 		fmt.Println()
 		fmt.Printf("The following errors were encountered during %s:\n\n", action)
 		w := tabwriter.NewWriter(os.Stdout, 0, 0, 2, ' ', 0)
-		_, _ = fmt.Fprintln(w, "STEP\tERROR\t")
-		for svc, err := range errorMessages {
-			_, _ = fmt.Fprintf(w, "%v\t%v\n", svc, err)
+		_, _ = fmt.Fprintln(w, "TYPE\tSTEP\tERROR\t")
+		for step, logMsg := range errorMessages {
+			if logMsg.level <= log.CurrentLevel() {
+				_, _ = fmt.Fprintf(w, "%v\t%v\t%v\n", logMsg.level, step, logMsg.message)
+			}
 		}
 		_ = w.Flush()
-		return cli.Exit("", 1)
+		if hasPriorityErrors(errorMessages, log.LevelError) {
+			return cli.Exit("", 1)
+		}
 	}
 	return nil
 }
@@ -261,7 +270,7 @@ func connectAction(ctx *cli.Context) error {
 
 	var start time.Time
 	durations := make(map[string]time.Duration)
-	errorMessages := make(map[string]error)
+	errorMessages := make(map[string]LogMessage)
 	hostname, err := os.Hostname()
 	if err != nil {
 		return cli.Exit(err, 1)
@@ -276,7 +285,10 @@ func connectAction(ctx *cli.Context) error {
 	var returnedMsg string
 	returnedMsg, err = registerRHSM(ctx)
 	if err != nil {
-		errorMessages["rhsm"] = fmt.Errorf("cannot connect to Red Hat Subscription Management: %w", err)
+		errorMessages["rhsm"] = LogMessage{
+			level: log.LevelError,
+			message: fmt.Errorf("cannot connect to Red Hat Subscription Management: %w",
+				err)}
 		fmt.Printf(errorPrefix + " Cannot connect to Red Hat Subscription Management\n")
 	} else {
 		fmt.Printf(connectedPrefix + " " + returnedMsg + "\n")
@@ -284,13 +296,18 @@ func connectAction(ctx *cli.Context) error {
 	durations["rhsm"] = time.Since(start)
 
 	/* 2. Register insights-client */
-	if _, exist := errorMessages["rhsm"]; exist {
-		fmt.Printf(disconnectedPrefix + " Skipping connection to Red Hat Insights\n")
+	if errors, exist := errorMessages["rhsm"]; exist {
+		if errors.level == log.LevelError {
+			fmt.Printf(disconnectedPrefix + " Skipping connection to Red Hat Insights\n")
+		}
 	} else {
 		start = time.Now()
 		err = showProgress(" Connecting to Red Hat Insights...", isColorful, registerInsights)
 		if err != nil {
-			errorMessages["insights"] = fmt.Errorf("cannot connect to Red Hat Insights: %w", err)
+			errorMessages["insights"] = LogMessage{
+				level: log.LevelError,
+				message: fmt.Errorf("cannot connect to Red Hat Insights: %w",
+					err)}
 			fmt.Printf(errorPrefix + " Cannot connect to Red Hat Insights\n")
 		} else {
 			fmt.Printf(connectedPrefix + " Connected to Red Hat Insights\n")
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -324,16 +324,17 @@ func connectAction(ctx *cli.Context) error {
 		// Generate a new http client configured with mutual TLS certificate
 		tlsConfig, err := config.CreateTLSClientConfig()
 		if err != nil {
-			errorMessages[BrandName] = fmt.Errorf("cannot configure TLS: %w", err)
+			log.Debug("Cannot configure TLS: ", err)
```
subpop, line 327: Do we really want to only log this output if the log level is greater than or equal to "debug"? This feels more like the sort of error we should log at the "error" level. Logging at the error level won't result in a change to the exit code, but it will ensure the error is reported at the default logging level.

subpop (changes requested):

main.go
ahitacat: I wanted to discuss this :smile:. I wasn't sure if putting the log level to error was a good idea or if it was better to use the "warning" level. This error doesn't imply the `rhc` has failed but it couldn't gather some information. 

I'm imagining the case there is a temporary error in the network connection that makes this call fail. If this error is logged in the "error" level this will be printed in the output (if the configuration file is left by default). But it will not be printed if the "warning" level is set.

I'm not sure what behavior we want for this example.

subpop: Error messages are only useful to users if they're actionable. Meaning, is there something the user can do in response to the error message to fix the problem. In this case, I doubt there is much they can do. Maybe we can log the TLS error and the HTTP client configuration error at a non-default log level (such as "warning"), and log a final error message if `getConfProfile` fails, at least informing the user that *something* that was supposed to work did not work.

ahitacat: That seems reasonable

subpop (approved):
LGTM

main.go
```diff
@@ -319,21 +319,21 @@ func connectAction(ctx *cli.Context) error {
...
 		client, err := http.NewHTTPClient(tlsConfig)
 		if err != nil {
-			errorMessages[BrandName] = fmt.Errorf("cannot configure HTTP client: %w", err)
+			log.Warn("Cannot configure HTTP client: ", err)
 		}
 		// Get the user profile
 		profile, err := getConfProfile(client)
 		if err != nil {
-			errorMessages[BrandName] = fmt.Errorf("Cannot get the user profile: %w", err)
+			log.Error("Cannot get the user profile: ", err)
```
jirihnidek, line 336: When it is possible to get profile and print results, then blue dot and message is printed. When it is not possible to get profile, then there is no red dot and no error/warning message. It is not consistent.

What is wrong about storing error in the map as we do for "rhsm", "insights-client" and "rhcd"? Why is getting profile inconsistent again? I can understand that it is not good to return some error code, when it is not possible to get information about profile, because system was just registered.

Printing all errors in table was probably not the best UX decision (as we learned here: https://github.com/RedHatInsights/rhc/pull/29#issuecomment-1401823780), but we will solve this later.

I can see that it easy to solve this issue with `log.Warn()` and `log.Error()`, but I think that we can do better.

I have proposal. Not store only error message in `errorMessage`, but store there also severity and later in `showErrorMessages()` function you can return `nil`, when only warnings were printed.

jirihnidek (changes requested):
I think that we can solve this issue in better way. I left one comment in the code.

main.go
ahitacat: Also, adding this as an error could be causing automatic playbooks/scripts to fail when `rhc` has correctly registered and `rhcd` is up and running. Only by not getting the profile, this is common situation for stage uses without proper configure `base_url`.

Adding those logs into errorMessage and then investigating inside that to know if the error is critical or not seems a bit over-engineering. I would like to have @subpop minds in here.

jirihnidek: > Wow I thought this PR was already merged :crying_cat_face:

Blame @subpop . He forgets to push merge button after approving PR :smile_cat:

subpop: Why do we collect errors in a map to begin with? That seems weird. Can we use a custom logger/printer function that prefixes the string with the correct dot and logs it when the error occurs? Or are we collecting then errors because it breaks the terminal animation?

jirihnidek: I think that the main reason is terminal animation... When we have some existing design pattern, then we should follow it. Why? Several reasons:

1. It looks better, when all parts of application are consistent.
2. If it decide to change terminal animation in the future, then it will be easier to redo, when all steps use same design pattern.

subpop: These are two very convincing points. I appreciate it when code follows the existing patterns within the application; it makes for easier long-term maintenance. It seems like perhaps we could store the log level value in the map as well; that would allow for consistent handling of all the errors.

ahitacat: I didn't want to blame anyone :laughing: 
Ok, this new approach makes sense for consistence. I'll be working on it (also add this as a jira ticket to track its status)

ahitacat:
## Latest changes

- Increased errorMessages to include the logLevel of that message.
  - Create a new struct to bind error and its level
  - Change showErrors to only print if there are error with the same priority or greater to the defined in configuration

I have used the Levels defined it go-log, instead of re-defined them.

New output:
```
Connecting localhost.localdomain to Red Hat.

● This system is already connected to Red Hat Subscription Management
● Activated the rhc daemon

STEP      DURATION   
rhsm      10ms       
insights  1m37.625s  
rhc       15ms       

TYPE  STEP  ERROR  
WARN  rhc   cannot get the user profile: cannot get system profile: 401 Unauthorized
```

main.go
```diff
@@ -37,6 +37,11 @@ const bwDisconnectedPrefix = "𐄂"
 const bwErrorPrefix = "!"
 const bwInfoPrefix = "·"
 
+type ErrorMessage struct {
```
jirihnidek, line 40: Maybe, I would rename this struct to `LogMessage`, because this structure can hold also warning and debug messages.

main.go
```diff
@@ -86,14 +91,16 @@ func showTimeDuration(durations map[string]time.Duration) {
...
+func showErrorMessages(action string, errorMessages map[string]ErrorMessage) error {
+	if hasErrorsToPrint(errorMessages) {
 		fmt.Println()
 		fmt.Printf("The following errors were encountered during %s:\n\n", action)
 		w := tabwriter.NewWriter(os.Stdout, 0, 0, 2, ' ', 0)
-		_, _ = fmt.Fprintln(w, "STEP\tERROR\t")
-		for svc, err := range errorMessages {
-			_, _ = fmt.Fprintf(w, "%v\t%v\n", svc, err)
+		_, _ = fmt.Fprintln(w, "TYPE\tSTEP\tERROR\t")
+		for svc, error := range errorMessages {
```
jirihnidek, line 100: Two notes:
* Local variable `error` collides with builtin `error`!
* What does `svc` stands for? I don't know it. If you don't know it too, then I would rename it something obvious like `id`.

jirihnidek (changes requested):
I like proposed change. Overall it is good now. I have few comments, suggestions and one request.

main.go
ahitacat: svc indicates the step: `rhsm`, `insights-client` or `rhc` maybe I can change it to step instead.

ahitacat:
# Latest Changes

- Change name variables as @jirihnidek suggested.
- Wrap long lines
- In showErrorMessages, only exit with 1 if an error with level `ERROR` has been reported. 

Old behavior.

```
$ sudo ./rhc connect
Connecting localhost.localdomain to Red Hat.
[...]
TYPE  STEP  ERROR  
WARN  rhc   cannot get the user profile: cannot get system profile: 401 Unauthorized
[ahitacat@localhost rhc]$ echo $?
1
```

New behavior
```
$ sudo ./rhc connect
Connecting localhost.localdomain to Red Hat.
[...]
TYPE  STEP  ERROR  
WARN  rhc   cannot get the user profile: cannot get system profile: 401 Unauthorized
[ahitacat@localhost rhc]$ echo $?
0
```

main.go
```diff
@@ -86,17 +91,21 @@ func showTimeDuration(durations map[string]time.Duration) {
...
+func showErrorMessages(action string, errorMessages map[string]LogMessage) error {
+	if hasPriorityErrors(errorMessages, log.CurrentLevel()) {
 		fmt.Println()
 		fmt.Printf("The following errors were encountered during %s:\n\n", action)
 		w := tabwriter.NewWriter(os.Stdout, 0, 0, 2, ' ', 0)
-		_, _ = fmt.Fprintln(w, "STEP\tERROR\t")
-		for svc, err := range errorMessages {
-			_, _ = fmt.Fprintf(w, "%v\t%v\n", svc, err)
+		_, _ = fmt.Fprintln(w, "TYPE\tSTEP\tERROR\t")
+		for step, er := range errorMessages {
```
jirihnidek, line 100: Why `er` and not `err`? The `err` is convention. I think.

util.go
```diff
@@ -132,3 +133,14 @@ func GuessAPIURL() (string, error) {
 
 	return uString, nil
 }
+
+// hasPriorityErrors checks if the errorMessage map has any error
+// with a higher priority than the logLevel configure.
+func hasPriorityErrors(errorMessages map[string]LogMessage, level log.Level) bool {
+	for _, er := range errorMessages {
```
jirihnidek, line 140: Same comment here. Why `er`, when `err` is used on so many places?

main.go
ahitacat: Exactly because of that, I wasn't sure to use err that is a convention that represents an error, to represent the LogMessage struct.

jirihnidek (approved):
LGTM, Thanks for updates.

PR #37, opened 2023-05-03
Make connection and disconnection faster

jirihnidek:
* I believe that we can just trigger "insights-client --register" and "insights-client --unregister" and there is no need to wait for result
* TODO: I think that we can run some jobs in parallel: starting rhcd, insights-client and getting profile.


jirihnidek:
We will solve this issue hopefully in different way soon.

PR #41, opened 2023-07-27
fix: Fix building "rhc" using make

jirihnidek:
* It is possible to build rhc using `make`
* It is also possible to clean repository using `make clean`, when some generated file has been already deleted.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -148,8 +148,8 @@ dist:
 .PHONY: clean
 clean:
 	go mod tidy
-	rm $(BIN)
-	rm $(DATA)
+	rm -f $(BIN)
+	rm -f $(DATA)
 
 .PHONY: tests
 tests:
--- a/Makefile
+++ b/Makefile
@@ -67,8 +67,7 @@ BUILDFLAGS += -mod=vendor
 endif
 
 BIN  = rhc
-DATA = rhc.bash \
-	   rhc.1.gz \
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -148,8 +147,8 @@ dist:
 .PHONY: clean
 clean:
 	go mod tidy
-	rm $(BIN)
-	rm $(DATA)
+	rm -f $(BIN)
```
subpop, line 150: Maybe adding the `-` prefix to these lines so they don't cause `make` to exit if the command exits with a non-zero exit code?

jirihnidek: Interesting: https://www.gnu.org/software/make/manual/make.html#index-_002d-_0028in-recipes_0029, but it still prints errors about non-existing files to `stdout`, and I think we want to suppress printing errors too.

subpop: I think you'd have to redirect stderr to /dev/null in that case.

jirihnidek: I know, but I consider current solution as more straightforward.

jirihnidek: @subpop Hi, are you requesting your proposal or is my solution good enough?

subpop: Oh. This is fine with me. I'm not concerned with suppressing output.

subpop (approved):

PR #50, opened 2023-10-02
fix: Fix issues of bash auto-complete script

jirihnidek:
* Card ID: RHELPLAN-164718
* Checking of issues using shell script analysis tool shellcheck should be without any issue now.
* Note: current version of bash auto-complete script provided by urfave package looks little bit different, but it is not compatible with rhc for some reason.

### DIFF
--- a/rhc.bash
+++ b/rhc.bash
@@ -1,21 +1,22 @@
 #! /bin/bash
 
-: ${PROG:=$(basename ${BASH_SOURCE})}
+: "${PROG:=$(basename "${BASH_SOURCE[0]}")}"
 
 _cli_bash_autocomplete() {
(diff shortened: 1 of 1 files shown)


ahitacat:
What is the compatibility issue with the current version of bash-completion script?

jirihnidek:
> What is the compatibility issue with the current version of bash-completion script?

Well, when I tried to use current version of `rhc` with the latest version of bash script provided by `urfave` go package, then it did not work and caused some errors.

If the main branch of `urfave` packages was used, then we would probably not see such errors. Main branch of `urfave` just does not play with current `rhc`. But I can see similar issues in the latest script provided by `urfave`. If I will have time I will try to create similar PR against `urfave`.

ahitacat (approved):
This looks good to me.

PR #55, opened 2023-11-07
fix: Upgrade yaml module to version v2.4.0

jirihnidek:
* Dependabot opened two issues related to yaml module:
  * https://github.com/RedHatInsights/rhc/security/dependabot/6
  * https://github.com/RedHatInsights/rhc/security/dependabot/5
* Dependabot was not able to create PR automatically for some unknown reason
* This commit should fix them both


ahitacat (approved):
LGTM

PR #69, opened 2023-11-15
fix: Set locale in all D-Bus RHSM calls according LANG env. var.

jirihnidek:
* It is important to call all RHSM D-Bus methods with some locale string, because candlepin server generates some error string, and candlepin server has localized these error messages. Thus, candlepin server needs to know locale string to generate error messages properly.
* All methods were called with locale hardcoded to "". It means that error messages could not be localized according users' preferences.
* Locale string is get from environment variable $LANG.

### DIFF
--- a/util.go
+++ b/util.go
@@ -144,3 +144,12 @@ func hasPriorityErrors(errorMessages map[string]LogMessage, level log.Level) boo
 	}
 	return false
 }
+
+// getLocale tries to get current locale
+func getLocale() string {
+	// FIXME: Locale should be detected in more reliable way. We are going to support
+	//        localization in better way. Maybe we could use following go module
+	//        https://github.com/Xuanwo/go-locale. Maybe some other will be better.
+	locale := os.Getenv("LANG")
+	return locale
+}
--- a/register.go
+++ b/register.go
@@ -16,13 +16,15 @@ func getConsumerUUID() (string, error) {
 		return "", err
 	}
 
+	locale := getLocale()
+
 	var uuid string
 	if err := conn.Object(
 		"com.redhat.RHSM1",
 		"/com/redhat/RHSM1/Consumer").Call(
 		"com.redhat.RHSM1.Consumer.GetUuid",
 		dbus.Flags(0),
-		"").Store(&uuid); err != nil {
+		locale).Store(&uuid); err != nil {
 		return "", unpackRHSMError(err)
 	}
 	return uuid, nil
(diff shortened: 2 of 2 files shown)


jirihnidek:
To test this bug fix or new functionality you can do following:

```
LANG=de_DE.UTF-8 ./rhc connect --username wrong_username --password wrong_password --organization wrong_org
Connecting thinkpad-p1 to Red Hat.

! Cannot connect to Red Hat Subscription Management
● Skipping connection to Red Hat Insights
● Skipping activation of yggdrasil service

TYPE   STEP  ERROR  
ERROR  rhsm  cannot connect to Red Hat Subscription Management: error: Ungültige Berechtigungsnachweise
```

... as you can see the error message at the end is translated to German (this message is generated by candlepin).

util.go
```diff
@@ -144,3 +144,9 @@ func hasPriorityErrors(errorMessages map[string]LogMessage, level log.Level) boo
 	}
 	return false
 }
+
+// getLocale tries to get current locale
+func getLocale() string {
+	locale := os.Getenv("LANG")
```
jirihnidek, line 150: This should be more complicated. Not sure about using https://github.com/Xuanwo/go-locale

ptoscano: Yes, there is a list of environment variables that gettext checks: https://www.gnu.org/software/gettext/manual/html_node/The-LANGUAGE-variable.html

Also, please note that `LANGUAGE` can be a colon-separated list.

m-horky: I think this is enough for the current use-case. We'll need to write more once we support localization of rhc itself; until then I'm fine with this solution.

The scope of this PR is to send the locale over the D-Bus, which works. Proper POSIX-compliant locale detection is out of scope of the PR in my opinion.

m-horky (approved):
LGTM. Will merge after the one comment thread is ACKed by the participants.

util.go
jirihnidek: It seems that `go-locale` supports all env. vars. described in https://www.gnu.org/software/gettext/manual/html_node/The-LANGUAGE-variable.html. The `go-locale` also supports every possible OS from `AIX` to `z/OS`. Maybe, it is overkill. Not sure.

We are going to support localization of `rhc` in better way. Some other go module used for localization could possibly provide similar/same functionality as `go-locale`.

I think that it would be good to defer decision how to detect locale for this case after proper localization is implemented for `rhc`.

I will add `FIXME` comment to the code, and open issue or card. Sounds good?

m-horky (approved):

PR #70, opened 2023-11-15
fix: Do not trigger build go actions twice on PR

jirihnidek:
* When PR was created or updated, then all jobs releated to Go were triggered twice. Not useful and wasting of time and resources
* Added workflow_dispatch to be able to trigger workflows manually


ahitacat (approved):
LGTM

PR #74, opened 2023-12-04
build: unbundle yggdrasil & worker-package-manager

ptoscano:
Make use of yggdrasil and the package manager worker as available from the system, rather than bundling static versions of them.

This makes it possible to build & use rhc against development versions of them.

### DIFF
--- a/.copr/Makefile
+++ b/.copr/Makefile
@@ -40,10 +40,8 @@ srpm: deps
 		$(spec)/.copr/rhc.spec.in > $(spec)/rhc.spec
 	make PKGNAME=$(PKGNAME) VERSION=$(VERSION)-$(RELEASE) dist
 	install -D -m644 $(PKGNAME)-$(VERSION)-$(RELEASE).tar.gz $(SOURCEDIR)
-	spectool --get-files --source 1 --sourcedir $(spec)/rhc.spec
-	spectool --get-files --source 2 --sourcedir $(spec)/rhc.spec
 	rpmbuild -bs $(spec)/rhc.spec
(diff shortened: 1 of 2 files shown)


ahitacat:
I have a concern here. 
We should have to make rhc depend on a certain version of yggdrasil. Because actual rhc will not work with yggdrasil 0.3, same happens with the workers.

ptoscano:
>  We should have to make rhc depend on a certain version of yggdrasil. Because actual rhc will not work with yggdrasil 0.3, same happens with the workers.

makes sense - bumped the `yggdrasil` BR to 0.4+

subpop:
What kind of dependency does `rhc` really have on `yggdrasil`? It doesn't use the `com.redhat.Yggdrasil1` D-Bus API yet. I believe all it expects is to find a systemd unit called `rhcd.service`. No version of `yggdrasil` will provide that unit file.

jirihnidek:
Current upstream version of rhc tries to find `yggdrasil.service`. So, I would say that upstream version of `rhc` has dependency on upstream version of `yggdrasil`. This rpm build is intended for upstream version of `rhc`. Thus, I would say that this dependency of `rhc` on `yggdrasil` is valid.

jirihnidek:
@ptoscano I just merged another PR to main branch. Can you please resolve the conflict?

ahitacat (approved):
I think my concerns have been resolved. LGTM. 
I'm leaving some time if @subpop or @jirihnidek  wants to approve this before merge it.

subpop (approved):

PR #75, opened 2023-12-05
feat: Better handling of colors and icons

m-horky:
This patch simplifies the way we handle and display status information about individual things rhc tracks.

1/ Message icons
Colors, icons and statuses were simplified. The (red) state of    'disconnected' was removed and was replaced by error/info depending on the context (e.g. 'status' reports disabled yggdrasil/rhcd at 'info' level; it is a valid state for some customer setups, it should not be in red).

2/ Message colors
Blue (info) color was switched for yellow, which is more common for green (ok) - info - error (red) scheme.

3/ Representation of the preference data
`UserInterfacePreferences` struct was replaced by `userInterfaceSettings`, which is set up once as a global variable `uiSettings` and is reused throughout the code, instead of being re-computed multiple times throughout the code.

4/ '--format' handling of 'status'
Previously, the 'status' action has been modifying the `uiSettings` when it was run. Now, all preparation is done via command's `Before`, making it self-contained.

In the future, when we split the main file into smaller self-contained blocks, we could do something like `if output.IsRich()` instead of accessing the struct members directly. This would ensure that the global state we set up is not modified at actual runtime.

### DIFF
--- a/main.go
+++ b/main.go
@@ -19,47 +19,67 @@ import (
 	"github.com/urfave/cli/v2/altsrc"
 )
 
-const redColor = "\u001B[31m"
-const greenColor = "\u001B[32m"
-const blueColor = "\u001B[36m"
-const endColor = "\u001B[0m"
-
-// Colorful prefixes
-const ttyConnectedPrefix = greenColor + "●" + endColor
-const ttyDisconnectedPrefix = redColor + "●" + endColor
-const ttyInfoPrefix = blueColor + "●" + endColor
-const ttyErrorPrefix = redColor + "!" + endColor
-
-// Black & white prefixes. Unicode characters
-const bwConnectedPrefix = "✓"
-const bwDisconnectedPrefix = "𐄂"
-const bwErrorPrefix = "!"
-const bwInfoPrefix = "·"
-
 type LogMessage struct {
 	level   log.Level
 	message error
 }
 
-// UserInterfacePreferences helps to hold UI preferences
-type UserInterfacePreferences struct {
-	IsColorful         bool
-	MachineReadable    bool
-	ConnectedPrefix    string
-	DisconnectedPrefix string
-	ErrorPrefix        string
-	InfoPrefix         string
+const (
+	colorGreen  = "\u001B[32m"
+	colorYellow = "\u001B[33m"
+	colorRed    = "\u001B[31m"
+	colorReset  = "\u001B[0m"
+)
+
+// userInterfaceSettings manages standard output preference.
+// It tracks colors, icons and machine-readable output (e.g. json).
+//
+// It is instantiated via uiSettings by calling configureUISettings.
+type userInterfaceSettings struct {
+	// isMachineReadable describes the machine-readable mode (e.g., `--format json`)
+	isMachineReadable bool
+	// isRich describes the ability to display colors and animations
+	isRich    bool
+	iconOK    string
+	iconInfo  string
+	iconError string
+}
+
+// uiSettings is an instance that keeps actual data of output preference.
+//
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -19,47 +19,60 @@ import (
 	"github.com/urfave/cli/v2/altsrc"
 )
 
-const redColor = "\u001B[31m"
-const greenColor = "\u001B[32m"
-const blueColor = "\u001B[36m"
-const endColor = "\u001B[0m"
```
jirihnidek, line 25: I would not remove these constants. Readability of following code will be only worse.

main.go
```diff
@@ -19,47 +19,60 @@ import (
...
+			isMachineReadable: false,
+			iconOK:            "✓",
+			iconInfo:          "·",
+			iconError:         "𐄂",
+		}
+	} else {
+		uiSettings = userInterfaceSettings{
+			isRich:            true,
+			isMachineReadable: false,
+			iconOK:            "\u001B[32m●\u001B[0m",
```
jirihnidek, line 61: As you can see here... Nobody knows what colors are used here. You can see here only some cryptic strings. Please use constants.

main.go
```diff
@@ -542,29 +521,19 @@ func statusAction(ctx *cli.Context) (err error) {
...
-			)
-			return cli.Exit(err, 1)
-		}
+	switch format {
+	case "":
+		break
+	case "json":
+		machineReadablePrintFunc = printJSONStatus
+	default:
+		return fmt.Errorf("unsupported format: %s", format)
```
jirihnidek, line 531: It seems that this code path could not happen, because unsupported formats are catched in `Before` function.

main.go
```diff
@@ -801,7 +773,27 @@ func main() {
 			Usage:       "Prints status of the system's connection to " + Provider,
 			UsageText:   fmt.Sprintf("%v status", app.Name),
 			Description: fmt.Sprintf("The status command prints the state of the connection to Red Hat Subscription Management, Red Hat Insights and %v.", Provider),
-			Action:      statusAction,
+			Before: func(ctx *cli.Context) error {
```
jirihnidek, line 776: Please convert this anonymous function to something like `beforeAction()`.

main.go
```diff
@@ -19,47 +19,60 @@ import (
...
+var uiSettings = userInterfaceSettings{}
+
+// configureUISettings is called by the CLI library when it loads up.
+// It sets up the uiSettings object.
+func configureUISettings(ctx *cli.Context) {
+	if ctx.Bool("no-color") {
+		uiSettings = userInterfaceSettings{
+			isRich:            false,
+			isMachineReadable: false,
+			iconOK:            "✓",
```
jirihnidek, line 53: We also received UX feedback saying that these cool icons are not so cool for blind people using screen readers. Maybe we should replace it with something readable by screen readers. I'm not telling that it should be part of this PR, but the name `iconOK` will likely have to change in the future.

jirihnidek (changes requested):
Thanks for the PR. Overall it looks good. I have few comments and requests.

main.go
m-horky: My idea was to handle it everywhere, because you never know when you'll forget to update one of those places. But sure, I can flip the logic around.

main.go
m-horky: ACK. I have just checked with Orca:
- the colored options are all read as "circle"
- the `--no-color` options are skipped (𐄂, !) or read aloud (✓).

I'll put this to my backlog, I have other items that would benefit from external UX review. I don't consider this the final solution, only an improvement to make it easier to translate later.

jirihnidek (approved):
LGTM, Thanks for updates :+1:

PR #81, opened 2023-12-20
Update go-ini import path

subpop:
I've moved go-ini to github, hopefully to facilitate external contributors. Update rhc to use the package hosted on github.

PR #89, opened 2024-01-12
refactor: update go-log path

ahitacat:
Update to github.com/subpop/go-log.

This change was motivate because of SourceHut service and its stability. We also change the go-ini to github https://github.com/RedHatInsights/rhc/pull/81


jirihnidek (approved):
LGTM

PR #90, opened 2024-01-15
fix: Try to fix copr build for rhel9 using copr.

jirihnidek:
* Change like this could be used for building `rhc` for RHEL9 and RHEL8 using copr.
  * https://copr.fedorainfracloud.org/coprs/jhnidek/rhc/build/6905770/

### DIFF
--- a/.copr/rhc.spec.in
+++ b/.copr/rhc.spec.in
@@ -19,14 +19,20 @@ Source0: %{name}-%{version}-@RELEASE@.tar.gz
 ExclusiveArch: %{go_arches}
 
 Requires:      insights-client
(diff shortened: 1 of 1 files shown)


jirihnidek:
Maybe it will be possible to build it without this hack, but it will require to do some changes of copr settings.

jirihnidek:
This will not be needed as you can see here: https://copr.fedorainfracloud.org/coprs/jhnidek/rhc/build/6907485/ Problem is only fedora rawhide, because `$releasever` is set to 40, but we use pattern `fedora-rawhide-$basearch` here: https://download.copr.fedorainfracloud.org/results/@yggdrasil/latest/ (not ``fedora-$releasever-$basearch` for rawhide).

PR #97, opened 2024-02-21
fix: Print correct error, when non-root user run "rhc status"

jirihnidek:
* Card ID: CCT-288
* When "rhc status" is run as non-root user, then status about insighst-client should not tell that insights-client is not registered.
* When "rhc status" is run by non-root user, then rhc prints decent error about insights client. It prints: "Insights client must be run as root."
* Also fixed the case, when status command is run with "--format json". When some error happens, then generated JSON contains real error message and not empty dictionary.

### DIFF
--- a/insights.go
+++ b/insights.go
@@ -1,7 +1,11 @@
 package main
 
 import (
+	"bytes"
+	"errors"
+	"fmt"
 	"os/exec"
+	"strings"
 )
 
 func registerInsights() error {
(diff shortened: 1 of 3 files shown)


insights.go
```diff
@@ -17,15 +20,27 @@ func unregisterInsights() error {
...
 	err := cmd.Run()
 
 	if err != nil {
 		// When the error is ExitError, then we know that insights-client only returned
 		// some error code not equal to zero. We do not care about error number.
-		if _, ok := err.(*exec.ExitError); ok {
-			return false, nil
+		var exitError *exec.ExitError
+		if errors.As(err, &exitError) {
+			// When stderr is empty, then we know that
```
subpop, line 34: This looks like a mistakenly copy/paste comment?

insights.go
```diff
@@ -17,15 +20,27 @@ func unregisterInsights() error {
...
+			stdErr := errBuffer.String()
+			if len(stdErr) == 0 {
+				return false, nil
+			} else {
+				if stdErr[len(stdErr)-1] == '\n' {
+					stdErr = stdErr[0 : len(stdErr)-1]
+				}
+				return false, fmt.Errorf("%s", stdErr)
```
subpop, line 42: Couldn't [`strings.TrimSpace`](https://pkg.go.dev/strings#TrimSpace) accomplish this in less code?

```suggestion
				return false, fmt.Errorf("%s", strings.TrimSpace(stdErr))
```

subpop (changes requested):

subpop (approved):
OK. LGTM!

PR #108, opened 2024-05-02
feat: Add CLI option --format to disconnect command

jirihnidek:
* It is possible to print output of disconnect command in machine-readable format. The --format requires argument and it could be only "json" ATM. It follows practice of status command.
* Refactored code a little to share some code with status command.

### DIFF
--- a/main.go
+++ b/main.go
@@ -298,10 +298,11 @@ func connectAction(ctx *cli.Context) error {
 			message: fmt.Errorf("cannot connect to Red Hat Subscription Management: %w",
 				err)}
 		fmt.Printf(
-			uiSettings.iconError + " Cannot connect to Red Hat Subscription Management\n",
+			"%v Cannot connect to Red Hat Subscription Management\n",
+			uiSettings.iconError,
 		)
 	} else {
-		fmt.Printf(uiSettings.iconOK + " " + returnedMsg + "\n")
+		fmt.Printf("%v %v\n", uiSettings.iconOK, returnedMsg)
 	}
 	durations["rhsm"] = time.Since(start)
 
@@ -309,7 +310,8 @@ func connectAction(ctx *cli.Context) error {
 	if errors, exist := errorMessages["rhsm"]; exist {
 		if errors.level == log.LevelError {
 			fmt.Printf(
-				uiSettings.iconError + " Skipping connection to Red Hat Insights\n",
+				"%v Skipping connection to Red Hat Insights\n",
+				uiSettings.iconError,
 			)
 		}
 	} else {
@@ -320,9 +322,9 @@ func connectAction(ctx *cli.Context) error {
 				level: log.LevelError,
 				message: fmt.Errorf("cannot connect to Red Hat Insights: %w",
 					err)}
-			fmt.Printf(uiSettings.iconError + " Cannot connect to Red Hat Insights\n")
+			fmt.Printf("%v Cannot connect to Red Hat Insights\n", uiSettings.iconError)
 		} else {
-			fmt.Printf(uiSettings.iconOK + " Connected to Red Hat Insights\n")
+			fmt.Printf("%v Connected to Red Hat Insights\n", uiSettings.iconOK)
 		}
 		durations["insights"] = time.Since(start)
 	}
@@ -331,7 +333,8 @@ func connectAction(ctx *cli.Context) error {
 	if errors, exist := errorMessages["rhsm"]; exist {
 		if errors.level == log.LevelError {
 			fmt.Printf(
-				uiSettings.iconError+" Skipping activation of %v service\n",
+				"%v Skipping activation of %v service\n",
+				uiSettings.iconError,
 				ServiceName,
 			)
 		}
@@ -344,9 +347,9 @@ func connectAction(ctx *cli.Context) error {
 				level: log.LevelError,
 				message: fmt.Errorf("cannot activate %s service: %w",
 					ServiceName, err)}
-			fmt.Printf(uiSettings.iconError+" Cannot activate the %v service\n", ServiceName)
+			fmt.Printf("%v Cannot activate the %v service\n", uiSettings.iconError, ServiceName)
 		} else {
-			fmt.Printf(uiSettings.iconOK+" Activated the %v service\n", ServiceName)
+			fmt.Printf("%v Activated the %v service\n", uiSettings.iconOK, ServiceName)
 		}
 		durations[ServiceName] = time.Since(start)
 	}
@@ -379,7 +382,7 @@ func connectAction(ctx *cli.Context) error {
 				message: fmt.Errorf("cannot get the user profile: %w",
 					err)}
 		} else {
-			fmt.Printf(uiSettings.iconInfo + " Enabled console.redhat.com services: ")
+			fmt.Printf("%v Enabled console.redhat.com services: ", uiSettings.iconInfo)
 			showConfProfile(&profile)
 			fmt.Printf("\n")
 		}
@@ -517,13 +520,13 @@ func disconnectAction(ctx *cli.Context) error {
 		errMsg := fmt.Sprintf("Cannot deactivate %s service: %v", ServiceName, err)
 		errorMessages[ServiceName] = LogMessage{
 			level:   log.LevelError,
-			message: fmt.Errorf(errMsg)}
+			message: fmt.Errorf("%v", errMsg)}
 		disconnectResult.YggdrasilStopped = false
 		disconnectResult.YggdrasilStoppedError = errMsg
-		interactivePrintf(uiSettings.iconError + " " + errMsg + "\n")
+		interactivePrintf("%v %v\n", uiSettings.iconError, errMsg)
 	} else {
 		disconnectResult.YggdrasilStopped = true
-		interactivePrintf(uiSettings.iconOK+" Deactivated the %v service\n", ServiceName)
+		interactivePrintf("%v Deactivated the %v service\n", uiSettings.iconOK, ServiceName)
 	}
 	durations[ServiceName] = time.Since(start)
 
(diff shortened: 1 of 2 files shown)


subpop:
I like the use of the `Before` function handler. That's a good idea to validate flags and user input!

jirihnidek:
BTW: I would like to add `--format` CLI option to `connect` command too, but I would like to do some refactoring of `main.go` first. This file is just too big and I would like to split it into `connect.go` and `disconnect.go`. Each command should be in separate file.

jirihnidek:
I agree that it would be good to have `--format json` as top-level flag, but I would like to have support for machine-readable output for commands. Should I implement `--format json` for `connect` command as part of this PR?

What bothers me is a fact that we will break compatibility with `rhc status --format json`.

TBH: I don't like the behavior of github.com/urfave/cli that you have to use flags of parent command immediately after main command. This is something unusual.

There is alfa version (v3) of github.com/urfave/cli (we still use v2) ... I will try it if the behavior is still the same in v3.

jirihnidek:
The v3 support persistent flags introduces in this PR: https://github.com/urfave/cli/pull/1568

The code looks like this:

```go
package main

import (
	"context"
	"fmt"
	"github.com/urfave/cli/v3"
	"log"
	"os"
)

func main() {
	cmd := cli.Command{
		Name: "rhc",
		Commands: []*cli.Command{
			{
				Name: "connect",
				Flags: []cli.Flag{
					&cli.StringFlag{
						Name: "username",
					},
				},
				Action: func(ctx context.Context, command *cli.Command) error {
					fmt.Printf("username: %s\n", command.String("username"))
					fmt.Printf("format: %s\n", command.String("format"))
					return nil
				},
			},
		},
		Flags: []cli.Flag{
			&cli.StringFlag{
				Name:       "format",
				Persistent: true,
			},
		},
	}

	err := cmd.Run(context.Background(), os.Args)
	if err != nil {
		log.Fatal(err)
	}
}
```

If flag is marked as persistent, then you can use global flag after sub-command like this:

```
jhnidek@thinkpad-p1:~/tmp/go_cli$ ./main --format json connect --username pepa
username: pepa
format: json
jhnidek@thinkpad-p1:~/tmp/go_cli$ ./main connect --username pepa --format json
username: pepa
format: json
jhnidek@thinkpad-p1:~/tmp/go_cli$ ./main --format xml connect --username pepa --format json
username: pepa
format: json
```

Unfortunately this feature hasn't been back-ported to v2.

subpop:
Oh, that's exactly what we should use. Let's proceed with separate `--format` flags for each subcommand, and when we update to cli/v3, we can refactor the code to use a single, persistent flag. That way we can support `--format` for the subcommands that need them and maintain compatibility with our CLI when we do eventually refactor.

main.go
```diff
@@ -401,49 +401,175 @@ func connectAction(ctx *cli.Context) error {
...
+		machineReadablePrintFunc = printJSONDisconnectResult
+	default:
+		break
+	}
+
+	// When printing of status is requested, then print machine-readable file format
+	// at the end of this function
+	if uiSettings.isMachineReadable {
+		defer func(disconnectResult *DisconnectResult) {
+			err := machineReadablePrintFunc(disconnectResult)
```
subpop, line 477: With `DisconnectResult` conforming to `error`, it could always be printed.
There's no need for a `machineReadablePrintFunc` in that case. Then the `defer`
function can always print it.

subpop (changes requested):

subpop (commented):
I noticed a significant amount of conditional checks on the value of `uiSettings.isMachineReadable`, and some data structure transformation between `DisconnectResult` and `error`. I recommend looking into a way to make `DisconnectResult` conform to the `error` interface. I think that could result in reducing a lot of the conditional checks, and might streamline some of the code paths that are trying to transform a `DisconnectResult` into and `error`.

You might even consider making a generic `Result` interface type that `DisconnectResult` can implement. This could be expanded to other command results such as `status` (with a `StatusResult`) and `connect` (with a `ConnectResult`.

main.go
```diff
@@ -401,49 +401,175 @@ func connectAction(ctx *cli.Context) error {
...
+	InsightsDisconnectedError string `json:"insights_disconnected_error,omitempty"`
+	YggdrasilStopped          bool   `json:"yggdrasil_stopped"`
+	YggdrasilStoppedError     string `json:"yggdrasil_stopped_error,omitempty"`
+	exitCode                  int
+}
+
+// printJSONDisconnectResult tries to print the result of disconnect as JSON to stdout.
+// When marshaling of systemStatus fails, then error is returned
+func printJSONDisconnectResult(disconnectResult *DisconnectResult) error {
+	data, err := json.MarshalIndent(disconnectResult, "", "    ")
```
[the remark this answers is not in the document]
jirihnidek: We currently use use JSON with indent for `status` command and the generated JSON document is small. I think that it is OK to be consistent and keep output with indent.

main.go
jirihnidek: The `machineReadablePrintFunc` returns error only in the case, when it is not possible to marshal structure into JSON document.

main.go
```diff
@@ -401,49 +401,175 @@ func connectAction(ctx *cli.Context) error {
...
 	hostname, err := os.Hostname()
+	if uiSettings.isMachineReadable {
+		disconnectResult.Hostname = hostname
+	}
 	if err != nil {
-		return cli.Exit(err, 1)
+		exitCode := 1
+		if uiSettings.isMachineReadable {
+			disconnectResult.HostnameError = err.Error()
+			disconnectResult.exitCode = exitCode
```
jirihnidek, line 519: We should terminate method here too.

main.go
```diff
@@ -401,49 +401,175 @@ func connectAction(ctx *cli.Context) error {
...
+
 	uid := os.Getuid()
 	if uid != 0 {
-		return cli.Exit(fmt.Errorf("error: non-root user cannot disconnect system"), 1)
+		errMsg := "non-root user cannot disconnect system"
+		exitCode := 1
+		if uiSettings.isMachineReadable {
+			disconnectResult.UID = uid
+			disconnectResult.UIDError = errMsg
+			disconnectResult.exitCode = exitCode
```
jirihnidek, line 502: We should terminate function here too.

main.go
```diff
@@ -401,49 +401,175 @@ func connectAction(ctx *cli.Context) error {
...
+	case "json":
+		machineReadablePrintFunc = printJSONDisconnectResult
+	default:
+		break
+	}
+
+	// When printing of status is requested, then print machine-readable file format
+	// at the end of this function
+	if uiSettings.isMachineReadable {
+		defer func(disconnectResult *DisconnectResult) {
```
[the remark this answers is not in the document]
jirihnidek: You cannot always declare `defer` function, because when machine readable output is not used, we print some progress messages, spinners, etc. We cannot print everything at the end.

main.go
```diff
@@ -401,49 +401,175 @@ func connectAction(ctx *cli.Context) error {
...
+			format,
+			`"json"`,
+		)
+		return cli.Exit(err, 1)
+	}
+}
+
+// DisconnectResult is structure holding information about result of
+// disconnect command. The result could be printed in machine-readable format.
+type DisconnectResult struct {
```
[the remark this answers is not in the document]
jirihnidek: I did some experiments and it would not help much, because we print progress messages, spinners, etc., when machine readable output is not used. It would be useless to return `DisconnectResult` in such case.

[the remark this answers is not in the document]
subpop: OK! Thanks for giving it a try.

main.go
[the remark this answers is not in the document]
subpop: Hm. Yes. Consistency wins out in this case too, since it's all just whitespace.

jirihnidek:
Refactored code a little bit. Hope it is little bit better now.

main.go
```diff
@@ -774,10 +893,18 @@ func main() {
...
+				&cli.StringFlag{
+					Name:    "format",
+					Usage:   "prints output of disconnection in machine-readable format (supported formats: \"json\")",
+					Aliases: []string{"f"},
+				},
+			},
 			Usage:       "Disconnects the system from " + Provider,
 			UsageText:   fmt.Sprintf("%v disconnect", app.Name),
 			Description: fmt.Sprintf("The disconnect command disconnects the system from Red Hat Subscription Management, Red Hat Insights and %v and deactivates the %v service. %v will no longer be able to interact with the system.", Provider, ServiceName, Provider),
+			Before:      beforeDisconnectAction,
```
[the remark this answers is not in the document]
jirihnidek: Interesting, maybe it could help us to combine "global" and "sub-command" cli options `--format` for v2.

I already did some experiments with combining "global" and "sub-command" cli options with v2, but no workaround work as expected. We can try this in another PR.

jirihnidek:
@subpop Hi, can you please review this PR again?

main.go
```diff
@@ -401,49 +401,188 @@ func connectAction(ctx *cli.Context) error {
...
+	default:
+		break
+	}
+
+	// When printing of status is requested, then print machine-readable file format
+	// at the end of this function
+	if uiSettings.isMachineReadable {
+		defer func(disconnectResult *DisconnectResult) {
+			// When exit code is zero, then print machine-readable output
+			// When exit code has non-zero value, then disconnectResult is returned as a error
```
subpop, line 504: This comment doesn't see to align with the behavior of the code. Is it left over
from some previous code? This is one of the pitfalls of using comments to
describe the algorithm of the code.

main.go
```diff
@@ -401,49 +401,188 @@ func connectAction(ctx *cli.Context) error {
...
+			if disconnectResult.exitCode == 0 && machineReadablePrintFunc != nil {
+				err := machineReadablePrintFunc(disconnectResult)
+				// When it was not possible to print result of disconnect to machine-readable format, then
+				// change returned error to CLI exit error to be able to set exit code to
+				// a non-zero value
+				if err != nil {
+					panic(fmt.Errorf("unable to print status as %s document: %s", format, err.Error()))
```
subpop, line 511: Typically, it's not necessary to explicitly call Error() on an error type. This
should be sufficient.

```suggestion
					panic(fmt.Errorf("unable to print status as %v document: %v", format, err))
```

main.go
```diff
@@ -401,49 +401,188 @@ func connectAction(ctx *cli.Context) error {
...
+				}
+			}
+		}(&disconnectResult)
+	}
+
+	disconnectResult.exitCode = 0
```
subpop, line 517: Is this necessary? The zero value of an `int` is `0` already.

subpop (changes requested):
Looks good over all. I just have a couple very minor questions.

main.go
jirihnidek: Refactored code to not duplicate code. Only `Error()` interface is used now.

subpop:
There appear to be some valid lint errors related to printf format strings.

main.go
subpop: Nice refactor! I like the change and use of Error interface.

main.go
```diff
@@ -401,49 +401,146 @@ func connectAction(ctx *cli.Context) error {
...
+}
+
+// beforeDisconnectAction ensures the used has supplied a correct `--format` flag
+func beforeDisconnectAction(ctx *cli.Context) error {
+	return setupFormatOption(ctx)
+}
+
+// rhcPrintf is method for printing human-readable output. It suppresses output, when
+// machine-readable format is used.
+func rhcPrintf(format string, a ...interface{}) {
```
subpop, line 467: The name of this function doesn't indicate its purpose clearly to me. Could we
call it something like `interactivePrintf`?

subpop (changes requested):

subpop (approved):
OK. This looks good to me!

PR #113, opened 2024-05-21
test: Added tests for rhc connect 

Archana-PandeyM:
Migrating rhc connect tests from test_client.py from iqe testsuite.
Currently some of the tests failing because of bug https://issues.redhat.com/browse/CCT-539

### DIFF
--- a/integration-tests/utils/__init__.py
+++ b/integration-tests/utils/__init__.py
@@ -1,0 +1,74 @@
+import sh
+
+
+def rhcd_service_is_active():
+    """Method to verify if rhcd is in active/inactive state
+    :return: True if rhcd in active state else False
+    """
+    try:
+        stdout = sh.systemctl("is-active rhcd".split()).strip()
+        return stdout == "active"
+    except sh.ErrorReturnCode_3:
+        return False
+
+
+def check_rhcd_journalctl(str_to_check, since_datetime=None, must_exist_in_log=True):
+    """This method helps in verifying strings in rhcd logs
+    :param str_to_check: string to be searched in logs
+    :param since_datetime: start time for logs
+    :param must_exist_in_log: True if str_to_check should exist in log else false
+    :return: True/False
+    """
+    if since_datetime:
+        rhcd_logs = sh.journalctl("-u", "rhcd", "--since", since_datetime)
+    else:
+        rhcd_logs = sh.journalctl("-u", "rhcd")
+
+    if must_exist_in_log:
+        return str_to_check in rhcd_logs
+    else:
+        return str_to_check not in rhcd_logs
+
+
+def prepare_args_for_connect(
+    test_config, auth: str = None, credentials: dict = None, server: str = None
+):
+    """Method to create arguments to be passed in 'rhc connect' command
+    This method expects either auth type or custom credentials
+    """
+    args = []
+    if credentials:
+        for k, v in credentials.items():
+            try:
+                value = test_config.get(v)
+                value = value[0] if isinstance(value, list) else value
+            except KeyError:
+                value = v
+            if value:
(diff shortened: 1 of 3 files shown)


integration-tests/test_client.py
```diff
@@ -0,0 +1,52 @@
+import contextlib
+from utils import rhcd_service_is_active
+
```
jirihnidek, line 3: I would rename this file to `test_connect.py`, because this repository contains only client.

I can also see that all new tests are just skipped: https://artifacts.dev.testing-farm.io/120598b7-e751-420b-81af-5ea45d2cd2af/ What influences that test is skipped?

jirihnidek: I would expect that this file also includes doc string containing explanation of its purpose. It would be also good to add reference to https://github.com/ptoscano/pytest-client-tools/, because this Python module is used here.

integration-tests/utils/__init__.py
```diff
@@ -0,0 +1,9 @@
+import sh
+
+
+def rhcd_service_is_active():
```
jirihnidek, line 4: The doc string is missing here. BTW: the term `rhcd` could be misleading, because upstream name is `yggdrasil`.

jirihnidek (changes requested):
Hi, thanks for your first PR. :+1: I have few questions and requests.

integration-tests/test_client.py
Archana-PandeyM: All new tests are getting skipped because we do not yet have configuration/credentials setup upstream. It should run downstream for now.

Archana-PandeyM: > I would expect that this file also includes doc string containing explanation of its purpose. It would be also good to add reference to https://github.com/ptoscano/pytest-client-tools/, because this Python module is used here.

this repo is in requirements.txt, can you be more specific about how to add reference, as in a comment about it would do? or something else is expected here?

jirihnidek: Each Python module should contain doc string with short explanation like this file:
https://github.com/candlepin/subscription-manager/blob/main/src/cloud_what/providers/aws.py

Why could it be useful? Imagine that you are community contributor and you want to extend integration tests, because you added new feature. It would be nice to see doc string with short explanation and recommended link, because https://github.com/ptoscano/pytest-client-tools/ contains more information and some documentation.

Something like this:

```
"""
This Python module contains integration tests of rhc. it uses pytest-client-tools Python module.
More information about this module could be found: https://github.com/ptoscano/pytest-client-tools/
"""
```

Archana-PandeyM: > Each Python module should contain doc string with short explanation like this file: https://github.com/candlepin/subscription-manager/blob/main/src/cloud_what/providers/aws.py
> 
> Why could it be useful? Imagine that you are community contributor and you want to extend integration tests, because you added new feature. It would be nice to see doc string with short explanation and recommended link, because https://github.com/ptoscano/pytest-client-tools/ contains more information and some documentation.
> 
> Something like this:
> 
> ```
> """
> This Python module contains integration tests of rhc. it uses pytest-client-tools Python module.
> More information about this module could be found: https://github.com/ptoscano/pytest-client-tools/
> """
> ```

so this docstring should exist in all test scripts?

jirihnidek: I'm not telling you to modify all existing test scripts of all our projects. :-) I'm telling you that it is good practice, and if we start adding new tests scripts to this project, then it would be good to add these doc strings to all new `.py` files.

Archana-PandeyM: Done.

Lorquas (approved):
LGTM, the blocking bug will be handled in another PR and this unblocks all the other tests.

PR #115, opened 2024-05-27
test: Added tests for rhc status

Archana-PandeyM:
One test currently fails due to known issue :https://issues.redhat.com/browse/CCT-525

### DIFF
--- a/integration-tests/test_status.py
+++ b/integration-tests/test_status.py
@@ -1,0 +1,100 @@
+"""
+This Python module contains integration tests for rhc. It uses pytest-client-tools Python module.
+More information about this module could be found: https://github.com/ptoscano/pytest-client-tools/
+"""
+
+import pytest
+from pytest_client_tools import util
+from utils import yggdrasil_service_is_active
+
+
+def test_status_connected(external_candlepin, rhc, test_config):
+    """Test RHC Status command when the host is connected.
+    test_steps:
+        1- rhc connect
+        2- rhc status
+    expected_output:
+        1- Validate following strings in status command output
+            "Connected to Red Hat Subscription Management"
+            "Connected to Red Hat Insights"
+            "The yggdrasil/rhcd service is active"
+    """
+    rhc.connect(
+        username=test_config.get("candlepin.username"),
+        password=test_config.get("candlepin.password"),
+    )
+    assert yggdrasil_service_is_active()
+    status_result = rhc.run("status", check=False)
+    assert status_result.returncode == 0
+    assert "Connected to Red Hat Subscription Management" in status_result.stdout
+    assert "Connected to Red Hat Insights" in status_result.stdout
+    if pytest.service_name == "rhcd":
+        assert "The Remote Host Configuration daemon is active" in status_result.stdout
+    else:
+        assert "The yggdrasil service is active" in status_result.stdout
+
+
+def test_status_disconnected(rhc):
+    """Test RHC Status command when the host is disconnected.
+    Ref: https://issues.redhat.com/browse/CCT-525
+    test_steps:
+        1- rhc disconnect
+        2- rhc status
+    expected_output:
+        1- Validate following strings in status command output
+            "Not connected to Red Hat Subscription Management"
+            "Not connected to Red Hat Insights"
+            "The yggdrasil/rhc service is inactive"
+    """
+    # 'rhc disconnect' to ensure system is already disconnected
+    rhc.run("disconnect", check=False)
+    status_result = rhc.run("status", check=False)
+    assert status_result.returncode == 0
+    assert "Not connected to Red Hat Subscription Management" in status_result.stdout
+    assert "Not connected to Red Hat Insights" in status_result.stdout
+    if pytest.service_name == "rhcd":
+        assert "The Remote Host Configuration daemon is active" in status_result.stdout
+    else:
+        assert "The yggdrasil service is inactive" in status_result.stdout
+
+
+@pytest.mark.skipif(
+    pytest.service_name != "rhcd",
+    reason="This test only supports restart of rhcd and not yggdrasil",
+)
+def test_rhcd_service_restart(external_candlepin, rhc, test_config):
+    """
+    Test rhcd service can be restarted on connected and  not on disconnected system.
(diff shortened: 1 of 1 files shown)


integration-tests/utils/__init__.py
```diff
@@ -12,6 +12,11 @@ def yggdrasil_service_is_active():
         return False
 
 
+def show_yggdrasil_service_status():
```
jirihnidek, line 15: I would add some doc string here. It is not clear what this function returns.

integration-tests/test_status.py
```diff
@@ -0,0 +1,96 @@
...
+        2- restart yggdrasil service via systemctl
+    expected_results:
+        1- Yggdrasil service should be restarted successfully and set in active state
+    """
+    with contextlib.suppress(Exception):
+        rhc.disconnect()
+
+    # reboot the test system, but currently it is not possible to actually reboot a
+    # system, so we will try to restart service vis systemd instead
+    # reboot_commands = ["tmt-reboot", "rhts-reboot", "reboot"] ### Change-Me
```
jirihnidek, line 75: Well, if it is not possible to test starting yggdrasil service after reboot ATM, shouldn't we rename this function and add proper function for this test case (starting of yggdrasil service after reboot), when it will be possible?

You know. It is confusing. When you look at the list of methods, then it seems that something is tested, but it is not true. You have to read all comments to realize that this function is a fake test.

jirihnidek (changes requested):
Thanks for PR. I have few comments and proposals.

integration-tests/utils/__init__.py
```diff
@@ -12,6 +12,14 @@ def yggdrasil_service_is_active():
...
+    """This method return status of yggdrasil service.
+    :return: status
+    """
+    status = sh.systemctl("status yggdrasil".split()).strip()
```
Lorquas, line 19: the `sh` library can throw additional tracebacks depending on the systemctl's exit code, which it likely will if this is ever run (this code only happens in exceptions).  To make debugging cleaner would this benefit from either exception handling, or even using `subprocess` - maybe even pytest_client_tools'  `logged_run` ?

integration-tests/test_status.py
```diff
@@ -0,0 +1,93 @@
...
+        1- Validate following strings in status command output
+            "Not connected to Red Hat Subscription Management"
+            "Not connected to Red Hat Insights"
+            "The yggdrasil service is inactive"
+    """
+    rhc.run("disconnect", check=False)
+    status_result = rhc.run("status", check=False)
+    assert status_result.returncode == 0
+    assert "Not connected to Red Hat Subscription Management" in status_result.stdout
+    assert "Not connected to Red Hat Insights" in status_result.stdout
```
Lorquas, line 54: This assert is failing in the downstream and upstream tests?

integration-tests/test_status.py
```diff
@@ -0,0 +1,93 @@
...
+    test_steps:
+        1- disconnect the system
+        2- restart yggdrasil service via systemctl
+        3- run rhc connect
+        4- restart yggdrasil service via systemctl
+    expected_results:
+        1- Yggdrasil service should be restarted successfully and set in active state
+    """
+    with contextlib.suppress(Exception):
+        rhc.disconnect()
```
Lorquas, line 70: is this effectively the same as `rhc.run("disconnect", check=False)` which is used in the previous test above?

Lorquas (commented):
Generally good!  I have a few questions (see inline)...

integration-tests/test_status.py
Archana-PandeyM: yes, because of known issue https://issues.redhat.com/browse/CCT-525

integration-tests/utils/__init__.py
Archana-PandeyM: make sense to use existing logged_run.Thanks for feedback!

Lorquas (approved):
Approved pending discussion on cct-525

integration-tests/utils/__init__.py
ptoscano: well no, `logged_run()` is currently NOT public API (even if it isn't named with `_`)

jirihnidek:
It seems that merging this PR is still blocked by some issue in insights-client on Fedora.

Lorquas:
> It seems that merging this PR is still blocked by some issue in insights-client on Fedora.

insights-client is having some non-related problems on latest python: https://issues.redhat.com/browse/RHINENG-11283

jirihnidek:
> > It seems that merging this PR is still blocked by some issue in insights-client on Fedora.
> 
> insights-client is having some non-related problems on latest python: https://issues.redhat.com/browse/RHINENG-11283

OK, then it would be good to skip failing test on Fedora ATM using e.g. `@pytest.mark.skipif(platform.freedesktop_os_release()["ID"] == "fedora", reason="RHINENG-11283")` and create another card to enable these test, when related issue with insights-client is fixed. It is not good practice to approve failed tests.

Archana-PandeyM:
@jirihnidek the problem with fedora build is resolved with insights-core fix! https://github.com/RedHatInsights/insights-core/pull/4154

integration-tests/test_status.py
```diff
@@ -0,0 +1,87 @@
...
+    """
+    Test yggdrasil service can be restarted on connected and disconnected system.
+    test_steps:
+        1- disconnect the system
+        2- restart yggdrasil service via systemctl
+        3- run rhc connect
+        4- restart yggdrasil service via systemctl
+    expected_results:
+        1- Yggdrasil service should be restarted successfully and set in active state
+    """
```
jirihnidek, line 65: This is actually test of yggdrasil and I'm not sure if it is good to have it here, because when yggdrasil could not be restarted on e.g. disconnected system, then we will not be able to fix it in this repository.

jirihnidek (approved):
LGTM, I would also add some test for `--format json`, but we can add it in the future in another PR.

Archana-PandeyM:
> LGTM, I would also add some test for `--format json`, but we can add it in the future in another PR.
yes we will increase the test coverage to have this tests in another PR.

PR #116, opened 2024-05-27
test: Added test for rhc disconnect

Archana-PandeyM:


Archana-PandeyM:
upstream tests are not using default insights-client.conf settings. 
One test failing upstream test_disconnect_when_already_disconnected, because default value of legacy_upload is True and output of "rhc disconnect" has different output for legacy and non legacy.
We decided that legacy upload will be discontinued and hence all downstream testing is non legacy.

jirihnidek (changes requested):
Thanks for the PR. I can see that new test cause failure of integration tests. Please fix/remove failed test or if it is known issue of insights-client, then mark this test using `@pytest.mark.skip(reason="bug in insights-client CCT-XYZ")`. It is not possible to add new integration tests that are failing.

integration-tests/test_disconnect.py
```diff
@@ -0,0 +1,89 @@
...
+    assert not yggdrasil_service_is_active()
+    if pytest.service_name == "rhcd":
+        assert (
+            "Deactivated the Remote Host Configuration daemon"
+            in disconnect_result.stdout
+        )
+    else:
+        assert "Deactivated the yggdrasil service" in disconnect_result.stdout
+    assert "Disconnected from Red Hat Insights" in disconnect_result.stdout
+    assert "Disconnected from Red Hat Insights" in disconnect_result.stdout
```
jirihnidek, line 44: It does not seem correct. You check twice for a presence of "Disconnected from Red Hat Insights", but you do not check for a presence of string "Disconnected from Red Hat Subscription Management" in `stdout`.

integration-tests/test_disconnect.py
```diff
@@ -0,0 +1,89 @@
...
+
+
+@pytest.mark.skip(
+    reason="Test cannot be run due to unresolved issue https://issues.redhat.com/browse/CCT-525"
+)
```
jirihnidek, line 49: :+1:

jirihnidek (changes requested):
Thanks for the update. :+1:  I think that one thing is not correct.

jirihnidek (approved):
LGTM

PR #120, opened 2024-06-20
feat(test): Install build on donwstream PR job

Lorquas:
Adding script to install from copr/packit build for downstream PR jobs.  Fixes CCT-572

### DIFF
--- a/systemtest/copr-setup.sh
+++ b/systemtest/copr-setup.sh
@@ -1,0 +1,24 @@
+#!/usr/bin/bash -eux
+dnf install -y dnf-plugins-core
+
+#Determine the repo needed from copr
+#Available repositories: 'centos-stream-8-x86_64', 'rhel-8-x86_64',
+# 'centos-stream-9-x86_64', 'rhel-9-x86_64', 'fedora-40-x86_64',
+# 'fedora-39-x86_64', 'fedora-rawhide-x86_64'
+source /etc/os-release
+if [ "$ID" == "centos" ]; then
+  ID='centos-stream'
+fi
+VERSION_MAJOR=$(echo ${VERSION_ID} | cut -d '.' -f 1)
+COPR_REPO="${ID}-${VERSION_MAJOR}-$(uname -m)"
+
+#get yggdrasil
+curl https://copr.fedorainfracloud.org/coprs/g/yggdrasil/latest/repo/$ID-$VERSION_MAJOR/group_yggdrasil-latest-$ID-$VERSION_MAJOR.repo \
+  -o /etc/yum.repos.d/yggdrasil.repo
+
+dnf install -y yggdrasil yggdrasil-worker-package-manager --disablerepo=* --enablerepo=*yggdrasil*
+
+# These PR packit builds have an older version number for some reason than the released...
+dnf remove -y rhc
+dnf copr -y enable packit/RedHatInsights-rhc-${ghprbPullId} ${COPR_REPO}
+dnf install -y rhc --disablerepo=* --enablerepo=*rhc*
--- a/systemtest/tests/integration/test.sh
+++ b/systemtest/tests/integration/test.sh
@@ -5,6 +5,10 @@ set -x
 # get to project root
 cd ../../../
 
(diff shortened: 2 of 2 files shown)


systemtest/tests/integration/test.sh
```diff
@@ -5,6 +5,10 @@ set -x
 # get to project root
 cd ../../../
 
+# Check for git hub pull request ID and install build if needed.
+# This is for the downstream PR jobs.
+[ -z "${ghprbPullId+x}" ] || ./systemtest/copr-setup.sh
```
jirihnidek, line 10: Output from Jenkins job:

```
                cmd: ./test.sh
                out: + cd ../../../
                out: + '[' -z x ']'
                out: + ./systemtest/copr-setup.sh
                out: ./test.sh: line 10: ./systemtest/copr-setup.sh: Permission denied
                out: Shared connection to ibm-x3250m4-02.lab.eng.rdu2.redhat.com closed.
                00:00:00 errr /systemtest/tests/integration (on default-0) [1/1]
```

I would expect some different output there.

BTW: Where does `ghprbPullId` come from? Is it documented somewhere? If yes, could you please add link to the documentation?

Why `"${ghprbPullId+x}"`? The `+x` always adds something to any string (even empty string) and `[ -z ]` is always false. Thus, the following script (`copr-setup.sh`) is always executed.

systemtest/copr-setup.sh
jirihnidek, line 1: The file has probably wrong permissions (missing `x` flag), and it cannot be executed.

jirihnidek (changes requested):
Thanks for the PR. The basic idea is good, but I'm afraid that it does not work as intended.

systemtest/tests/integration/test.sh
Lorquas: Thank you for the thorough review!

`ghprbPullID` comes from the jenkins [GitHub Pull Request Builder](https://plugins.jenkins.io/ghprb/) plugin.  It is what we're using to trigger these PR jobs and post the results back.  It is an environment variable that is injected into the build, that is set to the PR number, in this case '120'.  I kept it ghprbPullID simply so I wouldn't have to remember that the variable here is actually the ghprbPullID from jenkins.  I can rename this variable to anything to make its use more clear if that would help the readability of this.

The bash magic `"${var+x}"` :
This is a bash-ism that I added out of habit; I can remove it.  But for example:

```
#Test if variable is set
➜  ~ myPR='120'; [ -z "${myPR+x}" ] || echo 'myPR is set'
myPR is set

#Test if variable is set even if it is blank
➜  ~ myPR=''; [ -z "${myPR+x}" ] || echo 'myPR is set' 
myPR is set

#setting to a blank variable without the +x, blank and unset are treated the same
➜  ~ myPR=''; [ -z "${myPR}" ] || echo 'myPR is set' 
# no output

#testing for a variable that doesn't exist
➜  ~ unset myPR; [ -z "${myPR+x}" ] || echo 'myPR is set'
# no output
```

The idea here is to always run this on the downstream PR jobs and never anywhere else.  Functionally a blank PR number is about as useful as an unset variable.  But this way a set variable indicates that it should be a jenkins PR job and any failure will come quickly from failing to install copr repos instead of passing along to guest-setup and running the tests against the incorrect build.  The chances of that happening are extremely slim - unless the jenkins plugin fails an api call.  I could also remove this for clarity if you have any feeling one way or another.

I am working on all other suggestions, thank you.  Note that `rhc` has an identical PR up and I will update that to match once this is approved.

systemtest/copr-setup.sh
Lorquas: done.

systemtest/tests/integration/test.sh
```diff
@@ -5,6 +5,10 @@ set -x
 # get to project root
 cd ../../../
 
+# Check for git hub pull request ID and install build if needed.
```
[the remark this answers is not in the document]
Lorquas: done.

jirihnidek (approved):
LGTM, Thanks for changes and comments.

PR #53, opened 2023-10-19
Refactoring of status command and rhcd service

jirihnidek:
Code cleanup before fixing RHEL-2490

### Refactoring of status action
* Split `statusAction` to several logical functions and moved
  these functions to new new file `status.go`
* Refactored getting user interface preferences

### Refactoring of service management
* Previous management was confusing
  * Upstream name of the service is `yggdrasil`, but upstream
    version of `rhc` used default branded name `rhcd` from downstream
    for this service. Thus, it was not possible to use
    upstream version of `rhc` for activating/deactivating
    of upstream version of `yggdrasil`.
  * Brand name was not brand name of `rhc`, but it was used
    for branding of service name.
  * The service name was assembled from short name of `rhc`.
    `rhc + "d.service"`. Thus, it actually was not possible to
    brand the service.
* Removed `BrandName` and introduced `ServiceName` and `SERVICENAME`
  constants. `SERVICENAME` can be configured in Makefile
  * Modified spec files according this change
* Replaced all occurrence of keyword daemon with keyword service,
  because "service" is more accurate keyword and it could
  not scare people ;-)
* Changed output of `rhc status --format json` a little bit.
  Renamed "rhcd_running" to "yggdrasil_running" and
  "rhcd_error" to "yggdrasil_error".
* Refactored code a little bit

PR #122, opened 2024-06-27
fix(test):Refactoring to replace service name from rhcd to yggdrasil

Archana-PandeyM:
- Upstream name of the service is changed from rhcd to yggdrasil, so replacing the service name from rhcd to yggdrasil in tests also.
Ref: https://github.com/RedHatInsights/rhc/pull/53


jirihnidek (approved):
LGTM, thanks

PR #125, opened 2024-07-12
fix: Detect insights-client status in non-legacy mode

m-horky:
* Card ID: CCT-525

With `legacy_upload=True` mode, `insights-client --status` returns code
0 if the system is registered and 1 otherwise. In non-legacy mode, 0 is
always returned.

Even though the `/etc/insights-client/.registered` file is not a public
API, it is reliable and we should be able to use it to detect the
registration status.

### DIFF
--- a/insights.go
+++ b/insights.go
@@ -1,11 +1,8 @@
 package main
 
 import (
-	"bytes"
-	"errors"
-	"fmt"
+	"os"
 	"os/exec"
-	"strings"
 )
 
 func registerInsights() error {
@@ -21,29 +18,22 @@ func unregisterInsights() error {
 }
 
 func insightsIsRegistered() (bool, error) {
-	var errBuffer bytes.Buffer
-	cmd := exec.Command("/usr/bin/insights-client", "--status")
-	cmd.Stderr = &errBuffer
-
-	err := cmd.Run()
-
+	// While `insights-client --status` properly checks for registration status by
+	// asking Inventory, its two modes (legacy v. non-legacy API) behave
+	// differently (they return different texts with different exit codes) and
+	// we can't rely on the output or exit codes.
+	// The `.registered` file is always present on a registered system.
+	err := exec.Command("/usr/bin/insights-client", "--status").Run()
 	if err != nil {
-		// When the error is ExitError, then we know that insights-client only returned
-		// some error code not equal to zero. We do not care about error number.
-		var exitError *exec.ExitError
-		if errors.As(err, &exitError) {
-			// When stderr is not empty, then we should return this as error
-			// to be able to print this error in rhc output
-			stdErr := errBuffer.String()
-			if len(stdErr) == 0 {
-				return false, nil
-			} else {
-				return false, fmt.Errorf("%s", strings.TrimSpace(stdErr))
-			}
-		} else {
-			return false, err
-		}
+		return false, err
 	}
 
-	return cmd.ProcessState.Success(), err
+	_, err = os.Stat("/etc/insights-client/.registered")
+	if os.IsNotExist(err) {
+		return false, nil
+	}
+	if err != nil {
+		return false, err
+	}
+	return true, nil
 }
--- a/status.go
+++ b/status.go
@@ -66,7 +66,7 @@ func insightStatus(systemStatus *SystemStatus) {
 				systemStatus.InsightsConnected = false
 				systemStatus.InsightsError = err.Error()
 			} else {
-				fmt.Printf(uiSettings.iconError+" Cannot execute insights-client: %v\n", err)
+				fmt.Printf(uiSettings.iconError+" Cannot detect Red Hat Insights status: %v\n", err)
 			}
 		}
 	}


jirihnidek:
> In non-legacy mode, 0 is always returned. For this reason, we should rely on the message emitted to standard output.

Shouldn't we rather fix `insights-client`? I really don't like the idea of cancer code ... bad code causing bad code in other projects.

jirihnidek:
The logic of returning zero, when system is connected, is correct. Returning non-zero code, when system is not connected for various reasons, is also correct. The `subscription-manager status` returns such codes. I believe that we should be consistent and follow this behavior in other client tools including `rhc` and `insights-client`.

jirihnidek:
There is really low risk of doing such change in `insights-client`. If somebody checks returned value of `insights-client --status` e.g. in some script or any other automation, then it is also expected that some non zero value is returned.

I'm not willing to do wrong things in `rhc`, because there is some bug in `insights-client`. For this reason I'm closing this PR.

m-horky:
I'd like you to reconsider this PR. It seems too risky to change the return codes for insight-client. Since the behavior is delivered in a Core, it will be pushed everywhere, that's not how RHEL behaves.

It wouldn't be responsible. We _should_ include this patch, even though it's not as pretty as we'd have hoped. Unfortunately, the reality is what is is.

subpop:
I agree that poor code in insights-client should not create poor code in other projects. I also agree that exit codes from programs should be consistent. I also disagree with this patch, but more on the grounds that relying on output strings is dangerous. If a change in core later changes this output, suddenly rhc clients everywhere will start misbehaving. I am generally against tightly-coupling programs like this; it creates tenuous dependencies that become extremely complex to maintain. Making any change to insights-client comes with a very high risk of breaking other projects because of the exact reason I described in the previous sentence; it's terrible CLI has become a de-facto API that other tooling rely upon. Could we modify `rhc` to look for the existence of the ".registered" and ".unregistered" marker files instead of the standard output or exit codes of `insights-client`? It's still a fragile dependency, but it feels less likely to cause other programs to break.

m-horky:
That's not a bad idea. `.registered` and `.unregistered` is already de-facto API which is used semi-internally. While I'm not happy that the files are stored in `/etc/`, it's the best thing we have. I'll update the PR to try that out instead.

subpop:
It's not elegant, but the absence of those files can be more elegantly handled than string output changing without warning.

subpop (approved):
Looks good to me! Thanks!

subpop:
I want to wait for @jirihnidek opinion in this case before merging, since he raised concerns about the original implementation.

insights.go
```diff
@@ -21,29 +18,19 @@ func unregisterInsights() error {
...
-	err := cmd.Run()
-
+	// While `insights-client --status` properly checks for registration status by
+	// asking Inventory, its two modes (legacy v. non-legacy API) behave
+	// differently (they return different texts with different exit codes).
+	// The `.registered` file is always present on the registered system; the worst
+	// thing that may happen is that we're out of date and the insights-client does
+	// not yet know it should unregister. However, this detection is reliable
+	// enough for our reporting use-case.
+	_, err := os.Stat("/etc/insights-client/.registered")
```
jirihnidek, line 28: Hidden file in `/etc` directory... What a bright idea to use such file for any purpose. Why do we have e.g. POSIX?

This is also horrible, but it is less painful than parsing output of command. It is also mush faster than running `insights-client --status`.

Unfortunately, it looks like that it is not reliable. Or how should I understand to this:

```
[root@rhel9 ~]# test -f /etc/insights-client/.registered 
[root@rhel9 ~]# echo $?
0
[root@rhel9 ~]# insights-client --status
System is registered locally via .registered file. Registered at 2024-08-22T11:03:02.414759
Insights API says this machine is NOT registered.
System unregistered locally via .unregistered file
[root@rhel9 ~]# echo $?
1
```

Actually, why do I see it? I don't aware that I have done anything unusual with this system. The system is registered (consumer certificate/key is installed and it is valid, not deleted on server side).

jirihnidek (commented):
I have some questions and concerns.

It seems that there are conflicts, because I fixed all lint issues as part of another PR and this PR was merged to main branch. I have some questions.

insights.go
subpop: It's an unfortunate precedent that has been set by previous authors of `insights-client`. We can't really remove it until we have an opportunity to break CLI compatibility, so we're stuck with it for now. The way the marker files are intended to work is: if `/etc/insights-client/.registered` exists, the client believes it is registered with console.redhat.com. If `/etc/insights-client/.unregistered` exists, the client believes it is not registered. There are probably cases that can come about where those marker files get improperly set, but it's the best we've got besides parsing string output.

subpop: Does running `insights-client --status` alter the .registered/.unregistered marker files at all?

m-horky: It doesn't. It may only speed up the detection of the host being gone from Inventory -- if we want to ensure we have the most up to date information, we could run status. But I don't think that's necessary.

jirihnidek: Wait. As you can see here:

```
[root@rhel9 ~]# test -f /etc/insights-client/.registered 
[root@rhel9 ~]# echo $?
0
[root@rhel9 ~]# insights-client --status
System is registered locally via .registered file. Registered at 2024-08-22T11:03:02.414759
Insights API says this machine is NOT registered.
System unregistered locally via .unregistered file
[root@rhel9 ~]# echo $?
1
```

running `insights-client --status` deleted `/etc/insights-client/.registered` and created `/etc/insights-client/.unregistered`. So, it can alter these files.

Is it any corner case and we can ignore it?

m-horky: How did that `.registered` file get there, via proper client registration? If so, was it recently? If you haven't run the client in a long time (e.g. the VM was off), it might have been removed from Inventory after a staleness period, so it would make sense to unregister on status.

Running `--status` forces the client to check with Inventory. Regularly that's not a problem, because the system checks in once a day and the registration status is up to date, and this change (deletion of `.registered` and creation of `.unregistered`) would be done during a night.

m-horky: That being said, we can run `--status` before detecting the presence of the file. I've updated the PR.

insights.go
```diff
@@ -21,29 +18,18 @@ func unregisterInsights() error {
...
-	cmd := exec.Command("/usr/bin/insights-client", "--status")
-	cmd.Stderr = &errBuffer
-
-	err := cmd.Run()
-
+	// While `insights-client --status` properly checks for registration status by
+	// asking Inventory, its two modes (legacy v. non-legacy API) behave
+	// differently (they return different texts with different exit codes).
+	// The `.registered` file is always present on the registered system.
+	_ = exec.Command("/usr/bin/insights-client", "--status").Run()
```
jirihnidek, line 25: I would not ignore returned error here. E.g. If the file `/usr/bin/insights-client` did not exist, then I would return such error.

jirihnidek (changes requested):
It is more reliable approach, but I think that we should not ignore some critical error cases.

insights.go
```diff
@@ -21,29 +18,22 @@ func unregisterInsights() error {
...
+	err := exec.Command("/usr/bin/insights-client", "--status").Run()
 	if err != nil {
-		// When the error is ExitError, then we know that insights-client only returned
-		// some error code not equal to zero. We do not care about error number.
-		var exitError *exec.ExitError
-		if errors.As(err, &exitError) {
-			// When stderr is not empty, then we should return this as error
-			// to be able to print this error in rhc output
-			stdErr := errBuffer.String()
-			if len(stdErr) == 0 {
```
jirihnidek, line 38: I would not delete old functionality altogether, because it will change behavior as you can see on this example:

```
root@localhost:/path/to/rhc# cat /usr/bin/insights-client 
#!/bin/bash

exit 1
root@localhost:/path/to/rhc# ./rhc status
Connection status for thinkpad-p1:

● Cannot detect Red Hat Insights status: exit status 1
● The yggdrasil service is active

```

If insights-client returns non-zero value, can we `return false, nil` to be consistent with old behavior?

jirihnidek (changes requested):
I have one more request and it seems that integration tests are failing due to change in this PR. This should be also fixed before PR is merged.

insights.go
m-horky: I wouldn't say that would be correct, because you get non-zero code even on network errors. It might have been the previous behavior, but it could be misleading. I would rather report the cached value risking it's out of date then report it's not connected to insights because the network glitched up.

jirihnidek (approved):
I hope that we will be able to solve this in better way soon.

PR #126, opened 2024-07-17
fix: Return non-zero code, when some sub-status is wrong

jirihnidek:
* When some sub-status is not correct (system is not registered, insights-client is not registered or yggdrasil is not running), then return non-zero status code
* This change follow behavior of subscription-manager and it could help automation

### DIFF
--- a/main.go
+++ b/main.go
@@ -504,6 +504,7 @@ type SystemStatus struct {
 	InsightsError     string `json:"insights_error,omitempty"`
 	YggdrasilRunning  bool   `json:"yggdrasil_running"`
 	YggdrasilError    string `json:"yggdrasil_error,omitempty"`
+	returnCode        int
 }
 
 // printJSONStatus tries to print the system status as JSON to stdout.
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -608,6 +613,12 @@ func statusAction(ctx *cli.Context) (err error) {
 		fmt.Printf("\nManage your connected systems: https://red.ht/connector\n")
 	}
 
+	// At the end check if all statuses are correct.
+	// If not, return 1 exit code without any message.
+	if systemStatus.returnCode != 0 {
+		return cli.Exit("", 1)
```
subpop, line 619: Shouldn't this be?

```suggestion
		return cli.Exit("", systemStatus.returnCode)
```

subpop (commented):
Good idea. We should document the values of the return code in the man pages somehow too.

main.go
jirihnidek: Well, the value of `systemStatus.returnCode` depends on "failed" sub-statuses. It could be 1, 2, 3, but we return value 1 for all other error cases.

I think that we should return different values for various situations (not only 1), but we should think twice before we do it and create some design document for that.

subpop: Ok, so this change is just applying a non-zero return code when any sub-status is non-zero. Seems reasonable.

subpop (approved):
LGTM

subpop:
@Lorquas Is the downstream test failure something we should ignore for this PR?

jirihnidek:
@subpop Jenkins job is constantly failing. I think that we can ignore it.

PR #127, opened 2024-07-19
feat: Add --format json for connect command

jirihnidek:
* It is possible to run connect command with CLI option
  `--format json`. The output of connect command is in machine
  readable format in such case.

### DIFF
--- a/main.go
+++ b/main.go
@@ -105,18 +105,20 @@ func showTimeDuration(durations map[string]time.Duration) {
 // showErrorMessages shows table with all error messages gathered during action
 func showErrorMessages(action string, errorMessages map[string]LogMessage) error {
 	if hasPriorityErrors(errorMessages, log.CurrentLevel()) {
-		fmt.Println()
-		fmt.Printf("The following errors were encountered during %s:\n\n", action)
-		w := tabwriter.NewWriter(os.Stdout, 0, 0, 2, ' ', 0)
-		_, _ = fmt.Fprintln(w, "TYPE\tSTEP\tERROR\t")
-		for step, logMsg := range errorMessages {
-			if logMsg.level <= log.CurrentLevel() {
-				_, _ = fmt.Fprintf(w, "%v\t%v\t%v\n", logMsg.level, step, logMsg.message)
+		if !uiSettings.isMachineReadable {
+			fmt.Println()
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -401,49 +503,216 @@ func connectAction(ctx *cli.Context) error {
...
+	InsightsDisconnectedError string `json:"insights_disconnected_error,omitempty"`
+	YggdrasilStopped          bool   `json:"yggdrasil_stopped"`
+	YggdrasilStoppedError     string `json:"yggdrasil_stopped_error,omitempty"`
+	exitCode                  int
+	format                    string
+}
+
+// ConnectResult is structure holding information about results
+// of connect command. The result could be printed in machine-readable format.
+type ConnectResult struct {
```
subpop, line 548: Does `ConnectResult` not conform to `error` interface?

main.go
```diff
@@ -401,49 +503,216 @@ func connectAction(ctx *cli.Context) error {
...
+}
+
+// beforeDisconnectAction ensures the used has supplied a correct `--format` flag
+func beforeDisconnectAction(ctx *cli.Context) error {
+	return setupFormatOption(ctx)
+}
+
+// rhcPrintf is method for printing human-readable output. It suppresses output, when
+// machine-readable format is used.
+func rhcPrintf(format string, a ...interface{}) {
```
subpop, line 608: The name of this function doesn't seem to clearly indicate what's happening.
Maybe it should be renamed to something that more accurately identifies what it
does. `interactivePrintf` or `consolePrintf`?

main.go
```diff
@@ -401,49 +503,216 @@ func connectAction(ctx *cli.Context) error {
...
+		result = string(data)
+	default:
+		result = "error: unsupported document format: " + disconnectResult.format
+	}
+	return result
+}
+
+// printJSONConnectResult tries to print the result of connect as JSON to stdout.
+// When marshaling of systemStatus fails, then error is returned
+func printJSONConnectResult(connectResult *ConnectResult) error {
```
subpop, line 581: Could this function (and likewise for `printJSONDisconnectResult`) be converted
to a `String` method on the type, conforming the type to the `Stringer`
interface? Then values could be passed to `Printf` format strings that include
"%v" and printed when necessary?

subpop (approved):
Looks good to me!

PR #128, opened 2024-07-22
add installing katello rpm for rhc+satellite testing

zhangqianqian:


integration-tests/test_connect.py
```diff
@@ -18,6 +18,7 @@
 )
 
 logger = logging.getLogger(__name__)
+pytestmark = pytest.mark.usefixtures("install_katello_rpm")
 
 
 @pytest.mark.parametrize("auth", ["basic", "activation-key"])
```
Archana-PandeyM, line 24: out of curiosity: Can we run basic auth for satellite? If so then ok else we need to skip basic auth for satellite

Archana-PandeyM (changes requested):

integration-tests/test_connect.py
jirihnidek: Basic authentication is possible with any candlepin server. Thus, it is also possible, when Satellite is used.

Archana-PandeyM: Thanks George!

zhangqianqian: Hi @jirihnidek I recall that basic auth is deprecated with insights-client. So what about the basic auth for rhc?

integration-tests/test_connect.py
```diff
@@ -18,6 +18,7 @@
 )
 
 logger = logging.getLogger(__name__)
+pytestmark = pytest.mark.usefixtures("install_katello_rpm")
```
Archana-PandeyM, line 21: I would suggest that instead of using it in all test scripts, try autouse=True in fixture  "install_katello_rpm"

PR #133, opened 2024-08-07
fix(test): Updated tests to run for rhcd downstream

Archana-PandeyM:
Currently, running these tests downstream fail because the name of service is still rhcd in downstream package. As a fix I have added variable to identify the name of service and perform verification accordingly.

### DIFF
--- a/integration-tests/utils/__init__.py
+++ b/integration-tests/utils/__init__.py
@@ -1,35 +1,38 @@
+import pytest
 import sh
 
 
 def yggdrasil_service_is_active():
-    """Method to verify if yggdrasil is in active/inactive state
-    :return: True if yggdrasil in active state else False
+    """Method to verify if yggdrasil/rhcd is in active/inactive state
+    :return: True if yggdrasil/rhcd in active state else False
+    Note: upstream name of service is yggdrasil and downstream is rhcd
     """
     try:
-        stdout = sh.systemctl("is-active yggdrasil".split()).strip()
+        stdout = sh.systemctl(f"is-active {pytest.service_name}".split()).strip()
         return stdout == "active"
     except sh.ErrorReturnCode_3:
         return False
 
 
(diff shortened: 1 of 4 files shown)


zhangqianqian:
Run this PR code with PR https://github.com/RedHatInsights/rhc/pull/128 for rhc+satellite on RHEL-10.0-20240806.0, and the test results looks good.

integration-tests/utils/__init__.py
```diff
@@ -1,35 +1,34 @@
+import pytest
 import sh
 
 
-def yggdrasil_service_is_active():
-    """Method to verify if yggdrasil is in active/inactive state
-    :return: True if yggdrasil in active state else False
+def service_is_active():
```
jirihnidek, line 5: I would not rename this method (I would keep `yggdrasil_service_is_active`), because there is also rhsm.service that is used by rhc. It could be confusing what service is mentioned here. It is OK to keep upstream name yggdrasil for methods and variables.

integration-tests/utils/__init__.py
```diff
@@ -1,35 +1,34 @@
...
         return stdout == "active"
     except sh.ErrorReturnCode_3:
         return False
 
 
-def check_yggdrasil_journalctl(
-    str_to_check, since_datetime=None, must_exist_in_log=True
-):
-    """This method helps in verifying strings in yggdrasil logs
+def check_journalctl_logs(str_to_check, since_datetime=None, must_exist_in_log=True):
```
jirihnidek, line 16: I would keep original name for this method too.

jirihnidek (changes requested):
Overall it is good. I would only keep yggdrasil keyword in names of functions and variables to be more descriptive. It is better to keep this keyword in names of variables and function, then dropping this keyword altogether.

jirihnidek (approved):
Thanks for updates. LGTM now

PR #134, opened 2024-08-08
fix: fixed rhc dependency issue for downstream PR testing

Archana-PandeyM:
Running Pr jobs downstream fails because sometimes beaker composes dont have client tools installed, by adding this rhc install it will help mitigate dependency issues.


Lorquas (approved):
Looks good and the downstream worked as expected.

PR #140, opened 2024-08-15
feat: Saving test report as junit file

Lorquas:
Saving the test report as a junit file.  This will also fix failing rhc jenkins tests since they're now expecting a junit.xml


jirihnidek (approved):
LGTM

PR #146, opened 2024-09-12
test: Fixed test of rhc status

jirihnidek:
* The returned value is non-zero, when some sub-task returns non-zero value. This change was introduced in this PR: https://github.com/RedHatInsights/rhc/pull/126
* We should check for non-zero code, not zero code in this case


Archana-PandeyM:
LGTM.

PR #147, opened 2024-09-25
fix: Try to detect status of insights-client correctly

jirihnidek:
* When insights-client is not connected and non-legacy mode is used, then we should not print error message, but we should simply say that:

    "Not connected to Red Hat Insights"

  It is reason, why we have integration tests.
* I hope that we will be able to solve getting status using more reliable way in the future

### DIFF
--- a/insights.go
+++ b/insights.go
@@ -1,6 +1,8 @@
 package main
 
 import (
+	"errors"
+	"io/fs"
 	"os"
 	"os/exec"
 )
@@ -17,6 +19,10 @@ func unregisterInsights() error {
 	return cmd.Run()
 }
 
+// insightsIsRegistered checks whether insights-client reports its
+// status as registered or not. If the system is registered, `true` is
+// returned, otherwise `false` is returned, and `error` is filled with
+// an error value.
 func insightsIsRegistered() (bool, error) {
 	// While `insights-client --status` properly checks for registration status by
 	// asking Inventory, its two modes (legacy v. non-legacy API) behave
@@ -25,11 +31,21 @@ func insightsIsRegistered() (bool, error) {
 	// The `.registered` file is always present on a registered system.
 	err := exec.Command("/usr/bin/insights-client", "--status").Run()
 	if err != nil {
+		var exitError *exec.ExitError
+		if errors.As(err, &exitError) {
+			// If .unregistered exists, insights-client is confident
+			// it is not registered. We can suppress the error,
+			// we don't care why it returned non-zero exit code.
+			_, err := os.Stat("/etc/insights-client/.unregistered")
+			if err == nil {
+				return false, nil
+			}
+		}
 		return false, err
 	}
 
 	_, err = os.Stat("/etc/insights-client/.registered")
-	if os.IsNotExist(err) {
+	if errors.Is(err, fs.ErrNotExist) {
 		return false, nil
 	}
 	if err != nil {


jirihnidek:
Issue was introduced in this PR: https://github.com/RedHatInsights/rhc/pull/125

insights.go
```diff
@@ -25,11 +31,25 @@ func insightsIsRegistered() (bool, error) {
...
+		if errors.As(err, &exitError) {
+			// Check if .unregistered file exists. In that case insights-client
+			// is probably in non-legacy mode, and it is only unregistered.
+			// There is hopefully no reason to return any error in such case.
+			_, err := os.Stat("/etc/insights-client/.unregistered")
+			if err == nil {
+				return false, nil
+			}
+		} else {
+			return false, err
```
m-horky, line 46: We don't need this `else` branch, do we? It should ball back to the `return` just below.

insights.go
```diff
@@ -17,6 +19,10 @@ func unregisterInsights() error {
 	return cmd.Run()
 }
 
+// insightsIsRegistered tries to get information if system is registered
+// using insights-client, and it can also return error. Current implementation
+// of insights-client is suboptimal. For this reason getting status of
+// insights-client is horrible heuristics. Hopefully, it will change in the future.
```
m-horky, line 25: Could we drop the emotions here a bit? :)

> `...using insights-client. The package is not able to reliably communicate its status via return codes. We have to use its semi-private hidden files to determine the actual status.`

m-horky (changes requested):
Sure, this change makes sense. I'd like to see the comments altered, though.

insights.go
jirihnidek: No, we cannot! Emotions are spice of life.

insights.go
jirihnidek: You are right.

insights.go
```diff
@@ -25,11 +31,25 @@ func insightsIsRegistered() (bool, error) {
 	// The `.registered` file is always present on a registered system.
 	err := exec.Command("/usr/bin/insights-client", "--status").Run()
 	if err != nil {
+		// When the error is ExitError, then we know that insights-client only returned
+		// some error code not equal to zero. We do not care about error number.
+		var exitError *exec.ExitError
+		if errors.As(err, &exitError) {
+			// Check if .unregistered file exists. In that case insights-client
+			// is probably in non-legacy mode, and it is only unregistered.
+			// There is hopefully no reason to return any error in such case.
```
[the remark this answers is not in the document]
jirihnidek: OK

insights.go
m-horky: ...and they might be very confusing to someone new just trying to figure out why the code is so weird here :)

Archana-PandeyM:
some negative integration tests are failing in the jenkins for example 'test_connect_wrong_parameters', test expects a non zero exit code however even if the rhc connect did not succeed we receive exit code 0. this should not be happening.

insights.go
subpop: Opinion comments tend to draw attention away from the intention of the comment. Code that is non-obvious should be commented with a brief explanation for the deviation from the normal or expected pattern, and nothing more.

Comments tend to "bit rot" more than code itself because they have no influence over the behavior of the program. The more text present in a comment, the more chance there is for the comment to remain unmodified as the code around it changes, leading to comments that are ignored by developers, but left in the code base to mislead new developers in the future.

I try to comment code as little as possible, and instead try to write the code so that it documents itself.

subpop, line 25: Doc comments should document the objective or purpose for the function as well as when to expect certain conditions from the return values. They typically should not include a summary of the implementation of the function. This makes the comment more resilient to changes within the function. For example, when insights-client does improve its ability to report status more accurately, this comment could become inaccurate before we get a chance to come in and change the comment.

(Note: I don't know if this suggestion accurately describes the return conditions; that should be verified before applying this suggestion.)

```suggestion
// insightsIsRegistered checks whether insights-client reports its
// status as registered or not. If the system is registered, `true` is
// returned, otherwise `false` is returned, and `error` is filled with
// an error value.
```

insights.go
```diff
@@ -25,11 +31,25 @@ func insightsIsRegistered() (bool, error) {
 	// The `.registered` file is always present on a registered system.
```
subpop, line 31: This comment is a good example of a comment documenting the reason for the deviation from the expected behavior. It's easy to read, and explicitly describes how and why.

insights.go
```diff
@@ -25,11 +31,25 @@ func insightsIsRegistered() (bool, error) {
...
 	err := exec.Command("/usr/bin/insights-client", "--status").Run()
 	if err != nil {
+		// When the error is ExitError, then we know that insights-client only returned
+		// some error code not equal to zero. We do not care about error number.
```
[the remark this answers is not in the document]
subpop: Agree. This comment is redundant.

insights.go
```diff
@@ -25,11 +31,25 @@ func insightsIsRegistered() (bool, error) {
...
+			// Check if .unregistered file exists. In that case insights-client
+			// is probably in non-legacy mode, and it is only unregistered.
+			// There is hopefully no reason to return any error in such case.
+			_, err := os.Stat("/etc/insights-client/.unregistered")
+			if err == nil {
+				return false, nil
+			}
+		} else {
+			return false, err
+		}
```
subpop, line 47: Why do we care about the exit code from running insights-client? Would it be easier to check for the presence of both `/etc/insights-client/.registered` and `/etc/insights-client/.unregistered` and determine status from the existence or non-existence of those files?

subpop (changes requested):

insights.go
jirihnidek: Because current `insights-client` is "sub-optimal". You cannot trust to any of sign 100%. You have to combine several signs together to get something at least little bit reliable.

m-horky: You aren't helping yourself by disliking insights-client. You can't trust _any_ software 100%. Just accept its quirkiness as a fact and move on. Some parts of the whole stack are much less reliable than the client itself.

jirihnidek: It's not about liking or disliking insights-client. I'm only describing current situation. If we relied only on existence of `/etc/insights-client/.registered` and `/etc/insights-client/.unregistered`, then it would not be reliable as it was described in the case in https://github.com/RedHatInsights/rhc/pull/125#discussion_r1747070445.

subpop (approved):
This looks good. Concise changes. The logic is easy enough to follow.

PR #148, opened 2024-10-03
test: Added e2e test for rhc connect

Archana-PandeyM:
This test is part of test migrations and currently fails due to open issues https://issues.redhat.com/browse/CCT-862 and https://issues.redhat.com/browse/CCT-696. Currently it is marked to skipped. we will enable it once the issues are fixed

### DIFF
--- a/integration-tests/test_e2e.py
+++ b/integration-tests/test_e2e.py
@@ -1,0 +1,45 @@
+"""
+This Python module contains integration tests for rhc.
+It uses pytest-client-tools Python module.More information about this
+module could be found: https://github.com/ptoscano/pytest-client-tools/
+"""
+
+import contextlib
+import pytest
+import time
+from utils import (
+    yggdrasil_service_is_active,
+    prepare_args_for_connect,
+)
+
+
+@pytest.mark.skip("Test cannot be run due to unresolved issues CCT-862 and CCT-696")
+@pytest.mark.parametrize("auth", ["activation-key", "basic"])
+def test_rhc_client_connect_e2e(rhc, test_config, auth, external_inventory, subman):
+    """
+    Test RHC Client connect e2e using basic auth and activation key
+    test_steps:
+        1. Run RHC client connect
+    expected_results:
+        1. The rhcd service is running and active
+        2. A new host is created in Inventory
+        3. The rhc_client_id in system profile matches the subman-id
+    """
+    with contextlib.suppress(Exception):
+        rhc.disconnect()
+    command_args = prepare_args_for_connect(test_config, auth=auth)
+    command = ["connect"] + command_args
+    rhc.run(*command)
+    assert rhc.is_registered
+    assert yggdrasil_service_is_active()
+    timeout = 60.0
+    start = time.time()
+    while True:
+        system_profile = external_inventory.this_system_profile()
+        if "rhc_client_id" in system_profile:
+            assert system_profile["rhc_client_id"] == str(subman.uuid)
+            break
+        current = time.time()
+        if current - start > timeout:
+            raise ValueError("timeout")
+        time.sleep(10)


integration-tests/utils/__init__.py
```diff
@@ -77,3 +77,17 @@ def prepare_args_for_connect(
...
+    try:
+        result = sh.openssl(f"x509 -in {cert_file} -subject -nocert".split()).strip()
+        if "CN" in result:
+            rhc_client_id = result.split()[1].split("=")[1]
+            return rhc_client_id
+        else:
+            return None
+    except sh.ErrorReturnCode as e:
+        print(f"An error occured: {e} ")
+        return None
```
ptoscano, line 93: rather than this function, make explicit use of the `subman` fixture in `test_rhc_client_connect_e2e()`, and use `subman.uuid`

Archana-PandeyM: thanks!

integration-tests/test_e2e.py
```diff
@@ -0,0 +1,30 @@
+import contextlib
+import pytest
+from time import sleep
+from utils import (
+    yggdrasil_service_is_active,
+    prepare_args_for_connect,
+)
+
```
jirihnidek, line 8: Please add doc string here describing this module.

jirihnidek (commented):
This test is skipped on all testing farm machines and it fails on Jenkins. Why should we add such test? What actually does it test? It seems that tests fail, because rhsm.service or yggdrasil.service does not start. We should fix these issues first.

Archana-PandeyM:
> This test is skipped on all testing farm machines and it fails on Jenkins. Why should we add such test? What actually does it test? It seems that tests fail, because rhsm.service or yggdrasil.service does not start. We should fix these issues first.

We have had this issue with the tests being migrated from iqe to upstream. As Already mentioned in earlier PRs, upstreams tests are not running because currently we have not implemented a way to store credentials to connect to prod/stage and hence tests are not running.
Now downstream these basic tests are failing because we are continuously making changes upstream , for example currently the yggdrasil service cannot be started on RHEL systems because of configuration in conf.toml file. I am looking into it for possible workarouds.
 So we can decide on not adding any tests because in this phase all will either skip or fail.

integration-tests/test_e2e.py
```diff
@@ -0,0 +1,36 @@
...
+        3. The host rhc client id matches the rhc client id collected from the machine
+    """
+    with contextlib.suppress(Exception):
+        rhc.disconnect()
+    command_args = prepare_args_for_connect(test_config, auth=auth)
+    command = ["connect"] + command_args
+    rhc.run(*command)
+    assert rhc.is_registered
+    assert yggdrasil_service_is_active()
+    sleep(30)  # small wait time for inventory sync up
```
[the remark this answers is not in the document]
jstavel: There is 'loop_until' method in conftest.py. I guess it could help with this statement. It is necessary to define predicate for a loop to stop only. https://github.com/RedHatInsights/insights-client/blob/master/integration-tests/conftest.py#L43  an example of using the function: https://github.com/RedHatInsights/insights-client/blob/master/integration-tests/test_display_name_option.py#L71

integration-tests/test_e2e.py
```diff
@@ -0,0 +1,37 @@
...
+import contextlib
+import pytest
+from time import sleep
+from utils import (
+    yggdrasil_service_is_active,
+    prepare_args_for_connect,
+)
+
+
+@pytest.mark.skip("Test cannot be run due to unresolved issues CCT-862 and CCT-696")
```
Archana-PandeyM, line 16: I think it would be a good move to skip this tests till  the related issue are fixed. I will keep a note in the CCT card to re enable this test. This will help us close the test migrations. @jirihnidek

jirihnidek (approved):
Thanks for updates. LGTM

PR #150, opened 2024-10-04
ci: build & test also on centos-stream-10

ptoscano:
Factorize the existing tests blocks for centos-stream first, so adding the new version is easier.

### DIFF
--- a/.packit.yml
+++ b/.packit.yml
@@ -49,21 +49,9 @@ jobs:
 
   - job: tests
     trigger: pull_request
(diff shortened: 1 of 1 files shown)


ptoscano:
> Shouldn't we drop support for centos-stream-8 ? It is not supported anymore, right? I can still see it in the "targets".

That's a separate discussion, this is focusing only on c10s.

jirihnidek (approved):
LGTM

PR #163, opened 2024-11-04
test: Added integration test for rhc status --format json

jirihnidek:
* Card ID: CCT-945
* Added one test of the case, when status command is used with `--format json`. It checks that output is in JSON format and it includes all attributes

### DIFF
--- a/integration-tests/test_status.py
+++ b/integration-tests/test_status.py
@@ -4,6 +4,7 @@ More information about this module could be found: https://github.com/ptoscano/p
 """
 
 import pytest
+import json
 from pytest_client_tools import util
 from utils import yggdrasil_service_is_active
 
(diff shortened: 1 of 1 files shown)


jirihnidek:
@Archana-PandeyM can you please review this PR?

integration-tests/test_status.py
```diff
@@ -34,6 +35,32 @@ def test_status_connected(external_candlepin, rhc, test_config):
...
+    rhc.connect(
+        username=test_config.get("candlepin.username"),
+        password=test_config.get("candlepin.password")
+    )
+    status_result = rhc.run("status", "--format", "json", check=False)
+    assert status_result.returncode == 0
+    status_json = json.loads(status_result.stdout)
+    assert "hostname" in status_json
+    assert "rhsm_connected" in status_json
+    assert type(status_json["rhsm_connected"]) == bool
```
Archana-PandeyM, line 57: Thanks George for writing the test. Looks good to me , just one suggestion to check the actual value of the keys along with type of value- something like :
`assert status_json["insights_connected"] == True`

This might test the issue to some extent https://issues.redhat.com/browse/CCT-525

Archana-PandeyM (changes requested):

integration-tests/test_status.py
jirihnidek: We test value in another test. I don't think that it is good to check the same value twice. Why? If we added check for the value and then we will change the value returned from rhc in the future, then we would have to change expected values twice.

jirihnidek: @Archana-PandeyM ^ Do you have any comment to this my comment?

Archana-PandeyM: I agree that if we change the value in rhc later then that will require the test to be updated again, which should be the way to do it. But we cannot miss checking the actual values. I mean what if the value is still 'False' on a connected system. It is integration tests so it should reflect verification of exact values. We could keep both type and value checks.
Otherwise we will need another tests just for checking the values.

PR #168, opened 2024-11-13
feat: register with content templates

rverdile:
Card ID: CCT-769.

This is dependent on [this PR](https://github.com/candlepin/subscription-manager/pull/3462), which the requirements are still being discussed, so I'm leaving this in draft for now.    

This PR implements a new option to register with a content template, given the content template name. Since content templates are implemented as environments in candlepin, I've tried to implement this such that some code could be re-used in the future for registering with environments.

Feedback is appreciated and welcome :)

Still needs

- [x] To handle activation keys
- [x] To handle when GetEnvironments() is not available

### DIFF
--- a/rhsm.go
+++ b/rhsm.go
@@ -17,6 +17,8 @@ import (
 	"github.com/godbus/dbus/v5"
 )
 
+const EnvTypeContentTemplate = "content-template"
+
 func getConsumerUUID() (string, error) {
 	conn, err := dbus.SystemBus()
 	if err != nil {
@@ -85,7 +87,7 @@ func unpackOrgs(s string) ([]string, error) {
 // registerUsernamePassword tries to register system against candlepin server (Red Hat Management Service)
 // username and password are mandatory. When organization is not obtained, then this method
 // returns list of available organization and user can select one organization from the list.
-func registerUsernamePassword(username, password, organization, serverURL string) ([]string, error) {
+func registerUsernamePassword(username, password, organization string, environments []string, serverURL string) ([]string, error) {
 	var orgs []string
 	if serverURL != "" {
 		if err := configureRHSM(serverURL); err != nil {
@@ -132,6 +134,14 @@ func registerUsernamePassword(username, password, organization, serverURL string
 		return orgs, err
 	}
 
+	options := make(map[string]string)
+	if len(environments) != 0 {
+		options["environment_names"] = strings.Join(environments, ",")
+		options["environment_type"] = EnvTypeContentTemplate
+
+	}
+	options["enable_content"] = "true"
+
 	if err := privConn.Object(
 		"com.redhat.RHSM1",
 		"/com/redhat/RHSM1/Register").Call(
@@ -140,7 +150,7 @@ func registerUsernamePassword(username, password, organization, serverURL string
 		organization,
 		username,
 		password,
-		map[string]string{"enable_content": "true"},
+		options,
 		map[string]string{},
 		locale).Err; err != nil {
 
@@ -183,7 +193,7 @@ func registerUsernamePassword(username, password, organization, serverURL string
 	return orgs, nil
 }
 
-func registerActivationKey(orgID string, activationKeys []string, serverURL string) error {
+func registerActivationKey(orgID string, activationKeys []string, environments []string, serverURL string) error {
 	if serverURL != "" {
 		if err := configureRHSM(serverURL); err != nil {
 			return fmt.Errorf("cannot configure RHSM: %w", err)
@@ -229,6 +239,13 @@ func registerActivationKey(orgID string, activationKeys []string, serverURL stri
 		return err
 	}
 
+	options := make(map[string]string)
+	if len(environments) != 0 {
+		options["environment_names"] = strings.Join(environments, ",")
+		options["environment_type"] = EnvTypeContentTemplate
+
+	}
+
 	if err := privConn.Object(
 		"com.redhat.RHSM1",
 		"/com/redhat/RHSM1/Register").Call(
@@ -236,7 +253,7 @@ func registerActivationKey(orgID string, activationKeys []string, serverURL stri
 		dbus.Flags(0),
 		orgID,
 		activationKeys,
-		map[string]string{},
+		options,
 		map[string]string{},
 		locale).Err; err != nil {
 		return unpackRHSMError(err)
@@ -407,6 +424,7 @@ func registerRHSM(ctx *cli.Context) (string, error) {
 		password := ctx.String("password")
 		organization := ctx.String("organization")
 		activationKeys := ctx.StringSlice("activation-key")
+		contentTemplates := ctx.StringSlice("content-template")
 
 		if len(activationKeys) == 0 {
 			if username == "" {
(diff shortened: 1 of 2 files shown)


rhsm.go
```diff
@@ -229,14 +255,17 @@ func registerActivationKey(orgID string, activationKeys []string, serverURL stri
...
+
 	if err := privConn.Object(
 		"com.redhat.RHSM1",
 		"/com/redhat/RHSM1/Register").Call(
 		"com.redhat.RHSM1.Register.RegisterWithActivationKeys",
 		dbus.Flags(0),
 		orgID,
 		activationKeys,
-		map[string]string{},
+		options,
```
jirihnidek, line 268: The registration using activation key & org & content templates does not work.

You cannot do it so simple, when activation key(s) are used for registration. You cannot use environment names as environment IDs. You have to get IDs from environment names in the same way as you do it for registration using username & password.

jirihnidek (changes requested):
Thanks for your PR. Overall, it works. :+1: 

I have to apologize. We agreed that we will support only environment with type == content-template. Thus, it will be possible to simplify your PR on some places. You can find more details in my comments.

I have some other requests.

rhsm.go
```diff
@@ -132,6 +132,32 @@ func registerUsernamePassword(username, password, organization, serverURL string
 		return orgs, err
 	}
 
+	var options = make(map[string]string)
+	options["enable_content"] = "true"
+
+	if len(environments) > 0 && organization != "" {
+		var useEnvIDs = false
+		envList, err := getEnvironmentsList(privConn, username, password, organization)
+		if err != nil {
```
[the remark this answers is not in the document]
rverdile: I agree this is a nice simplification. And now I'm thinking alternating to using the ID would not be useful in the case of content templates. @ptoscano wrote the ticket so he may want to provide some input here.

rhsm.go
rverdile: The GetEnvironments() method requires username and password, so the intention here was to assume that the given values are the IDs. 

But now I am wondering if using activation keys with templates should be prevented? The ID is not really usable in the case of content templates. @ptoscano wrote the ticket he want to provide some input.

jirihnidek: Yeah, you are right. Getting environments requires username and password...

Hmm, I would be easy to simply say that  content templates are supported only with username and password authentication, but user experience would not be the best.

Other approach: requiring environment name for username & password and requiring environment ID (typically some UUID) for activation key authentication would be very very confusing.

ptoscano: The problem here is due to the difference in behaviour of two types of "credentials": username & password, activation key(s) (AKs) + org id:
- username & password are real credentials: they are used when authenticating to Candlepin (via HTTP Basic Auth), and thus the identity of the caller is immediately known even before the code of the actual endpoint is run
- AKs are "credentials", yes, however they are actually metadata passed when calling the `POST /consumers` endpoint. They currently cannot be used for anything else than that

Environments have been historically handled in a "special way": so far they are Katello-only feature, and for reasons I'm not even sure about they were implemented in Katello directly (since Katello proxies the communication to its Candlepin instance) as part of AKs. When connecting directly to Candlepin there is (or better, used to be) no possibility to specify environments with AKs. Hence, if the IDs and names of AKs are different, they don't generally matter in a Katello setup.

Strictly speaking, the org ID has the same inconsistency as environments:
- sub-man does resolve the org name from the available ones when registering with username & password, or even what is passed via `--org`
- sub-man uses `--org` as-is when registering with AKs

The above inconsistency for org ID practically has not been an issue because:
- when registering to Hosted, the org ID and names are the same (numeric strings)
- when registering to Katello, the automated registration ways such as the global registration take care of pre-filling the script to run (and the script uses AKs)
- registration to Katello via username & password is not commonly done

Because of the above, IMHO there is not much that can be done to avoid the inconsistencies from Candlepin; my recommendation is to:
- resolve content templates names -> IDs when registering via username & password, as that is usually done manually (and thus the user better expect user-facing bits)
- assume that the specified content templates are IDs when registering via AKs

jirihnidek: > Because of the above, IMHO there is not much that can be done to avoid the inconsistencies from Candlepin; my recommendation is to:
>
> * resolve content templates names -> IDs when registering via username & password, as that is usually done 
  manually (and thus the user better expect user-facing bits)
> * assume that the specified content templates are IDs when registering via AKs

No, I will never accept anything like this!

jirihnidek: I have been thinking about this problem and I think that cleanest solution is to use environment IDs in both cases (authentication using username & password, authentication using activation-key & org). Could environment IDs be some e.g. UUID. Yes, it could happen. Do users care? Not really, because if environment names were supported, then they would probably copy paste environment names to command line anyway. So, it does not matter, what they copy paste.

jlsherrill: A candlepin dev pointed out that candlepin can natively accept the environment name:

```
Candlepin supports using either environment names or ids during registration. Just pass it like this in the consumer payload:
{
  "environments": [
     {
       "name": "my_environment_name"
       }
    ]
 }
```

Any reason to not go that route?

jirihnidek: It is probably doable, but something like this would be necessary for subscription-manager first: https://github.com/candlepin/subscription-manager/pull/3480

jirihnidek:
Following change in `rhsm.service` https://github.com/candlepin/subscription-manager/pull/3480 would allow to simplify implementation of content template.
1. It will be possible to use environment names directly for both authentication methods
2. It will be possible to specify `environment_type` in argument of `Register()` and `RegisterWithActivationKeys()`
3. Thus, there will be no need to call `GetEnvironments()` D-Bus method and registration will not be blocked by another REST API call.

rverdile:
Made the changes based on the new subscription manager PR!

jirihnidek:
@rverdile I just merged https://github.com/candlepin/subscription-manager/pull/3480 to main ... I will review your PR tomorrow.

rhsm.go
```diff
@@ -229,14 +239,21 @@ func registerActivationKey(orgID string, activationKeys []string, serverURL stri
 		return err
 	}
 
+	options := make(map[string]string)
+	if len(environments) != 0 {
+		options["environment_names"] = strings.Join(environments, ",")
+		options["environment_type"] = environmentType
```
jirihnidek, line 245: I think that it is not necessary to pass `environmentType` as an argument to this function and you can use:

```go
options["environment_type"] = EnvTypeContentTemplate
```

because we will use only `"content-template"`.

rhsm.go
```diff
@@ -132,6 +134,14 @@ func registerUsernamePassword(username, password, organization, serverURL string
 		return orgs, err
 	}
 
+	options := make(map[string]string)
+	if len(environments) != 0 {
+		options["environment_names"] = strings.Join(environments, ",")
+		options["environment_type"] = environmentType
```
jirihnidek, line 140: I think that it is not necessary to pass `environmentType` as an argument to this function and you can use:

```go
options["environment_type"] = EnvTypeContentTemplate
```

because we will use only `"content-template"`.

.gitignore
```diff
@@ -7,3 +7,4 @@ BUILDROOT
 RPMS
 SOURCES
 SRPMS
+.idea
```
jirihnidek, line 10: Please do not add this line to `.gitignore` as part of this PR. Such change is not related to this PR.

jirihnidek (changes requested):
Thanks for the update :+1: . It looks good. We are almost there. Can you please rebase your PR? We started to use meson build system and your branch does not reflect it. I also have three small requests.

rverdile:
updated, thanks!

jirihnidek (commented):
LGTM

Issue #88, opened 2024-01-12
Use log/slog instead of go-log


ahitacat:
https://github.com/subpop/go-log is deprecated. `rhc` should start using log/slog instead.

jirihnidek:
Useful links:

DuckBoss:
This issue has been resolved with the merged PR: https://github.com/RedHatInsights/rhc/pull/241

PR #170, opened 2024-11-21
feat: Add support of configuring host to use Satellite server

jirihnidek:
* Card ID: CCT-993
* Introduced new command "configure" and immediately added sub-command "satellite" to this command. "satellite" subcommand has --url CLI option and you can set hostname or URL of server Satellite server.
  * When valid URL or hostname is provided, then rhc tries to read /api/ping endpoint and it tries to verify that given server is Satellite server
  * When Satellite server is verified, then it tries to download bootstrap script from /pub/katello-rhsm-consumer
  * When script is downloaded, then this script is run as root user.
* Is little bit risky to run some script downloaded from web server. For this reason we will try to use different approach in the future. Satellite server will provide public REST API endpoints providing CA certs and rendered rhsm.conf. We will simply download these files and install them to system. Unfortunately, these REST API endpoints will be available in some future version of Satellite.

### DIFF
--- a/configure_cmd.go
+++ b/configure_cmd.go
@@ -1,0 +1,144 @@
+package main
+
+import (
+	"fmt"
+	"github.com/briandowns/spinner"
+	"github.com/urfave/cli/v2"
+	"os"
+	"os/exec"
+	"time"
+)
+
+// beforeSatelliteAction is run before satelliteAction is run. It is used
+// for checking CLI options.
+func beforeSatelliteAction(ctx *cli.Context) error {
+	// First check if machine-readable format is used
+	err := setupFormatOption(ctx)
+	if err != nil {
+		return err
+	}
+
+	satelliteUrlStr := ctx.String("url")
+	if satelliteUrlStr == "" {
+		return fmt.Errorf("no url provided using --url CLI option")
+	}
+
+	uuid, err := getConsumerUUID()
+	if err != nil {
+		return fmt.Errorf("unable to get consumer UUID: %s", err)
+	}
+	if uuid != "" {
+		return fmt.Errorf("cannot configure connected system to use satellite; disconnect system firtst")
+	}
+
+	return checkForUnknownArgs(ctx)
+}
+
+// satelliteAction tries to get bootstrap script from Satellite server and run it.
+// When it is not possible to download the script or running script returns
+// non-zero exit code, then error is returned.
+//
+// It is really risky to download to run some script downloaded from the URL as a
+// root user without any restriction. For this reason, we at least check that
+// provided URL is URL of Satellite server.
+//
+// We would like to use different approach in the future. We would like to use
+// some API endpoints not restricted by username & password for getting CA certs
+// and rendered rhsm.conf, because it would be more secure, but it is not possible ATM
+func satelliteAction(ctx *cli.Context) error {
+	var configureSatelliteResult ConfigureSatelliteResult
+	configureSatelliteResult.format = ctx.String("format")
+	satelliteUrlStr := ctx.String("url")
+
+	uid := os.Getuid()
+	if uid != 0 {
+		errMsg := "non-root user cannot connect system"
+		exitCode := 1
+		return cli.Exit(fmt.Errorf("error: %satSpinner", errMsg), exitCode)
+	}
+
+	hostname, err := os.Hostname()
+	if uiSettings.isMachineReadable {
+		configureSatelliteResult.Hostname = hostname
+	}
+	if err != nil {
+		exitCode := 1
+		if uiSettings.isMachineReadable {
+			configureSatelliteResult.HostnameError = err.Error()
+			return cli.Exit(configureSatelliteResult, exitCode)
+		} else {
+			return cli.Exit(fmt.Errorf("could not acquire hostname: %w", err), exitCode)
+		}
+	}
+
+	satelliteUrl, err := normalizeSatelliteScriptUrl(satelliteUrlStr)
+	if err != nil {
+		return cli.Exit(fmt.Errorf("could not parse satellite url: %w", err), 1)
+	}
+
+	configureSatelliteResult.SatelliteServerHostname = satelliteUrl.Hostname()
+	configureSatelliteResult.SatelliteServerScriptUrl = satelliteUrl.String()
+
+	var satSpinner *spinner.Spinner = nil
+	if uiSettings.isRich {
+		satSpinner = spinner.New(spinner.CharSets[9], 100*time.Millisecond)
+		satSpinner.Suffix = fmt.Sprintf(" Configuring '%v' to use Satellite %v", hostname, satelliteUrl.Host)
+		satSpinner.Start()
+		// Stop spinner after running function
+		defer func() { satSpinner.Stop() }()
+	}
+
+	if satSpinner != nil {
+		satSpinner.Suffix = fmt.Sprintf(" Connecting to Satellite server: %v", satelliteUrl.Host)
+	}
+
+	satClient := NewSatelliteClient(satelliteUrl)
+	_, err = satClient.Ping()
+	if err != nil {
+		return cli.Exit(fmt.Errorf("unable to verify that given server is Satellite server: %v", err), 1)
+	}
+
+	configureSatelliteResult.IsServerSatellite = true
+
+	if satSpinner != nil {
+		satSpinner.Suffix = fmt.Sprintf(" Downloading configuration from %v", satelliteUrl.Host)
+	}
+	satelliteScriptPath, err := satClient.downloadScript(ctx)
+	if err != nil {
+		return cli.Exit(fmt.Errorf("could not download satellite bootstrap script: %w", err), 1)
+	}
+
+	if ctx.Bool("keep-artifacts") != true {
+		defer func() {
+			// TODO: If error happens, then log this error
+			_ = os.Remove(*satelliteScriptPath)
+		}()
+	}
+
+	if satSpinner != nil {
+		satSpinner.Suffix = fmt.Sprintf(
+			" Configuring '%v' to use Satellite server: %v",
+			hostname,
+			satelliteUrl.Host,
+		)
+	}
+
+	// Run the bash script. It should install CA certificate, change configuration of rhsm.conf.
+	// In theory, it can do almost anything.
+	cmd := exec.Command("/usr/bin/bash", *satelliteScriptPath)
+	err = cmd.Run()
+	if err != nil {
+		return cli.Exit(fmt.Errorf("execution of %v script failed: %w", satelliteScriptPath, err), 1)
+	}
+
+	configureSatelliteResult.HostConfigured = true
+
+	if uiSettings.isRich {
+		satSpinner.Suffix = ""
+		satSpinner.Stop()
+	}
+
+	interactivePrintf("Host '%v' configured to use Satellite server: %v\n", hostname, satelliteUrl.Host)
+
+	return cli.Exit(configureSatelliteResult, 0)
+}
--- a/constants.go
+++ b/constants.go
@@ -39,6 +39,7 @@ var (
 	SysconfDir        string
 	LocalstateDir     string
 	DbusInterfacesDir string
+	VarLibDir         string
 )
 
 func init() {
@@ -74,6 +75,9 @@ func init() {
 	}
 	if DbusInterfacesDir == "" {
 		DbusInterfacesDir = filepath.Join(DataDir, "dbus-1", "interfaces")
+	}
+	if VarLibDir == "" {
+		VarLibDir = filepath.Join(LocalstateDir, "lib", "rhc")
 	}
 
 	if ShortName == "" {
--- a/main.go
+++ b/main.go
@@ -175,6 +175,40 @@ func main() {
 			Action:      connectAction,
 		},
 		{
+			Name:      "configure",
+			Usage:     "Configure the host",
+			UsageText: fmt.Sprintf("%v configure [sub-command]", app.Name),
+			Subcommands: []*cli.Command{
+				{
+					Name:        "satellite",
+					Usage:       "Configure the host to use with Satellite server",
+					UsageText:   fmt.Sprintf("%v configure satellite [command options]", app.Name),
+					Description: "The satellite sub-command configure the system for connection to a Satellite server",
+					Flags: []cli.Flag{
+						&cli.StringFlag{
+							Name:    "url",
+							Usage:   "URL of the Satellite server",
+							Aliases: []string{"u"},
+						},
+						&cli.StringFlag{
+							Name:    "format",
+							Usage:   "prints output of satellite configure in machine-readable format (supported formats: \"json\")",
+							Aliases: []string{"f"},
+						},
+						&cli.BoolFlag{
+							Name:    "keep-artifacts",
+							Usage:   "Keep artifacts (downloaded bootstrap script)",
+							Hidden:  true,
+							Value:   false,
+							Aliases: []string{"k"},
+						},
+					},
+					Before: beforeSatelliteAction,
+					Action: satelliteAction,
+				},
+			},
+		},
+		{
 			Name: "disconnect",
 			Flags: []cli.Flag{
 				&cli.StringFlag{
--- a/satellite.go
+++ b/satellite.go
@@ -1,0 +1,212 @@
+package main
+
+import (
+	"crypto/tls"
+	"encoding/json"
+	"fmt"
+	"github.com/urfave/cli/v2"
+	"io"
+	"net/http"
+	"net/url"
+	"os"
+	"path/filepath"
+)
+
+// SatellitePingResponse is used for un-marshaling JSON document
+// returned from Satellite server
+type SatellitePingResponse struct {
+	Results struct {
+		Katello struct {
+			Services struct {
+				Candlepin struct {
+					Status     string `json:"status"`
+					DurationMs string `json:"duration_ms"`
+				} `json:"candlepin"`
+			} `json:"services"`
+			Status string `json:"status"`
+		} `json:"katello"`
+	} `json:"results"`
+}
+
+// ConfigureSatelliteResult is structure holding information about results
+// of configuring host from Satellite server. The result could be printed
+// in machine-readable format.
+type ConfigureSatelliteResult struct {
+	Hostname                 string `json:"hostname"`
+	HostnameError            string `json:"hostname_error,omitempty"`
+	IsServerSatellite        bool   `json:"is_server_satellite_running"`
+	SatelliteServerHostname  string `json:"satellite_server_hostname"`
+	SatelliteServerScriptUrl string `json:"satellite_server_script_url"`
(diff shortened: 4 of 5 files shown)


configure_cmd.go
```diff
@@ -0,0 +1,133 @@
...
+	satelliteScriptPath := path.Join(varLibDir, "katello-rhsm-consumer")
+	defer os.Remove(satelliteScriptPath)
+	scriptFile, err := os.Create(satelliteScriptPath)
+	if err != nil {
+		return fmt.Errorf("could not create %v file: %w", satelliteScriptPath, err)
+	}
+	defer scriptFile.Close()
+	err = os.Chmod(satelliteScriptPath, 0700)
+	if err != nil {
+		return fmt.Errorf("could not set permissions on %v file: %w", satelliteScriptPath, err)
```
subpop, line 104: This should be wrapped in a `cli.Exit`.

configure_cmd.go
```diff
@@ -0,0 +1,133 @@
...
+	}
+
+	err = downloadSatelliteScript(scriptFile, satelliteUrl, s)
+	if err != nil {
+		return fmt.Errorf("could not download satellite bootstrap script: %w", err)
```
subpop, line 109: This should be wrapped in a `cli.Exit`.

configure_cmd.go
```diff
@@ -0,0 +1,133 @@
...
+	}
+
+	if s != nil {
```
subpop, line 112: Reading over the code, I reached this point and thought to myself "what's `s`?" That means this variable name needs to be more descriptive. It should probably be named `spinner`. Using single letter variable names is convenient, but it's important to consider the lifetime of the variable. If it's going to be long-lived, it's probably better to give it a descriptive name.

configure_cmd.go
```diff
@@ -0,0 +1,133 @@
...
+		s.Suffix = fmt.Sprintf(" Configuring '%v' to use Satellite server: %v", hostname, satelliteUrl.Host)
+	}
+	// Run the script. It should install CA certificate, change configuration of rhsm.conf.
+	// In theory, it can do almost anything.
+	cmd := exec.Command(satelliteScriptPath)
+	err = cmd.Run()
+	if err != nil {
+		return fmt.Errorf("execution of %v script failed: %w", satelliteScriptPath, err)
```
subpop, line 120: This should be wrapped in a `cli.Exit`.

configure_cmd.go
```diff
@@ -0,0 +1,133 @@
...
+	} else {
```
subpop, line 121: This else clause can be dropped and leave the `configureSatelliteResult.HostConfigured = true` line in the main function body.

configure_cmd.go
```diff
@@ -0,0 +1,133 @@
...
+	err = downloadSatelliteScript(scriptFile, satelliteUrl, s)
+	if err != nil {
+		return fmt.Errorf("could not download satellite bootstrap script: %w", err)
+	}
+
+	if s != nil {
+		s.Suffix = fmt.Sprintf(" Configuring '%v' to use Satellite server: %v", hostname, satelliteUrl.Host)
+	}
+	// Run the script. It should install CA certificate, change configuration of rhsm.conf.
+	// In theory, it can do almost anything.
```
subpop, line 116: Scary. We should follow this up with a distributed SELinux policy to restrict the scope execution for `rhc` in general.

subpop (changes requested):
I have not read over `satellite.go` yet. I'll get to that in a follow up review. I left a scattering of comments through the other files. I have not tested any of the functionality yet. This is just from a reading over of the code.

configure_cmd.go
```diff
@@ -0,0 +1,133 @@
...
+	"github.com/briandowns/spinner"
+	"github.com/urfave/cli/v2"
+	"os"
+	"os/exec"
+	"path"
+	"time"
+)
+
+const (
+	varLibDir = "/var/lib/rhc"
```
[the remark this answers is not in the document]
jirihnidek: Yeah, I did want to do it in this way and then I forgot to do it. Thanks for reminder.

jirihnidek:
> I would also like to take a UX pass over the strings in the new command. We should ensure consistent terminology as much as we can.

OK, I updated the PR.

chris1984:
Hi, Can we be sure to support multiple activation keys when registering to Satellite?

jirihnidek:
> Hi, Can we be sure to support multiple activation keys when registering to Satellite?

Hi Chris, `rhc connect` already supports registration using multiple activation keys.

You can use CLI option `--activation-key` multiple times:

```
[root@rhel10 ~]# rhc connect --activation-key KEY-01 --activation-key KEY-02 --organization MY-ORG
```

It is also emphasize in the help of `connect` command:

```
[root@rhel10 ~]# rhc connect --help
NAME:
   rhc connect - Connects the system to Red Hat

USAGE:
   rhc connect [command options]

DESCRIPTION:
   The connect command connects the system to Red Hat Subscription Management, Red Hat Insights and Red Hat and activates the yggdrasil service that enables Red Hat to interact with the system. For details visit: https://red.ht/connector

OPTIONS:
   --username USERNAME, -u USERNAME                               register with USERNAME
   --password PASSWORD, -p PASSWORD                               register with PASSWORD
   --organization ID, -o ID                                       register with ID
   --activation-key KEY, -a KEY [ --activation-key KEY, -a KEY ]  register with KEY
   --format value, -f value                                       prints output of connection in machine-readable format (supported formats: "json")
   --help, -h                                                     show help
```

configure_cmd.go
```diff
@@ -0,0 +1,130 @@
...
+		return cli.Exit(fmt.Errorf("could not download satellite bootstrap script: %w", err), 1)
+	}
+
+	if satSpinner != nil {
+		satSpinner.Suffix = fmt.Sprintf(" Configuring '%v' to use Satellite server: %v", hostname, satelliteUrl.Host)
+	}
+	// Run the script. It should install CA certificate, change configuration of rhsm.conf.
+	// In theory, it can do almost anything.
+	cmd := exec.Command(satelliteScriptPath)
+	err = cmd.Run()
```
subpop, line 115: I'm thinking ahead here toward the ability to troubleshoot this execution when
things go wrong. Should we log the stdout and stderr of the run to a file
somewhere? I don't think it's necessary to pipe the stdout and stderr into this
parent process, but logging the script's output might be helpful.

satellite.go
```diff
@@ -0,0 +1,214 @@
...
+	"encoding/json"
+	"fmt"
+	"github.com/briandowns/spinner"
+	"io"
+	"net/http"
+	"net/url"
+	"os"
+)
+
+type SatellitePingResponse struct {
```
subpop, line 14: This is a big structure, and the HTTP API does not document this object at all.
If these fields are not required as part of the handling of the ping response,
it's safer to not try to unmarshal them at all.

satellite.go
```diff
@@ -0,0 +1,214 @@
...
+	HostnameError            string `json:"hostname_error,omitempty"`
+	IsServerSatellite        bool   `json:"is_server_satellite_running"`
+	SatelliteServerHostname  string `json:"satellite_server_hostname"`
+	SatelliteServerScriptUrl string `json:"satellite_server_script_url"`
+	HostConfigured           bool   `json:"host_configured"`
+	format                   string
+}
+
+// Error implement error interface for structure ConfigureSatelliteResult
+func (configureSatelliteResult ConfigureSatelliteResult) Error() string {
```
subpop, line 80: There is a Go convention to refer to the 'self' object in a method as a single
or small variable name. I think this could be made a bit more concise by
replacing `configureSatelliteResult` with `result`. Maybe then rename the
existing `result` string to `message` or something like that.

satellite.go
```diff
@@ -0,0 +1,214 @@
...
+	satellitePingUrl.RawQuery = ""
+
+	// We have to use insecure HTTPs connection, because most of the customers use
+	// self-signed certificates
+	tlsConfig := tls.Config{
+		InsecureSkipVerify: true,
+	}
+	transport := http.DefaultTransport.(*http.Transport).Clone()
+	transport.TLSClientConfig = &tlsConfig
+	client := &http.Client{Transport: transport}
```
subpop, line 127: Is this same client (with its TLS configuration) not required to download the
script in `downloadSatelliteScript`?

satellite.go
```diff
@@ -0,0 +1,214 @@
...
+		return fmt.Errorf("unable to read body of ping satellite server response: %v", err)
+	}
+
+	satellitePingResponse := SatellitePingResponse{}
+	err = json.Unmarshal(resBody, &satellitePingResponse)
+	if err != nil {
+		return fmt.Errorf("unable to parse ping satellite server response: %v", err)
+	}
+
+	// In theory, we could check for the status of candlepin and candlepin auth, etc.,
```
subpop, line 148: Is it safe to assume that if the HTTP server responds with 200, that's an
appropriate response to a "ping" request?

subpop (changes requested):
The `satellite.go` code feels disjointed. I think we could clean up the function pattern by making a dedicated HTTP client and building some methods on that. This way the two requests we use today (ping and /pub/katello-rhsm-consumer) can share the same HTTP client (TLS configuration, base URL, etc).

Then this client can be instantiated from within `satelliteAction` and used to perform the various operations it wants to.

The ping method feels incomplete too. If the response body is not documented on the Internet somewhere, I don't like unmarshalling all those fields, especially if they are not used. We could probably get away with only a status field, if we're checking for liveness by sending the server a ping.

configure_cmd.go
jirihnidek: I agree that it would be useful to log more information from `rhc`. Problem is that we do not have any logging system for `rhc`. Should we address this issue after solving this issue https://github.com/RedHatInsights/rhc/issues/88 ?

satellite.go
jirihnidek: Yeah, I wasn't able to find any documentation for output of this REST API call (https://satellite.sat.engineering.redhat.com/apidoc/v2/ping/ping.en.html). I think it would be worth asking Satellite developers how stable format of this object is.

I would like to emphasize that this ping is only temporary solution that should avoid calling `rhc configure satellite --url <some.url>` against servers that are not satellite servers. More checks we do the better it is. We will replace this solution with calling different REST API call in hopefully near future. I believe that there is low risk that structure of returned object will change soon.

What about reducing this structure to something minimalist like this:

```go
type SatellitePingResponse struct {
	Results struct {
		Katello struct {
			Services struct {
				Candlepin struct {
					Status     string `json:"status"`
					DurationMs string `json:"duration_ms"`
				} `json:"candlepin"`
			} `json:"services"`
			Status string `json:"status"`
		} `json:"katello"`
	} `json:"results"`
}
```

satellite.go
jirihnidek: It seems that it is possible to access `/pub` using (http and https). It is possible to access `/api/ping` using only https. In theory we can try to use one client using only one https connection.

satellite.go
jirihnidek: 1) I can delete this commented block of code.
2) In theory we can also remove `json.Unmarshal()` of the body, but as I already said, it is safer to do more checks that the server is really Satellite server than do less checks, because we run downloaded script as a root.

jirihnidek: If user provide full URL to some no-satellite script and there will be also REST API endpoint `/api/ping` (not so unlikely), then the script will be downloaded and run as a root user. For this reason I would rather check that returned object of `/api/ping` is really response from Satellite.

subpop (commented):
> > The `satellite.go` code feels disjointed. I think we could clean up the function pattern by making a dedicated HTTP client and building some methods on that. This way the two requests we use today (ping and /pub/katello-rhsm-consumer) can share the same HTTP client (TLS configuration, base URL, etc).
> 
> Do you mean to reuse this https://github.com/RedHatInsights/rhc/blob/main/internal/http/client.go again? It seems that this file is unused ATM. It will probably require to extend this file to be able to use it for communication with Satellite server. 

Maybe not reuse it directly, but create a `SatelliteHTTPClient` type that preconfigures the TLS configuration used by a specific backing HTTP client, and that has methods attached to it to make the REST API calls.

```
type SatelliteHTTPClient struct {
    client *http.Client
}

func (s *SatelliteHTTPClient) Ping() (SatelliteHTTPPingResponse, error) { ... }
```

It might seem like a lot more boilerplate code up front, but I think it will position this command to evolve more Satellite features in the future without needing to refactor things.

configure_cmd.go
subpop: I agree that deferring it out of this PR is the right decision, since this PR is
already sizable. But I'm concerned that we won't prioritize it enough in a
followup PR.

satellite.go
subpop: Reducing it is important. I think that will reduce the risk of unmarshalling
errors that can turn up with a change to the HTTP response.

satellite.go
subpop: Yea, I think we can consolidate the HTTP clients into a specialized one for
performing Satellite REST requests.

satellite.go
subpop: That makes sense to me. Considering the risk we're taking here (downloading a
shell script and running it as root), we should do as much as we can to ensure
that this script is coming from a trusted source.

constants.go
```diff
@@ -39,6 +39,7 @@ var (
 	SysconfDir        string
 	LocalstateDir     string
 	DbusInterfacesDir string
+	VarLibDir         string
```
subpop, line 42: There's something about this name that doesn't sit right with me. "VarLibDir" makes me think `/var/lib`, not `/var/lib/rhc`. I'd rather we name it something more accurate, like `RhcLocalstateDir`.

satellite.go
```diff
@@ -0,0 +1,218 @@
...
+
+// NewSatelliteClient creates instance of SatelliteHTTPClient and
+// configure it to use HTTPS
+func NewSatelliteClient(satelliteURL *url.URL) *SatelliteHTTPClient {
+	satClient := SatelliteHTTPClient{}
+	// We have to use insecure HTTPs connection, because most of the customers use
+	// self-signed certificates
+	tlsConfig := tls.Config{
+		InsecureSkipVerify: true,
+	}
```
subpop, line 77: So we're downloading a shell script from a URL and executing it as root, *and* we're downloading that shell script from an unsigned HTTPS endpoint. 😱 

Do customers actually use self-signed certificates? Or do they use certificates that are signed by an internal/private signing authority? Could we expose the option to add a CA root certificate as part of this TLS configuration so that customers can at least use their own signing authority? And maybe make them opt-in to fully disabling the certificate verification?

satellite.go
```diff
@@ -0,0 +1,218 @@
...
+	satellitePingResponse := SatellitePingResponse{}
+	err = json.Unmarshal(resBody, &satellitePingResponse)
+	if err != nil {
+		return nil, fmt.Errorf("unable to parse ping satellite server response: %v", err)
+	}
+
+	return &satellitePingResponse, nil
+}
+
+// downloadSatelliteScript tries to download script from provided URL to given file
```
subpop, line 129: This method documentation looks like a leftover from the previous implementation. It doesn't match the current method name and it describes a parameter that it no longer accepts.

```suggestion
// downloadScript tries to download script from provided URL to given file
```

configure_cmd.go
```diff
@@ -0,0 +1,137 @@
...
+// for checking CLI options.
+func beforeSatelliteAction(ctx *cli.Context) error {
+	// First check if machine-readable format is used
+	err := setupFormatOption(ctx)
+	if err != nil {
+		return err
+	}
+
+	satelliteUrlStr := ctx.String("url")
+	if satelliteUrlStr == "" {
```
subpop, line 22: We can mark the `cli.Flag` defining this flag as `Required: true`. That makes this check unnecessary.

main.go
```diff
@@ -174,6 +174,40 @@ func main() {
...
+				{
+					Name:        "satellite",
+					Usage:       "Configure the host to use with Satellite server",
+					UsageText:   fmt.Sprintf("%v configure satellite [command options]", app.Name),
+					Description: "The satellite sub-command configure the system for connection to a Satellite server",
+					Flags: []cli.Flag{
+						&cli.StringFlag{
+							Name:    "url",
+							Usage:   "URL of the Satellite server",
+							Aliases: []string{"u"},
```
subpop, line 191: This flag should be marked as required:

```suggestion
							Aliases: []string{"u"},
							Required: true,
```

configure_cmd.go
```diff
@@ -0,0 +1,137 @@
...
+
+	uuid, err := getConsumerUUID()
+	if err != nil {
+		return fmt.Errorf("unable to get consumer UUID: %s", err)
+	}
+	if uuid != "" {
+		return fmt.Errorf("cannot configure connected system to use satellite; disconnect system firtst")
+	}
+
+	return checkForUnknownArgs(ctx)
```
subpop, line 34: I don't know if I've commented on this before, but I like this pattern of returning a call through another function. This function is named very well and this clearly describes what happens through this function. Nicely written.

subpop (changes requested):
There's a lint error caught by the linting actions I noticed, and I left a few comments here and there. I'll get to some functional testing shortly.

main.go
```diff
@@ -174,6 +174,40 @@ func main() {
 			Before:      beforeConnectAction,
 			Action:      connectAction,
 		},
+		{
+			Name:      "configure",
+			Usage:     "Configure the host",
```
subpop, line 179: Looking at this text surrounded by the other text in the help output, I think
this needs to change to something like:

"Configures the behavior of the system"

```
COMMANDS:
   connect     Connects the system to Red Hat
   configure   Configure the host
   disconnect  Disconnects the system from Red Hat
   status      Prints status of the system's connection to Red Hat
   help, h     Shows a list of commands or help for one command
```

subpop (commented):
Some thoughts on the user experience.

I definintely don't like the insecure certificate experience. I first set up my instance like so: `rhc configure satellite --url https://my-satelite.local` and got the error back:

```
could not download satellite bootstrap script: could not download file https://my-satellite.local/pub/katello-rhsm-consumer : Get "https://my-satellite.local/pub/katello-rhsm-consumer": tls: failed to verify certificate: x509: certificate signed by unknown authority
```

We need to expect users will encounter this, and should provide two possible solutions:

1) The ability to add a certificate authority to the TLS configuration so that the server certificate is valid (`--ca-root`)
2) The ability to opt-out of all security (`--insecure`)

I'm not pleased with the idea of a (2), but I know our options are limited with the scope of host identification we get from Satellite currently. That being said, if we are going to allow for the insecure downloading of content, Id' much rather we do so with explicit opt-in from the user.

My execution of the script failed. All I got on the console was: `execution of 0x40001eeab0 script failed: exit status 127` Unless we know this script is robust enough to never fail, we really need to provide a means for users to diagnose these errors. At the very least, let's pipe the stdout and stderr of the script to a log file. This error message isn't descriptive enough for a user to start troubleshooting.

jirihnidek:
We agreed to rework this PR... Moving back to draft.

PR #171, opened 2024-11-22
feat[PoC]: Enable analytics on demand

m-horky:
* Card ID: CCT-941

This patch explores the possibility of using 'configure features' subcommand to perform rhc configuration to provide more options to connecting users. It shouldn't be merged as-is, as it contains features that may conflict with current use-cases and may not handle all of them correctly.

### DIFF
--- a/feature.go
+++ b/feature.go
@@ -1,0 +1,182 @@
+package main
+
+import (
+	"encoding/json"
+	"fmt"
+	"os"
+
+	"github.com/subpop/go-log"
+	"github.com/urfave/cli/v2"
+)
+
+// FeaturesStatePath points to a filesystem location where the state file is stored.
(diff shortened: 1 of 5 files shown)


m-horky:
> I would not display enabled/disabled features in the output of --help

Agreed, it was just an idea.

> I would display enabled/disabled features in the output --show

Sounds good.

> It will be necessary to implement support for machine readable format for all use cases

Yes. I haven't done that for this PoC.

> The identity should not be feature. It is IMHO something must have for everything else.Thus, It does not make sense to have it as special feature.

Yes, agreed as well. The actual initial implementation will likely use content as the core feature. Here I wanted to experiment a bit with separating the two, but it wouldn't work for the timeline we're targeting. Maybe in the future :)

PR #179, opened 2024-12-09
feat: Added levels of connectivity to connect command

jirihnidek:
* Card ID: CCT-1005
* First step of levels of connectivity
* When connect command is used, then it is possible to
  enabled or disable three features using command line
  options `--enable-feature` or `--disable-feature`. Following
  features are supported
  - `content`
  - `analytics`
  - `remote-management`
* Added checks for allowed combination of CLI options
* Modified UI little bit to be able to visualize that
  feature was not started
* Output to JSON format is still supported
* When machine readable format is used, then add information
  about enabled/disabled content to generated JSON file
* When machine readable output is used, then list of
   enabled/disable features is added to the generated JSON doc
* Unified UI for colorful and not colorful mode
* Changed UI for `disconnect` command to looks similar to `connect` command
* Connecting looks like this in interactive mode:

```
Connecting localhost to Red Hat.

Features preferences: [✓]content, [✓]analytics, [✓]remote-management

 [✓] This system is already connected to Red Hat Subscription Management
  [✓] Analytics ... Connected to Red Hat Insights
  [✓] Remote Management ... Activated the yggdrasil service

```

### DIFF
--- a/connect_cmd.go
+++ b/connect_cmd.go
@@ -3,26 +3,35 @@ package main
 import (
 	"encoding/json"
 	"fmt"
+	"os"
+	"strings"
+	"time"
+
 	"github.com/subpop/go-log"
 	"github.com/urfave/cli/v2"
-	"os"
-	"time"
 )
+
+type FeatureResult struct {
+	Enabled    bool   `json:"enabled"`
+	Successful bool   `json:"successful"`
+	Error      string `json:"error,omitempty"`
+}
 
 // ConnectResult is structure holding information about results
 // of connect command. The result could be printed in machine-readable format.
 type ConnectResult struct {
-	Hostname              string `json:"hostname"`
-	HostnameError         string `json:"hostname_error,omitempty"`
-	UID                   int    `json:"uid"`
-	UIDError              string `json:"uid_error,omitempty"`
-	RHSMConnected         bool   `json:"rhsm_connected"`
-	RHSMConnectError      string `json:"rhsm_connect_error,omitempty"`
-	InsightsConnected     bool   `json:"insights_connected"`
-	InsightsError         string `json:"insights_connect_error,omitempty"`
-	YggdrasilStarted      bool   `json:"yggdrasil_started"`
-	YggdrasilStartedError string `json:"yggdrasil_started_error,omitempty"`
-	format                string
+	Hostname         string `json:"hostname"`
+	HostnameError    string `json:"hostname_error,omitempty"`
+	UID              int    `json:"uid"`
+	UIDError         string `json:"uid_error,omitempty"`
+	RHSMConnected    bool   `json:"rhsm_connected"`
+	RHSMConnectError string `json:"rhsm_connect_error,omitempty"`
+	Features         struct {
+		Content          FeatureResult `json:"content"`
+		Analytics        FeatureResult `json:"analytics"`
+		RemoteManagement FeatureResult `json:"remote_management"`
+	} `json:"features"`
+	format string
 }
 
 // Error implement error interface for structure ConnectResult
@@ -56,6 +65,8 @@ func beforeConnectAction(ctx *cli.Context) error {
 	password := ctx.String("password")
 	organization := ctx.String("organization")
 	activationKeys := ctx.StringSlice("activation-key")
+	enabledFeatures := ctx.StringSlice("enable-feature")
+	disabledFeatures := ctx.StringSlice("disable-feature")
 
 	if len(activationKeys) > 0 {
 		if username != "" {
@@ -71,6 +82,11 @@ func beforeConnectAction(ctx *cli.Context) error {
 		if username == "" || password == "" {
 			return fmt.Errorf("--username/--password or --organization/--activation-key are required when a machine-readable format is used")
 		}
+	}
+
+	err = checkFeatureInput(&enabledFeatures, &disabledFeatures)
+	if err != nil {
+		return err
 	}
 
 	return checkForUnknownArgs(ctx)
@@ -112,13 +128,44 @@ func connectAction(ctx *cli.Context) error {
 
 	interactivePrintf("Connecting %v to %v.\nThis might take a few seconds.\n\n", hostname, Provider)
 
+	var featuresStr []string
+	for _, feature := range KnownFeatures {
+		if feature.Enabled {
+			if uiSettings.isMachineReadable {
+				switch feature.ID {
+				case "content":
+					connectResult.Features.Content.Enabled = true
+				case "analytics":
+					connectResult.Features.Analytics.Enabled = true
+				case "remote-management":
+					connectResult.Features.RemoteManagement.Enabled = true
+				}
+			}
+			featuresStr = append(featuresStr, "["+symbolOK+"]"+feature.ID)
+		} else {
+			if uiSettings.isMachineReadable {
+				switch feature.ID {
+				case "content":
+					connectResult.Features.Content.Enabled = false
+				case "analytics":
+					connectResult.Features.Analytics.Enabled = false
+				case "remote-management":
+					connectResult.Features.RemoteManagement.Enabled = false
+				}
+			}
+			featuresStr = append(featuresStr, "[ ]"+feature.ID)
+		}
+	}
+	featuresListStr := strings.Join(featuresStr, ", ")
+	interactivePrintf("Features preferences: %s\n\n", featuresListStr)
+
 	var start time.Time
 	durations := make(map[string]time.Duration)
 	errorMessages := make(map[string]LogMessage)
 	/* 1. Register to RHSM, because we need to get consumer certificate. This blocks following action */
 	start = time.Now()
 	var returnedMsg string
-	returnedMsg, err = registerRHSM(ctx)
+	returnedMsg, err = registerRHSM(ctx, ContentFeature.Enabled)
 	if err != nil {
 		connectResult.RHSMConnected = false
 		errorMessages["rhsm"] = LogMessage{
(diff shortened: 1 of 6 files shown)


jirihnidek:
Open question:
* How should we list enabled/disabled features in generated JSON document?
TODO:
* Allow to enabled/disable features in interactive mode, when user has to input username & password

connect_cmd.go
```diff
@@ -112,13 +124,30 @@ func connectAction(ctx *cli.Context) error {
 
 	interactivePrintf("Connecting %v to %v.\nThis might take a few seconds.\n\n", hostname, Provider)
 
+	var featuresStr []string
+	for _, feature := range KnownFeatures {
+		if feature.Enabled {
+			if uiSettings.isMachineReadable {
+				connectResult.EnabledFeatures = append(connectResult.EnabledFeatures, feature.ID)
+			}
+			featuresStr = append(featuresStr, "["+symbolOK+"]"+feature.ID)
```
m-horky, line 133: I think there should be a space between the checkbox and the text.

connect_cmd.go
```diff
@@ -112,13 +124,30 @@ func connectAction(ctx *cli.Context) error {
...
+		} else {
+			if uiSettings.isMachineReadable {
+				connectResult.DisabledFeatures = append(connectResult.EnabledFeatures, feature.ID)
+			}
+			featuresStr = append(featuresStr, "[ ]"+feature.ID)
```
m-horky, line 138: I think there should be a space between the checkbox and the text.

m-horky (changes requested):
The patch seems solid. I won't comment much on the code itself; we'll likely see further changes done in near future as we stabilize it before new features come.

1.

The output seems to jump around, were the two spaces intentional?

```
$ rhc connect --disable-feature analytics
...
 [ ] Analytics ... Connecting to Red Hat Insights disabled
  [✓] Remote Management ... Activated the yggdrasil service
```

2.

When disconnecting, Insights reported status 1 even though we should have detected we're (already) unregistered:

```
 [✓] Deactivated the yggdrasil service
 [𐄂] Cannot disconnect from Red Hat Insights: exit status 1
 [✓] Disconnected from Red Hat Subscription Management
```

3.

The same goes for disconnection: with nothing connected in the first place, the output doesn't say yggdrasil wasn't enabled in the first place. That's a non-blocking note, but it feels strange.

4.

```
$ rhc connect --enable-feature remote-management --disable-feature remote-management
cannot enable feature: "remote-management": feature "remote-management" explicitly disabled
```

Are we able to display something prettier and more human-understandable? Non-blocking note, may be addressed later.

m-horky:
> Or we can introduce something like this. This has potential to add there more attributes in the future:
> 
> ```json
> {
>     "hostname": "localhost",
>     "uid": 0,
>     "features": {
>         "content": {
>             "enabled": false,
>         },
>         "analytics": {
>             "enabled": true,
>         },
>         "remote-management": {
>             "enabled": true,
>         },
>     },
>     "rhsm_connected": true,
>     "content_enabled": false,
>     "insights_connected": true,
>     "yggdrasil_started": true
> }
> ```

I like the nested approach. Would it make sense to move the `_connected`, `_enabled` keys into the `features` map?

```json
"features": {
  "content": {
    "enabled": true,
    "connected": true,
  },
  "analytics": {
    "enabled": true,
    "connected": false,
  },
  ...
}
```

rhsm.go
```diff
@@ -132,6 +133,13 @@ func registerUsernamePassword(username, password, organization, serverURL string
 		return orgs, err
 	}
 
+	var enabledContentStr string
+	if enableContent {
+		enabledContentStr = "true"
+	} else {
+		enabledContentStr = "false"
+	}
```
subpop, line 141: I believe this could be simplified into a single `Sprintf` statement, using the `Stringer` implementation of the `bool` type.
```suggestion
	enableContentStr := fmt.Sprintf("%v", enableContent)
```

subpop (changes requested):
One small change I think we can make. I will test the behavior of this branch and follow up with another review.

rhsm.go
```diff
@@ -229,14 +237,21 @@ func registerActivationKey(orgID string, activationKeys []string, serverURL stri
...
+
 	if err := privConn.Object(
 		"com.redhat.RHSM1",
 		"/com/redhat/RHSM1/Register").Call(
 		"com.redhat.RHSM1.Register.RegisterWithActivationKeys",
 		dbus.Flags(0),
 		orgID,
 		activationKeys,
-		map[string]string{},
+		map[string]string{"enable_content": enabledContentStr},
```
subpop, line 254: When I try to register with my activation key, I get the following error:

```
[link@lima-rhel-10 rhc]$ sudo ./builddir/rhc connect -o 12671438 -a developer
Connecting lima-rhel-10 to Red Hat.

Features preferences: [✓]content, [✓]analytics, [✓]remote-management

 [𐄂] Cannot connect to Red Hat Subscription Management
  [𐄂] Skipping generation of Red Hat repository file
  [𐄂] Skipping connection to Red Hat Insights
  [𐄂] Skipping activation of yggdrasil service

TYPE   STEP  ERROR
ERROR  rhsm  cannot connect to Red Hat Subscription Management: error: Unknown arguments: dict_keys(['enable_content'])
```

Does the D-Bus method RegisterWithActivationKeys not accept 'enable_content' as
a parameter option? If so, does this mean that we cannot opt out of content if
registering with an activation key?

connect_cmd.go
```diff
@@ -127,77 +156,102 @@ func connectAction(ctx *cli.Context) error {
...
+				interactivePrintf(
+					"  [%v] Skipping connection to Red Hat Insights\n",
+					uiSettings.iconError,
+				)
 			}
 		} else {
-			connectResult.InsightsConnected = true
-			interactivePrintf("%v Connected to Red Hat Insights\n", uiSettings.iconOK)
+			start = time.Now()
+			err = showProgress(" Connecting to Red Hat Insights...", registerInsights)
```
subpop, line 198: With the introduction of the space-indented results printout in the interactive
mode, the in-progress spinner looks awkward.

```
 / Connecting to Red Hat Insights...
```

Can we put that spinner inside brackets too? That can reduce the bouncing back
and forth as new lines are printed.

```
  [/] Connecting to Red Hat Insights...
```

connect_cmd.go
```diff
@@ -127,77 +156,102 @@ func connectAction(ctx *cli.Context) error {
...
 			}
-		} else {
-			connectResult.YggdrasilStarted = true
-			interactivePrintf("%v Activated the %v service\n", uiSettings.iconOK, ServiceName)
+			durations[ServiceName] = time.Since(start)
 		}
-		durations[ServiceName] = time.Since(start)
-		interactivePrintf("\nSuccessfully connected to Red Hat!\n")
+	} else {
+		interactivePrintf(" [ ] Management .... Starting %s service disabled\n", ServiceName)
```
subpop, line 250: When I disabled the 'remote-management' feature, it looks like the space-indents
were missing one:

```
  [✓] Analytics ... Connected to Red Hat Insights
 [ ] Management .... Starting yggdrasil service disabled
```

subpop (changes requested):

rhsm.go
jirihnidek: @subpop Good point. I am working on the fix fix for `rhsm.service`. It should be simple.

connect_cmd.go
jirihnidek: I would like to have some white space between checkbox and the text too, but standard "space" is IMHO too long and final line could be confusing at the first glance:

Current behavior:

```
Features preferences: [✓]content, [ ]analytics, [ ]remote-management
```

Proposed behavior:

```
Features preferences: [✓] content, [ ] analytics, [ ] remote-management
```

connect_cmd.go
jirihnidek: Same comment here.

connect_cmd.go
m-horky: That looks fine to me, I wouldn't worry too much.

connect_cmd.go
jirihnidek: Yes, it is doable.

connect_cmd.go
jirihnidek: Good catch. Matyas also reported this issue. :+1:

jirihnidek:
@m-horky @subpop OK, it seems we have consensus that we will use the third variant of proposed JSON format.

BTW: We can do big changes of JSON format ATM, because it hasn't been shipped in any official release. We can drop `content_enabled`, `insights_connected` and `yggdrasil_started` and move all information to `features`, because `insights_connected` would be duplicate of `features["analytics"].connected`. I'm also proposing to not use `connected` keyword, but I would use word `successful`, because word `connected` does not make much sense e.g. for feature `content`

```json
{
    "hostname": "localhost",
    "uid": 0,
    "rhsm_connected": true,
    "features": {
        "content": {
            "enabled": false,
            "successful": true,
        },
        "analytics": {
            "enabled": true,
            "successful": false,
        },
        "remote-management": {
            "enabled": true,
            "successful": true,
        },
    },
}
```

Archana-PandeyM:
Hi, this is my first attempt to get hands on for this feature. When I am trying to enable only one feature I still see everything getting enabled, for example  I am trying to enable only content but all three feature are enabled. This should happen, I should not need to mention disable-feature explicitly- 
```
`[root@kvm-01-guest13 ~]# rhc connect --username ******* --password ******* --enable-feature content
Connecting kvm-01-guest13.lab.eng.brq2.redhat.com to Red Hat.

Features preferences: [✓]content, [✓]analytics, [✓]remote-management

  [✓] Analytics ... Connected to Red Hat Insights
  [✓] Remote Management ... Activated the yggdrasil service

[root@kvm-01-guest13 ~]#`
```

subpop:
> @subpop @jirihnidek I have one more concern, when rhc connect is run without --enable-feature or --disable-feature then by default connectivity level should be “level 2: analytics” (Did this requirement change overtime, not sure so I am just highlighting )

rhc today enables all three levels by default, so the design here maintains that. I'll check with Christian if he has any expectation around changing the default enabled features.

Archana-PandeyM:
@subpop from my point of view I see a major confusion in the way this is implemented, If everything is enabled by default, then what is use of --enable-feature? for example 

- if user wants to enable only content and analytics with current implementation then only using --disable-feature remote-management would be sufficient.
- if user wants to enable only content then --disable-feature analytics should be sufficient because without analytics no use of remote-management 
Once the user is connected --enable-feature --disable-feature cannot be run (for that specific purpose we will have 'configure' in future(probably) as suggested on the design doc).

I could be missing something here but still what is the use of --enable-feature if everything is enabled by default and these options cannot be used once registered.

Archana-PandeyM:
With current implementation if  'Analytics' is disabled then 'remote-management' should be disabled too.  but looks like there is issue-:
```
[root@kvm-01-guest13 ~]# rhc connect --username ********* --password *********  --disable-feature analytics
Connecting kvm-01-guest13.lab.eng.brq2.redhat.com to Red Hat.

Features preferences: [✓]content, [ ]analytics, [✓]remote-management

  [ ] Analytics ... Connecting to Red Hat Insights disabled
  [✓] Remote Management ... Activated the yggdrasil service

```

In the command output above 'analytics' is disabled but 'remote-management' is enabled. We should throw error instead or  disable both.

jirihnidek:
@Archana-PandeyM you are right. I forgot to cover this case. I altered behavior a little bit, when `analytics` is disabled, then "remote-management" is also disabled. I also added reason to the output, why the feature was disabled:

```
root@localhost:/# rhc connect --username admin --password admin --organization donaldduck \
    --disable-feature analytics
Connecting thinkpad-p1 to Red Hat.

Features preferences: [✓]content, [ ]analytics, [ ]remote-management

  [ ] Analytics ... Connecting to Red Hat Insights disabled
  [ ] Management .... Starting yggdrasil service disabled (required feature "analytics" is disabled)

```

constants.go
```diff
@@ -92,3 +92,6 @@ func init() {
 		ServiceName = "yggdrasil"
 	}
 }
+
+const smallIndent = " "
+const mediumIndent = "  "
```
subpop, line 97: I kinda feel like these belong in `interactive.go`. They primarily exist to
support the interactive output.

subpop (changes requested):
One tiny consideration about the location of `smallIndent` and `mediumIndent`. Otherwise, this looks good to me from a code perspective.

I tested the behavior too and it works as expected with username and password. Activation keys still don't work due to https://github.com/candlepin/subscription-manager/pull/3488.

Archana-PandeyM:
@jirihnidek  One last observation for this round of testing, I noticed that even though I requested to disable 'content' and  the rhc connect says the file was not generated, Still redhat.repo file has repository details. Is the change in subman build(Do I need to install a seperate build)? Here are the reproducer steps:
```
`[root@kvm-01-guest13 ~]# ll /etc/yum.repos.d/redhat.repo
ls: cannot access '/etc/yum.repos.d/redhat.repo': No such file or directory
[root@kvm-01-guest13 ~]# 
[root@kvm-01-guest13 ~]# rhc connect --username *********** --password **********  --disable-feature content
Connecting kvm-01-guest13.lab.eng.brq2.redhat.com to Red Hat.

Features preferences: [ ]content, [✓]analytics, [ ]remote-management

  [ ] Content ... Red Hat repository file not generated                       <<<<< file not generated
  [✓] Analytics ... Connected to Red Hat Insights
  [ ] Management .... Starting yggdrasil service disabled (required feature "content" is disabled)

[root@kvm-01-guest13 ~]# ll /etc/yum.repos.d/redhat.repo
-rw-r--r--. 1 root root 48261 Jan 15 17:02 /etc/yum.repos.d/redhat.repo          <<<< file is still generated and content enabled
[root@kvm-01-guest13 ~]# yum repolist | grep rhel
rhel-10-for-x86_64-appstream-rpms                                         Red Hat Enterprise Linux 10 for x86_64 - AppStream (RPMs)
rhel-10-for-x86_64-baseos-rpms                                            Red Hat Enterprise Linux 10 for x86_64 - BaseOS (RPMs)
[root@kvm-01-guest13 ~]# 

`
```

jirihnidek:
@Archana-PandeyM yeah, you also need to use build of subscription-manager from this PR: https://github.com/candlepin/subscription-manager/pull/3488 (do not forget to restart rhsm.service) Otherwise, the content is not disabled due to bug in subscription-manager (nonempty value including `"false"` is represented as a true).

constants.go
jirihnidek: @subpop You are right. I updated the PR.

Archana-PandeyM:
@jirihnidek Thanks!
I tried with the mentioned rhel build and could reproduce the issue.
```

[root@kvm-01-guest13 ~]# rpm -qa | grep subscription-manager
subscription-manager-rhsm-certificates-20220623-6.el10.noarch
python3-subscription-manager-rhsm-1.30.3-1.20250106164229731439.pr3488.5.gcb97911fd.el10.x86_64
subscription-manager-1.30.3-1.20250106164229731439.pr3488.5.gcb97911fd.el10.x86_64
libdnf-plugin-subscription-manager-1.30.3-1.20250106164229731439.pr3488.5.gcb97911fd.el10.x86_64
[root@kvm-01-guest13 ~]# 
[root@kvm-01-guest13 ~]# 
[root@kvm-01-guest13 ~]# 
[root@kvm-01-guest13 ~]# systemctl restart rhsm.service
[root@kvm-01-guest13 ~]# rhc connect --username ********** --password ************* --disable-feature content
Connecting ****************** to Red Hat.

Features preferences: [ ]content, [✓]analytics, [ ]remote-management

  [ ] Content ... Red Hat repository file not generated
  [✓] Analytics ... Connected to Red Hat Insights
  [ ] Management .... Starting yggdrasil service disabled (required feature "content" is disabled)

[root@kvm-01-guest13 ~]# 

[root@kvm-01-guest13 ~]# ll /etc/yum.repos.d/redhat.repo 
-rw-r--r--. 1 root root 48261 Jan 21 08:31 /etc/yum.repos.d/redhat.repo
[root@kvm-01-guest13 ~]# yum repolist | grep rhel
rhel-10-for-x86_64-appstream-rpms                                         Red Hat Enterprise Linux 10 for x86_64 - AppStream (RPMs)
rhel-10-for-x86_64-baseos-rpms                                            Red Hat Enterprise Linux 10 for x86_64 - BaseOS (RPMs)
[root@kvm-01-guest13 ~]#
```

jirihnidek:
OK, I have result of investigation. It seems that `insights-client --register` calls `subscription-manager refresh`. Thus, content is enabled despite rhsm.service itself does not install any content.

When D-Bus methods `Register*()` are called with option `enable_content: false`, then the content is installed later by `rhsmcertd.service` during periodical certificate check-in despite `/etc/pki/entitlement` is empty.

If we really want to disable content, then we probably have to disable managing content in `/etc/rhsm/rhsm.conf` (`manage_repos = 1`).

jirihnidek:
User can run `rhc connect` multiple times with different set of levels of connectivity. It seems that it can enable/disable some feature, but it has no effect as you can see here:

```
[root@localhost ~]# rhc connect --username ********** --password ********** --disable-feature content
Connecting localhost to Red Hat.

Features preferences: [ ]content, [✓]analytics, [ ]remote-management

  [ ] Content ... Red Hat repository file not generated
  [✓] Analytics ... Connected to Red Hat Insights
  [ ] Management .... Starting yggdrasil service disabled (required feature "content" is disabled)

[root@localhost ~]# rhc status

✓ Connected to Red Hat Subscription Management
✓ Connected to Red Hat Insights
● The yggdrasil service is inactive

[root@localhost ~]# rhc connect --username ********** --password ********** --disable-feature analytics
Connecting localhost to Red Hat.

Features preferences: [✓]content, [ ]analytics, [ ]remote-management

 [✓] This system is already connected to Red Hat Subscription Management
  [ ] Analytics ... Connecting to Red Hat Insights disabled
  [ ] Management .... Starting yggdrasil service disabled (required feature "analytics" is disabled)

[root@localhost ~]# rhc status

✓ Connected to Red Hat Subscription Management
✓ Connected to Red Hat Insights
● The yggdrasil service is inactive

```

I think that `rhc connect` should return some error on already connected system and non-zero exit code.

jirihnidek:
The `rhc status` should also print some information about enabled/disabled content, but it should be part of another PR.

subpop:
> The `rhc status` should also print some information about enabled/disabled content, but it should be part of another PR.

Agreed.

subpop:
> User can run `rhc connect` multiple times with different set of levels of connectivity. It seems that it can enable/disable some feature, but it has no effect as you can see here:
> 
> ```
> [root@localhost ~]# rhc connect --username ********** --password ********** --disable-feature content
> Connecting localhost to Red Hat.
> This might take a few seconds.
> 
> Features preferences: [ ]content, [✓]analytics, [ ]remote-management
> 
>  [✓] Connected to Red Hat Subscription Management
>   [ ] Content ... Red Hat repository file not generated
>   [✓] Analytics ... Connected to Red Hat Insights
>   [ ] Management .... Starting yggdrasil service disabled (required feature "content" is disabled)
> 
> Successfully connected to Red Hat!
> 
> Manage your connected systems: https://red.ht/connector
> 
> 
> [root@localhost ~]# rhc status
> Connection status for localhost:
> 
> ✓ Connected to Red Hat Subscription Management
> ✓ Connected to Red Hat Insights
> ● The yggdrasil service is inactive
> 
> Manage your connected systems: https://red.ht/connector
> 
> 
> [root@localhost ~]# rhc connect --username ********** --password ********** --disable-feature analytics
> Connecting localhost to Red Hat.
> This might take a few seconds.
> 
> Features preferences: [✓]content, [ ]analytics, [ ]remote-management
> 
>  [✓] This system is already connected to Red Hat Subscription Management
>   [✓] Content ... Red Hat repository file generated
>   [ ] Analytics ... Connecting to Red Hat Insights disabled
>   [ ] Management .... Starting yggdrasil service disabled (required feature "analytics" is disabled)
> 
> Successfully connected to Red Hat!
> 
> Manage your connected systems: https://red.ht/connector
> 
> 
> [root@localhost ~]# rhc status
> Connection status for localhost:
> 
> ✓ Connected to Red Hat Subscription Management
> ✓ Connected to Red Hat Insights
> ● The yggdrasil service is inactive
> 
> Manage your connected systems: https://red.ht/connector
> ```
> 
> I think that `rhc connect` should return some error on already connected system and non-zero exit code.

This makes sense. The first run of `rhc connect` enables insights-client. The second one doesn't disable it because `--disable-feature=analytics` doesn't actively disable a feature; it just skips activating a feature. Maybe we need to implement the two `EnableFunc` and `DisableFunc` handler functions for the features and make sure that they (a) act idempotently; and (b) are invoked first `DisableFunc` and then conditionally `EnableFunc` if the flag is requested.

The original design intent of `rhc connect` is to be idempotent. This way it can be run multiple times (as part of scripts and other tooling). I can see there being a design flaw in that design intent though; if you run `rhc connect` using one set of credentials, and then run it again using a different set of credentials, it might attempt to register with the newer credentials without first unregistering the first set of credentials. We should probably revisit this design.

jirihnidek:
@Archana-PandeyM @subpop the issue of not disabled content is fixed in the PR for subscription-manager: https://github.com/candlepin/subscription-manager/pull/3500

subpop:
> Shouldn't we also introduce `--force` CLI option to `connect` command?

No. Introducing a `--force` flag becomes ambiguous; what does it mean to "force a connection"? Will it register the system somehow anyway, if the credentials are invalid? "Force" does not clearly or explicitly communicate "first the system will be unregistered before attempting to be registered with the credentials".

If it is an error to run "register/connect" on a system that is already registered/connected, then the `connect` command should exit, reporting that error. The user should be expected to first unregistered the system by running `disconnect`. Then they follow up that command with `connect` again.

Now, that means that we probably *do* need to make changes to this PR so that `rhc connect` will exit with an error if the system is already registered. That will work around the issue of features previously enabled not being correctly disabled.

jirihnidek:
> Now, that means that we probably _do_ need to make changes to this PR so that `rhc connect` will exit with an error if the system is already registered. That will work around the issue of features previously enabled not being correctly disabled.

OK, added another commit returning error, when `rhc connect` is run on already connected system.

subpop (approved):
I think this is ready to go from a code perspective. I did basic functional testing as well (with the caveat that I installed subscription-manager from candlepin/subscription-manager#3500 in order to verify that rhc behaves correctly), but primarily focused on the code during my reviews.

jirihnidek:
@m-horky @subpop @Archana-PandeyM are we waiting on anything or can this PR be merged?

The PR: https://github.com/candlepin/subscription-manager/pull/3500 was merged to main branch.

Archana-PandeyM:
Thanks George! LGTM.

PR #183, opened 2025-01-14
fix: Temporary hide --content-template CLI option of connect

jirihnidek:
* It is not possible to use content templates in production ATM. For this reason, we decided to mark `--content-template` CLI option as hidden. Thus, it will be possible to test it, but it will not be exposed to users and customers.


jirihnidek:
Closing, because it seems that it is possible to test this functionality now.

PR #189, opened 2025-02-03
test: Skipping test test_rhc_worker_playbook_install_after_rhc_connect

Archana-PandeyM:
This test is failing due to unresolved issue CCT-696, After the issue is fixed the test might need to be rewritten also, so disabling the test unless issue is fixed.


jirihnidek (approved):
LGTM

PR #192, opened 2025-02-04
fix: Content template cannot be used, when content is disabled

jirihnidek:
* When content is disabled with --disable-feature content, then --content-template CLI option cannot be used, because it does not have any sense to use content-template environment with content, when content is disabled.

### DIFF
--- a/connect_cmd.go
+++ b/connect_cmd.go
@@ -76,6 +76,7 @@ func beforeConnectAction(ctx *cli.Context) error {
 	activationKeys := ctx.StringSlice("activation-key")
 	enabledFeatures := ctx.StringSlice("enable-feature")
 	disabledFeatures := ctx.StringSlice("disable-feature")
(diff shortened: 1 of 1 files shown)


connect_cmd.go
```diff
@@ -98,6 +99,10 @@ func beforeConnectAction(ctx *cli.Context) error {
 		return err
 	}
 
+	if ContentFeature.Enabled == false && len(contentTemplates) > 0 {
```
subpop, line 102: Looks like there is a lint suggestion here that should be resolved.

subpop (changes requested):

connect_cmd.go
jirihnidek: Yeah, I know, I will fix it, but to be honest, I do not like it ... it is so easy to overlook it ... See: `!Item`

subpop (approved):
Code looks good.

PR #196, opened 2025-02-10
test: Copying artifacts to tmt logs directory

Archana-PandeyM:


Lorquas (approved):
LGTM; this is the same stuff that is in insights-client's test.sh

PR #197, opened 2025-02-10
test: Marking two of the negative tests xfail

Archana-PandeyM:
These tests currently fail due to unresolved issue CCT-1155. To be enabled when issue is fixed.


jirihnidek (approved):
LGTM

PR #198, opened 2025-02-11
feat: Improved UX experience during 'rhc disconnect'

jirihnidek:
* Card ID: CCt-1162
* When some feature (e.g. analytics) was disabled during
  `rhc connect`, using `--disable-feature "analytics"` CLI option,
  then later do not try to unregister insights-client, when
  `rhc disconnect` is run. There is printed only informative
  message that given "feature" has been already disconnected.
* This change also improve the case, when system is connected
  using `rhc connect`, but later user run command
  `insights-client --unregister`. No error should be printed
  in such case.
* TODO: Fix integration tests
* When you run `rhc disconnect` multiple times, then output
  looks like this:

```
root@localhost:/# rhc disconnect
Disconnecting localhost from Red Hat.

 [●] The yggdrasil service is already inactive
 [●] Already disconnected from Red Hat Insights
 [●] Already disconnected from Red Hat Subscription Management

```

### DIFF
--- a/activate.go
+++ b/activate.go
@@ -3,7 +3,6 @@ package main
 import (
 	"context"
 	"fmt"
-
 	systemd "github.com/redhatinsights/rhc/internal/systemd"
 )
 
@@ -35,6 +34,25 @@ func activateService() error {
 	}
 
 	return nil
+}
+
+// isServiceInState returns true, when yggdrasil.service is in given state
+func isServiceInState(wantedState string) (bool, error) {
+	conn, err := systemd.NewConnectionContext(context.Background(), systemd.ConnectionTypeSystem)
+	if err != nil {
+		return false, fmt.Errorf("cannot connect to systemd: %v", err)
+	}
+	defer conn.Close()
+
+	state, err := conn.GetUnitState("yggdrasil.service")
+	if err != nil {
+		return false, fmt.Errorf("cannot get unit state: %v", err)
+	}
+	if state == wantedState {
+		return true, nil
+	} else {
+		return false, nil
+	}
 }
 
 // deactivateService tries to stop and disable the rhc-canonical-facts.timer,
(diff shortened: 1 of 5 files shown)


Archana-PandeyM:
@jirihnidek This change will break an integration tests test_disconnect_when_already_disconnected, earlier it was expected that on unregistering an already unregistered system will show show the legit error that insights-client can not be disconnected .

jirihnidek:
> @jirihnidek This change will break an integration tests `test_disconnect_when_already_disconnected`, earlier it was expected that on unregistering an already unregistered system will show show the legit error that insights-client can not be disconnected .

Hmm, it seems that we should change behavior of `rhc disconnect` little bit and rephrase these messages not only due to levels of connectivity, but also due to to fact that user can interact with `insights-client` and `yggdrasil` service independently on `rhc`. And then we should also change expectation of `test_disconnect_when_already_disconnected` test. I will explain it more.

When user connects RHEL system using `rhc connect`, then later user can run `insights-client --unregister`, `systemct stop yggdrasil` or `subscription-manager unregister`. Thus, `rhc disconnect` should only try to disconnect system. When something is not possible to stop or unregister, then it should not be considered as a error, but legit state and print e.g. `insights-client already unregistered`.

jvlcek (approved):
As hard as I've tried I really can not find anything wrong with this code change. ;)

I tried running the code with the -disable-feature followed by disconnect and multiple consecutive disconnects. It seems to work correctly to me.

LGTM

I approve the PR. Let's get it merged.

Archana-PandeyM:
@jirihnidek Thanks for fixing the messages, it looks good to me now. 
We should merge this PR and update the related test in another PR (test is currently skipped).

Archana-PandeyM (approved):
LGTM!

PR #199, opened 2025-02-11
fix: Allow to use activation key(s) and org with --format json

jirihnidek:
* Card ID: CCT-1161
* When `--format json` is used, then all credentials has to be provided. Not only `username` & `password` has to be considered, but also `activation-key` and `organization` has to be allowed to use in such case.
* Added one more check: combination of activation-key and password should not be allowed.

### DIFF
--- a/integration-tests/test_connect.py
+++ b/integration-tests/test_connect.py
@@ -20,29 +20,63 @@ from utils import (
 logger = logging.getLogger(__name__)
 
 
-@pytest.mark.parametrize("auth", ["basic", "activation-key"])
-def test_connect(external_candlepin, rhc, test_config, auth):
-    """Test if RHC can connect to CRC using basic auth and activation key,
-    Also verify that yggdrasil/rhcd service is in active state afterward.
+@pytest.mark.parametrize(
+    "auth, output_format",
+    [
+        ("basic", None),
+        ("basic", "json"),
+        ("activation-key", None),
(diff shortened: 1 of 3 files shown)


integration-tests/test_connect.py
```diff
@@ -20,29 +20,60 @@
...
+        ("basic", None),
+        ("basic", "json"),
+        ("activation-key", None),
+        ("activation-key", "json"),
+    ]
+)
+def test_connect(external_candlepin, rhc, test_config, auth, output_format):
     """
+     Test if RHC can connect to CRC using basic auth and activation key,
+     Also verify that yggdrasil/rhcd service is in active state afterward.
```
Archana-PandeyM, line 35: @jirihnidek Thanks for the test, LGTM to me. Just a small request/suggestion can we add a mention of machine and non machine readable format here in description of test.

Archana-PandeyM (changes requested):

integration-tests/test_connect.py
jirihnidek: @Archana-PandeyM Thanks for review. This and other comments were updated.

subpop (approved):
Code review looks good to me.

jvlcek (approved):
Thank you @jirihnidek These code changes look good to me.

Archana-PandeyM:
LGTM!

PR #203, opened 2025-02-13
test: Updating name of yggdrasil service to rhcd in test for rhel9.

Archana-PandeyM:
Name of yggdrasil daemon is rhcd in rhel 9, added check to autodetect the name of service and perform assertions as requested


jirihnidek (approved):
LGTM

PR #206, opened 2025-02-17
test: Check for bootc/image-mode deployments

Aftermath:
DNF commands cannot be run without special options, and our images will be pre-prepared for use in tests.

Card: CCT-1183


Lorquas (approved):
LGTM

PR #209, opened 2025-02-20
feat(ci): more tests upstream

Lorquas:
TODO: WRITE DESCRIPTION


.packit.yml
```diff
@@ -90,9 +90,9 @@ jobs:
     trigger: pull_request
     identifier: "unit/rhel"
     targets:
-      rhel-9-x86_64:
+      centos-stream-10-x86_64: # TODO: Change me back to RHEL
```
subpop, line 93: Does this need to be changed to RHEL before we merge?

subpop (changes requested):
This looks good to me. I have one question about this TODO comment.

.packit.yml
Lorquas: This needs to be changed to RHEL whenever copr adds rhel-10 bulding

subpop: Ok. We'll leave it on centos-stream-10 then. Thanks!

subpop (approved):

Issue #45, opened 2023-09-14
Error returned by function calls is not explicitly checked


subpop:
Running the package through a static analyzer results in the following findings:

```
1. rhc/rhc-0.2.4/util.go:129: suppressed_error: The error returned by the function "Parse("redhat_access/r/insights/platform/config-manager/v2/profiles/current")" is not explicitly checked.
2. rhc/rhc-0.2.4/util.go:129: remediation: Explicitly check the return error.
#   127|   				return "", fmt.Errorf("cannot get base URL: %w", err)
#   128|   			}
#   129|-> 			p, _ := url.Parse("redhat_access/r/insights/platform/config-manager/v2/profiles/current")
#   130|   			uString = base.ResolveReference(p).String()
#   131|   		}
```

```
1. rhc/rhc-0.2.4/util.go:120: suppressed_error: The error returned by the function "Parse("api/config-manager/v2/profiles/current")" is not explicitly checked.
2. rhc/rhc-0.2.4/util.go:120: remediation: Explicitly check the return error.
#   118|   		if strings.Contains(serverHost, "subscription.rhsm.redhat.com") {
#   119|   			baseURL, _ = url.Parse("https://cert.console.redhat.com")
#   120|-> 			p, _ := url.Parse("api/config-manager/v2/profiles/current")
#   121|   			uString = baseURL.ResolveReference(p).String()
#   122|   		} else {
```

```
1. rhc/rhc-0.2.4/util.go:119: suppressed_error: The error returned by the function "Parse("https://cert.console.redhat.com")" is not explicitly checked.
2. rhc/rhc-0.2.4/util.go:119: remediation: Explicitly check the return error.
#   117|   		// Check if it is the default api server
#   118|   		if strings.Contains(serverHost, "subscription.rhsm.redhat.com") {
#   119|-> 			baseURL, _ = url.Parse("https://cert.console.redhat.com")
#   120|   			p, _ := url.Parse("api/config-manager/v2/profiles/current")
#   121|   			uString = baseURL.ResolveReference(p).String()
```

```
1. rhc/rhc-0.2.4/util.go:107: suppressed_error: The error returned by the function "Parse("api/config-manager/v2/profiles/current")" is not explicitly checked.
2. rhc/rhc-0.2.4/util.go:107: remediation: Explicitly check the return error.
#   105|   			return "", fmt.Errorf("cannot get base URL: %w", err)
#   106|   		}
[...]

PR #213, opened 2025-02-25
Fix error checking documented in Issue #45

cnsnyder:
This PR adds checks for the errors potentially returned in a number of places as documented here : https://github.com/RedHatInsights/rhc/issues/45

This code is not as DRY as I'd like it, all feedback welcome! I hope to solve issues and improve!

Cheers!

### DIFF
--- a/main.go
+++ b/main.go
@@ -83,11 +83,20 @@ func showTimeDuration(durations map[string]time.Duration) {
 	if log.CurrentLevel() >= log.LevelDebug {
 		fmt.Println()
 		w := tabwriter.NewWriter(os.Stdout, 0, 0, 2, ' ', 0)
-		_, _ = fmt.Fprintln(w, "STEP\tDURATION\t")
-		for step, duration := range durations {
-			_, _ = fmt.Fprintf(w, "%v\t%v\t\n", step, duration.Truncate(time.Millisecond))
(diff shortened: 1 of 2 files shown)


Lorquas:
Can one of the admins verify this patch?

subpop (approved):
LGTM

subpop:
> This code is not as DRY as I'd like it, all feedback welcome! I hope to solve issues and improve!

I think there's a follow-up commit or PR we can introduce here that changes the behavior of the `showErrorMessages` function to return a normal error, instead of a bunch of empty `cli.Exit()` errors. Then we update the callers of `showErrorMessage` to check for the returned error and wrap it in a `cli.Exit`.

DuckBoss (approved):
LGTM, thanks for the changes.

I had the same thoughts as Link when I saw the `showErrorMessages` method using `cli.Exit()`, and I would prefer the logic changes to be in a separate small PR instead of this one.

subpop (changes requested):
The commit linter is failing. 650578bca763a5b1dd799145bd663b9285a6c731 contains a capitalized "Fix". It should be lower-case to conform to Conventional Commits. Do you mind amending the commit and changing the message?

m-horky:
Hey, this pull request has been pending here for a long time. I'm closing it for now, we'll address the issue later if it comes up again.

PR #228, opened 2025-04-23
fix: Return non-zero code, when wrong cli options are used.

jirihnidek:
* Card ID: CCT-1329
* The `rhc` returned 0 exit code in cases, when e.g. not allowed
  combinations of CLI options were used.
* When e.g. wrong combination of CLI option is used and it is
  detected in `beforeConnectAction()` function then this function
  has to return `cli.Exit()` error with appropriate exit code.

## Summary by Sourcery

Improve error handling and exit codes for the RHC CLI tool to provide more accurate and informative error responses when incorrect command-line options are used.

Bug Fixes:
- Ensure the CLI returns non-zero exit codes when invalid or conflicting command-line options are provided
- Improve error handling to return appropriate exit codes for different error scenarios

Enhancements:
- Refactor error handling in beforeConnectAction to use cli.Exit() for more precise error reporting
- Add more specific error handling for various CLI option validation scenarios

Chores:
- Update error handling to log connection and file closing errors
- Improve error message clarity for CLI option validation

### DIFF
--- a/connect_cmd.go
+++ b/connect_cmd.go
@@ -52,8 +52,11 @@ func (connectResult ConnectResult) Error() string {
 	return result
 }
 
-// beforeConnectAction ensures that user has supplied a correct CLI options
-// and there is no conflict between provided options
+// beforeConnectAction ensures that user has supplied correct CLI options
+// and there is no conflict between them. When there is anything wrong,
+// then this function will invoke cli.Exit() with an appropriate error
+// message and error code. The exit codes are defined in the
+// constants.go module
 func beforeConnectAction(ctx *cli.Context) error {
 	// First check if machine-readable format is used
 	err := setupFormatOption(ctx)
@@ -64,10 +67,13 @@ func beforeConnectAction(ctx *cli.Context) error {
 	// When machine is already connected, then return error
 	uuid, err := getConsumerUUID()
 	if err != nil {
-		return fmt.Errorf("unable to get consumer UUID: %s", err)
+		return cli.Exit(
+			fmt.Sprintf("unable to get consumer UUID: %s", err),
+			ExitCodeSoftware,
+		)
 	}
 	if uuid != "" {
-		return fmt.Errorf("this system is already connected")
+		return cli.Exit("this system is already connected", ExitCodeUsage)
 	}
 
 	username := ctx.String("username")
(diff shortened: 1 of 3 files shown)


jirihnidek:
@sourcery-ai review

jirihnidek:
I will add integration tests to this PR, but adding such tests is blocked by https://github.com/RedHatInsights/rhc/pull/230

connect_cmd.go
```diff
@@ -52,8 +52,10 @@ func (connectResult ConnectResult) Error() string {
 	return result
 }
 
-// beforeConnectAction ensures that user has supplied a correct CLI options
-// and there is no conflict between provided options
+// beforeConnectAction ensures that user has supplied correct CLI options
+// and there is no conflict between provided options. When there is anything
+// wrong, then this function has to return cli.Exit() error. The exit codes
+// are defined in constants.go module
```
jvlcek, line 58: I just have some word-smithing suggestions:

Please consider the following updates:

// beforeConnectAction ensures that **the** user has supplied **the** correct CLI options
// and there is no conflict between ~~provided options~~ **them**. When there is anything
// wrong this function ~~has to return~~ **will invoke** cli.Exit() **with an appropriate** error **message and
// error code**.
// The exit codes are defined in **the** constants.go module

connect_cmd.go
```diff
@@ -64,10 +66,13 @@ func beforeConnectAction(ctx *cli.Context) error {
 	// When machine is already connected, then return error
 	uuid, err := getConsumerUUID()
 	if err != nil {
-		return fmt.Errorf("unable to get consumer UUID: %s", err)
+		return cli.Exit(
+			fmt.Sprintf("unable to get consumer UUID: %s", err),
```
jvlcek, line 70: I just noticed some inconsistency for when err and err.Error() are used
as part of the arguments to cli.Exit()

In some places, like here, err is used and in others err.Error() is used.

Should err.Error() always be use?

Please confirm they are being used correctly?

jvlcek (changes requested):
@jirihnidek Thank you for this PR. It looks good. I just have one word-smithing suggestions and would like to ask you to confirm the seemingly inconsistent usage of err and err.Error() is correctly being passed as the argument to the new cli.Exit() calls.

connect_cmd.go
jirihnidek: Using both variants (`fmt.Sprintf("formater %s", err)` or `fmt.Sprintf("formater %s", err.Error())`) is correct. The `error` is interface type, which has `Error()` method returning string.

The documentation https://pkg.go.dev/fmt says: "If an operand implements the error interface, the Error method will be invoked to convert the object to a string, which will then be formatted as required by the verb (if any)."

jvlcek: Thank you for confirming @jirihnidek

connect_cmd.go
jirihnidek: Fixed according your suggestions. Thanks.

jirihnidek:
@Archana-PandeyM extended a little bit integration tests to use `exit_code` in few cases.

PR #230, opened 2025-04-25
feat: Remove hidden CLI option --server from connect command

jirihnidek:
* Card ID: CCT-1335
* This CLI option could never been used in production, because configuring satellite server requires more actions. Thus, this CLI option was useless. The connect command should not have any unexpected side effect like changing of configuration file.
* Do not use `--server` in integration tests, because this CLI option was removed from connect command

### DIFF
--- a/integration-tests/test_connect.py
+++ b/integration-tests/test_connect.py
@@ -80,36 +80,34 @@ def test_connect(external_candlepin, rhc, test_config, auth, output_format):
 
 
 @pytest.mark.parametrize(
-    "credentials,server",
+    "credentials,return_code",
     [
-        (  # username and password: valid, server: invalid
-            {"username": "candlepin.username", "password": "candlepin.password"},
-            "http://non-existent.server",
-        ),
-        (  # organization and activation-key: valid, server: invalid
+        (  # username: invalid, password: valid
             {
-                "organization": "candlepin.org",
-                "activation-key": "candlepin.activation_keys",
+                "username": "non-existent-user",
+                "password": "candlepin.password"
             },
-            "http://non-existent.server",
-        ),
-        (  # password and server: valid, username: invalid
-            {"username": "non-existent-user", "password": "candlepin.password"},
             None,
         ),
-        (  # activation-key and server: valid, organization: invalid
+        (  # username: valid, password: invalid
+            {
+                "username": "candlepin.username",
+                "password": "xpto123"
+            },
+            None,
+        ),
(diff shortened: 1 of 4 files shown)


jirihnidek:
Failed RHEL-10 build seems false positive. The build finished correctly. All required rpm files were built.

main.go
```diff
@@ -227,6 +222,13 @@ func main() {
 					Usage:   "prints status in machine-readable format (supported formats: \"json\")",
 					Aliases: []string{"f"},
 				},
+				&cli.StringFlag{
+					Name:        "schema",
+					Usage:       "Print schema of machine-readable format (supported formats: \"json\")",
+					Aliases:     []string{"s"},
+					Required:    false,
+					DefaultText: "",
+				},
```
jvlcek, line 231: Should this be a separate PR? Maybe I'm missing something but I don't get the connection between removing the hidden CLI option --server and adding this status subcommand schema.

jvlcek (approved):
@jirihnidek Thank you for this PR. It seems OK to me. I just have one question about adding the schema status subcommand. Should that be a separate PR, or at least a separate commit,  or am I missing the connection?

integration-tests/test_connect.py
```diff
@@ -80,36 +80,34 @@ def test_connect(external_candlepin, rhc, test_config, auth, output_format):
 
 
 @pytest.mark.parametrize(
-    "credentials,server",
+    "credentials,return_code",
```
Archana-PandeyM, line 83: Thanks for the PR George! I have only one suggestion to only remove --server from parametrized argument list and not introduce return_code. These are all negative test cases and the return code should ideally non-zero like we are trying to implement here https://github.com/RedHatInsights/rhc/pull/228

Archana-PandeyM (changes requested):

integration-tests/test_connect.py
jirihnidek: I already tried to do it in the way you are suggesting, but it required too big changes in code due to Python syntax. Python has a problem with list of tuples containing only item in this case. It would require reformatting all uses cases.

I considered it easier to replace `server` with `return_code` (not used ATM), which we will want to add later anyway. It will be possible to use `return_code` after merging of #228.

main.go
jirihnidek: It should be fixed now.

jvlcek (approved):
LGTM 
Thank you @jirihnidek

PR #235, opened 2025-05-12
feat: Added info about content to status command

jirihnidek:
* Card ID: CCT-1348
* When content is enabled/disabled during `rhc connect` using
  `--enabled-feature content` or `--disable-feature content`,
  then this information was not displayed in `rhc status`
  command
* This commit add this information to the output of
  `rhc status` and `rhc status --format json`
* The information is read from rhsm.conf using RHSM D-Bus API
* Content is considered as enabled only in situation, when
  system is registered
* The output of `rhc status` is consistent with other commands
  like `rhc connect` and `rhc disconnect`.
* The output can look like this now:

```
root@localhost:/# rhc status

  [ ] Content is disabled
  [✓] Connected to Red Hat Insights
  [ ] The yggdrasil service is inactive
```


jirihnidek:
@Archana-PandeyM ^ Better?

jirihnidek:
@Archana-PandeyM extended unit tests to cover enabled/disabled content and extended integrations tests of "rhc status" in general.

Archana-PandeyM:
There is a failed test, I think we need to change the exist code check.

jirihnidek:
> There is a failed test, I think we need to change the exist code check.

Hi Archana, thanks. You are right. It was caused by copy-pasting code from another method. But it seems that other tests started to fail due to `insights-client`. It seems that something has changed in `insights-client` since the PR was created.

DuckBoss (approved):
LGTM, thanks :+1:

PR #237, opened 2025-06-03
feat(test): Added packit setup and integration tests for rhc 0.2

Archana-PandeyM:
This PR focuses on adding integration tests for rhc 0.2 version. Basically picking tests from main branch with suitable/required modifications.


integration-tests/test_connect.py
DuckBoss, line 1: Does `test_connect.py` also need to check for rhc-worker-playbook installation? I see that test case is missing between el10 and el9.

DuckBoss (approved):
Mostly LGTM, I approved it but I had a question noted below.

integration-tests/test_connect.py
Archana-PandeyM: The feature to install worker upon registration by config-manager has not been doing well. We are again evaluating options to implement it by redesigning. Also the test in main branch was disabled for same reason. There is an article published with RHEL 10 release to manually install worker after registration.

But I do plan to add a e2e test for rhc in another PR, future task.

PR #243, opened 2025-06-23
feat: improve canonical-facts error message

jlocash:
* Card ID: CCT-864
* Instead of simply propagating the original error, wrap it with a more informative message


DuckBoss (approved):
LGTM, thanks for the changes :+1: . I agree with you that it isn't really descriptive enough, but I think for the purposes of identifying the source of the missing file error it does its job.

jirihnidek (approved):
LGTM

PR #250, opened 2025-08-01
fix: The "rhc status" should handle errors properly

jirihnidek:
* Card ID: CCT-1526
* When there is some issue e.g. with rhsm.service, then the
  "rhc status" should not terminate immediately, but it
  should print error for given feature and continue with
  printing statuses of other feature
* When `--format json` is used, then errors should be filled
  for related features
* Refactored calling `insightStatus()` function to be consistent
  with other functions
* When it is not possible to access RHSM D-Bus API due
  to e.g. masked `rhsm.service`, then return error as is.
* When yggdrasil wasn't possible to start for various reasons,
  then `rhc status` did not print any error at all. The
  service was indicated as inactive despite the service was
  not present at all. When the service was masked, then
  it was also indicated as inactive
* When there is any problem with yggdrasil service, then
  appropriate message is printed. The message is gathered
  from D-Bus property
* Added integration tests of "rhc status" for the cases, when
  `rhsm.service` is masked

### DIFF
--- a/integration-tests/test_status.py
+++ b/integration-tests/test_status.py
@@ -16,6 +16,7 @@ More information about this module could be found: https://github.com/RedHatInsi
 
 import pytest
 import json
+import subprocess
 from pytest_client_tools import util
 from utils import yggdrasil_service_is_active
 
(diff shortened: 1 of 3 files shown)


jirihnidek:
Note: the `insights-client` is still broken rawhide, because it contains Python code that is not compatible with the latest Python on Fedora rawhide.

jirihnidek:
@Archana-PandeyM I FYI: I added integration test to this PR for this case

DuckBoss (approved):

jlocash (approved):
Thanks for these changes. I agree that printing each status message instead of hard exiting on the first is more intuitive

integration-tests/test_status.py
```diff
@@ -201,6 +202,248 @@ def test_status_disconnected(rhc):
...
+    # 'rhc disconnect' to ensure the system is already disconnected
+    rhc.run("disconnect", check=False)
+    # stop and mask rhsm.service
+    subprocess.run(["systemctl", "stop", "rhsm.service"])
+    subprocess.run(["systemctl", "mask", "rhsm.service"])
+    status_result = rhc.run("status", check=False)
+    assert status_result.returncode != 0
+    assert "Could not activate remote peer" in status_result.stdout
+    assert "Not connected to Red Hat Insights" in status_result.stdout
+    if pytest.service_name == "rhcd":
```
Archana-PandeyM, line 302: same here.

Archana-PandeyM:
LGTM!

jlocash:
Latest changes LGTM. I also ran a quick test by installing the RPM and running `rhc status` after masking `rhsm.service` and everything works as expected.

Dev: LGTM

PR #253, opened 2025-08-07
test: Add stability checking to rhcd_service_is_active()

Archana-PandeyM:
test_rhcd_service_restart was failing intermittently at line 109
Race condition where systemctl restart hadn't fully completed before service state was checked
Test was checking service state immediately after restart without waiting for stabilization.
Implemented automatic waiting before checking service state.


jirihnidek (approved):
LGTM

PR #254, opened 2025-08-07
test: improve systemd test coverage

jlocash:
* Card ID: CCT-1100
* Adds tests covering new connections
* Adds tests covering basic unit operations operations on valid, invalid, and non-existent units:

### DIFF
--- a/internal/systemd/systemd_test.go
+++ b/internal/systemd/systemd_test.go
@@ -5,80 +5,257 @@ import (
 	"os"
 	"path/filepath"
 	"testing"
-
-	"github.com/google/go-cmp/cmp"
-	"github.com/google/go-cmp/cmp/cmpopts"
 )
 
-func TestUnit(t *testing.T) {
+func TestNewConnectionContext(t *testing.T) {
 	if _, has := os.LookupEnv("DBUS_SESSION_BUS_ADDRESS"); !has {
 		t.Skip("DBUS_SESSION_BUS_ADDRESS undefined")
 	}
 
 	tests := []struct {
-		description string
-		input       struct {
-			unitFile string
-			now      bool
-		}
-		want      string
-		wantError error
+		name           string
+		connectionType ConnectionType
+		expectError    bool
 	}{
 		{
-			description: "activate simple.service",
-			input: struct {
-				unitFile string
-				now      bool
-			}{
-				unitFile: "testdata/simple.service",
-				now:      true,
-			},
-			want: "active",
+			name:           "user_connection",
+			connectionType: ConnectionTypeUser,
+			expectError:    false,
+		},
+		{
+			name:           "system_connection",
+			connectionType: ConnectionTypeSystem,
+			expectError:    false,
 		},
 	}
 
-	for _, test := range tests {
-		t.Run(test.description, func(t *testing.T) {
-			conn, err := NewConnectionContext(context.Background(), ConnectionTypeUser)
-			if err != nil {
-				t.Fatal(err)
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			conn, err := NewConnectionContext(context.Background(), tt.connectionType)
+
+			if tt.expectError {
+				if err == nil {
+					t.Error("expected error but got none")
+				}
+				return
 			}
 
-			abs, err := filepath.Abs(test.input.unitFile)
-			if err != nil {
-				t.Fatal(err)
+			if tt.connectionType == ConnectionTypeSystem && err != nil {
+				t.Skipf("System connection failed: %v", err)
+				return
 			}
 
-			// Link the testdata unit files into the search paths so they can be
-			// referenced by name. This simulates the context in which this
-			// package is more likely to be run.
-			_, err = conn.conn.LinkUnitFilesContext(conn.ctx, []string{abs}, true, false)
 			if err != nil {
-				t.Fatal(err)
+				t.Fatalf("unexpected error: %v", err)
 			}
 
-			err = conn.EnableUnit(filepath.Base(test.input.unitFile), test.input.now, true)
-			defer func(t *testing.T) {
-				if err := conn.DisableUnit(filepath.Base(test.input.unitFile), test.input.now, true); err != nil {
-					t.Fatal(err)
-				}
-			}(t)
-			if test.wantError != nil {
-				if !cmp.Equal(err, test.wantError, cmpopts.EquateErrors()) {
-					t.Errorf("%#v != %#v", err, test.wantError)
-				}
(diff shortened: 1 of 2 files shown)


internal/systemd/systemd_test.go
```diff
@@ -5,80 +5,253 @@ import (
...
+
+	t.Run("stop_nonexistent", func(t *testing.T) {
+		err := conn.StopUnit(unitName, false)
+		if err == nil {
+			t.Error("expected error when stopping non-existent unit")
+		}
+	})
+
+	t.Run("disable_nonexistent", func(t *testing.T) {
+		if err = conn.DisableUnit(unitName, false, true); err == nil {
```
jirihnidek, line 253: Please, use the same pattern for all checks.

internal/systemd/systemd_test.go
```diff
@@ -5,80 +5,253 @@ import (
...
+
+	unitName := "nonexistent.service"
+
+	t.Run("get_state_nonexistent", func(t *testing.T) {
+		state, err := conn.GetUnitState(unitName)
+		if err == nil {
+			t.Logf("GetUnitState for non-existent unit returned: %q", state)
+		} else {
+			t.Logf("GetUnitState failed for non-existent unit: %v", err)
+		}
```
jirihnidek, line 228: Shouldn't we call `t.Error()` in the case, when `err == nil`.

internal/systemd/systemd_test.go
```diff
@@ -5,80 +5,253 @@ import (
...
+func TestUnitOperationsInvalid(t *testing.T) {
+	if _, has := os.LookupEnv("DBUS_SESSION_BUS_ADDRESS"); !has {
+		t.Skip("DBUS_SESSION_BUS_ADDRESS undefined")
+	}
+
+	conn, err := NewConnectionContext(context.Background(), ConnectionTypeUser)
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer conn.Close()
```
jirihnidek, line 165: Please use closure for `conn.Close()`, because this function call can return `error`. It would be good to call `t.Errorf()` in such case or handle such situation somehow.

internal/systemd/systemd_test.go
```diff
@@ -5,80 +5,253 @@ import (
...
+func TestUnitOperationsNonExistent(t *testing.T) {
+	if _, has := os.LookupEnv("DBUS_SESSION_BUS_ADDRESS"); !has {
+		t.Skip("DBUS_SESSION_BUS_ADDRESS undefined")
+	}
+
+	conn, err := NewConnectionContext(context.Background(), ConnectionTypeUser)
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer conn.Close()
```
jirihnidek, line 218: Use closure here too.

internal/systemd/systemd_test.go
```diff
@@ -5,80 +5,253 @@ import (
...
+		}
+	})
+
+	t.Run("get_state_invalid_unit", func(t *testing.T) {
+		state, err := conn.GetUnitState(unitName)
+		if err != nil {
+			t.Logf("GetUnitState failed for invalid unit: %v", err)
+		} else {
+			t.Logf("GetUnitState returned: %q", state)
+		}
```
jirihnidek, line 205: Why the `t.Error()` is not called in one of the case?

internal/systemd/systemd_test.go
```diff
@@ -5,80 +5,253 @@ import (
...
+	}
+	defer conn.Close()
+
+	err = conn.Reload()
+	if err != nil {
+		t.Logf("Reload failed: %v", err)
+	}
+}
+
+func TestUnitOperationsValid(t *testing.T) {
```
jirihnidek, line 79: Please add here some comment saying that all functions calls should not return error in normal situation.

jirihnidek (commented):
Thanks for the PR. Overall it is OK. I have few comments and suggestions. In general please call `conn.Close()` in closure like this:

```go
defer func() {
	err = conn.Close()
	if err != nil {
		// Handle the error somehow.
		// Probably something like this `t.Errorf(err)`
	}
}()
```

You use different patterns for handling error. I think that we should use one pattern for all checks. Something like this:

```go
t.Run("start_invalid_unit", func(t *testing.T) {
    err := conn.StartUnit(unitName, false)
    if err == nil {
        t.Error("expected error when starting invalid unit")
    } else {
         t.Logf("expected error returned: %v", err)
    }
})
```

If it is OK that error is returned as well error is not returned from some function call, then it should be explained in the comment.

internal/systemd/systemd_test.go
jlocash: It looks like systemd will return an "inactive" ActiveState for units that don't exist:
```
# systemctl show does-not-exist.service --property=ActiveState
ActiveState=inactive
```
The code for this test is a little awkward so I will clean it up and add some comments explaining the behavior

internal/systemd/systemd_test.go
```diff
@@ -5,80 +5,257 @@ import (
 	"os"
 	"path/filepath"
 	"testing"
-
-	"github.com/google/go-cmp/cmp"
-	"github.com/google/go-cmp/cmp/cmpopts"
 )
 
-func TestUnit(t *testing.T) {
+func TestNewConnectionContext(t *testing.T) {
```
jirihnidek, line 10: Please write comment here explaining purpose of this unit test.

internal/systemd/systemd_test.go
jirihnidek: Oh, I didn't realize that this function does not return error at all. Strange.

internal/systemd/systemd_test.go
```diff
@@ -5,80 +5,257 @@ import (
...
+		}
+	})
+
+	t.Run("start_invalid_unit", func(t *testing.T) {
+		if err = conn.StartUnit(unitName, false); err == nil {
+			t.Error("expected error when starting invalid unit")
+		}
+	})
+
+	// systemd will return an ActiveState of "inactive" for invalid/non-existent units
```
jirihnidek, line 199: You duplicated this comment three times in this file and it does not make too much sense. What is `ActiveState`? I would rephrase it to something like this: `// systemd will return an "inactive" state for invalid/non-existent units`

jirihnidek (changes requested):
Thanks for your updates. I have few more comments and requests.

internal/systemd/systemd_test.go
jlocash: `ActiveState` is the systemd unit property name, but I will rephrase it like you suggested

m-horky (approved):
ACK

PR #259, opened 2025-08-19
fix: Fix docstrings in `test_status.py` so the new test cases can be uploaded to Polarion

pbartiko:
This PR fixes docstrings from newly added test cases in `test_status.py` (https://github.com/RedHatInsights/rhc/pull/250)


jirihnidek (approved):
We agreed that all rhcd strings will be removed in the future PR. LGTM

PR #262, opened 2025-08-25
chore: Move configuration to own package

m-horky:
* Card ID: CCT-1583

To improve code separation, move the configuration object to its own package.

Generated-By: Claude (claude-sonnet-4)


m-horky:
> Is there any design document how would we like to refactor rhc?
I'm not aware of any. Since we do not plan to have `rhc` be a library for others, putting it into `internal/` might make it crystal clear.

> Why did you introduce new directory `pkg`?
I've done it before I've opened the rest of the PRs, there's no real reason. I'll update this PR, moving it to internal as well.

PR #266, opened 2025-08-28
ci: enable s390x and ppc64le build targets

jlocash:
* Card ID: CCT-1407
* Added s390x and ppc64le targets to build jobs triggered by commits and PRs


jirihnidek (approved):
LGTM

PR #270, opened 2025-09-05
chore: Move interactive terminal detection to own package

m-horky:
* Card ID: CCT-1587

This patch removes big part of terminal and output-related logic into its own package. This will allow us to decouple CLI and business logic from each other later.

This patch deviates from the original implementation by separating the no-color preference from rich experience like interactive input and spinners.


m-horky:
Questions for the review to consider:
- I like the separation in `ui/` (where we have `struct.go` for structure definitions, with other methods using them). Do we want to make it a preferred code structure long-term?
- Are the `ui.IsInteractive`, `ui.IsOutputMachineReadable`, `ui.IsOutputRich` named in such a way it is clear what they represent?
- Are the `ui.Printf` and `ui.Spinner` method names descriptive enough?
- I've reordered the arguments for `printf`: first goes the method (as the most important), then the prefix, then suffix/message. Are the arguments against?
- You need to read the docs to configure output (see the arguments to `ui.ConfigureOutput`). Is there a prettier way?

pkoprda:
@m-horky please resolve conflicts

status_cmd.go
```diff
@@ -31,22 +32,22 @@ func rhsmStatus(systemStatus *SystemStatus) error {
 	}
 	if uuid == "" {
 		systemStatus.returnCode += 1
-		if uiSettings.isMachineReadable {
+		if ui.IsOutputMachineReadable() {
```
jlocash, line 35: I think a lot of these `if ui.IsOutputMachineReadable() {...} else {...}` are redundant, since `ui.Printf` will ignore the printed message if `IsOutputMachineReadable() == false`.

I think something like this would have the same effect with less code:
```go
...
	if uuid == "" {
		systemStatus.returnCode += 1
		systemStatus.RHSMConnected = false
		ui.Printf(
			"%s[ ] Not connected to Red Hat Subscription Management\n",
			ui.Indent.Small,
		)
	} else {
		systemStatus.RHSMConnected = true
		ui.Printf(
			"%s[%v] Connected to Red Hat Subscription Management\n",
			ui.Indent.Small,
			ui.Icons.Ok,
		)
	}
...
```

This is perhaps better suited for another PR, but just a thought

util.go
```diff
@@ -87,18 +87,13 @@ func checkForUnknownArgs(ctx *cli.Context) error {
...
 	format := ctx.String("format")
 	switch format {
 	case "":
 		return nil
 	case "json":
-		uiSettings.isMachineReadable = true
-		uiSettings.isRich = false
 		return nil
 	default:
 		err := fmt.Errorf(
```
jlocash, line 99: A simple `if` would suffice here, or if we expect to add other formats later, then it could be flattened to something like:
```go
switch format {
    case "", "json":
        return nil
    default:
        err := ...
}
```

internal/ui/print.go
jlocash, line 1: In general, I try to keep type declarations and methods as close together as possible, and split when a single file grows beyond a reasonable point. I'm not sure there's enough content in this package to warrant separating them into separate files. There's maybe 120 lines total in this package, so I think a single file would suffice.

jlocash (commented):
Thanks for the PR! Overall I think the code reads much better with your changes. I left a few minor comments. :smile:

status_cmd.go
m-horky: That's actually very reasonable. Thanks for the suggestion.

internal/ui/print.go
m-horky: This is definitely a matter of opinions. I grew to like to separate DTOs (the strucs used outside of the module) from methods that create, digest or manipulate them.

It doesn't matter much here, because almost everything is only referenced in the `ui.go` and nowhere else, but for bigger modules (that might implement i.e. foreign REST API) I found it quite useful.

I'll put it all into a single file for now, but I'd like us to reconsider later, if it grows into a pattern we'd enjoy working with.

jlocash (approved):
Looks like checks are passing now. LGTM :+1:

PR #278, opened 2025-09-16
test: Switch integration tests to use activation key authentication

Archana-PandeyM:
- Replaced username/password auth with activation key in tests. This ensures compatibility with satellite environments that don't support basic auth.


jirihnidek (approved):
LGTM

PR #282, opened 2025-09-17
chore(ci): Update GitHub testing matrix of golang

jirihnidek:
* Drop testing of golang 1.23 and start to test golang 1.25, because only two last minor versions of Go are supported. More details here: https://go.dev/doc/devel/release#policy
* When you install golang on RHEL-10.0, then 1.24.6 is installed. When you install golang on CentOS Stream 10, then 1.25.0 is installed


m-horky (approved):
LGTM

PR #284, opened 2025-09-26
chore: Move rhsm.go to own internal module

jirihnidek:
* Card ID: CCT-1590
* Moved the rhsm.go to internal/rhsm/rhsm.go
* Renamed some functions to be available as API
* No functional change


m-horky (changes requested):
ACK in general. One request.

PR #297, opened 2025-11-11
test: Add test cases for enable/disable features in 'rhc connect'

pbartiko:
Card ID: https://issues.redhat.com/browse/CCT-1091

Added test cases for various feature enable/disable combinations.
Added test cases for enable/disable non-existent feature.

The test cases use centralized feature definitions and auto-generated test case combinations. This improves maintainability and makes it easier to add new features in the future.

### DIFF
--- a/integration-tests/test_connect.py
+++ b/integration-tests/test_connect.py
@@ -12,6 +12,7 @@ import time
 import logging
 import pytest
 from datetime import datetime
+from itertools import combinations
 import sh
 import subprocess
 from pytest_client_tools.restclient import RestClient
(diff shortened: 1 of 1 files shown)


integration-tests/test_connect.py
```diff
@@ -527,3 +527,348 @@ def get_template_repos_by_name(template_name, username, password, template_url):
...
+            ["content", "analytics", "remote-management"],
+            {
+                "remote_management": False,
+                "content": False,
+                "analytics": False,
+            },
+            False,
+        ),
+    ],
+)
```
jirihnidek: This testing matrix is huge. If we will add another feature in the future, then there could be huge growth of the matrix (influenced by combination). If the approach from previous comment is used, then the testing could be also simplified and there will be no need to create such huge matrix. The testing matrix could be automatically generated inside testing function.

jirihnidek: If you consider this idea as too complicated to implement, then we can do refactoring later, when needed.

integration-tests/test_connect.py
```diff
@@ -527,3 +527,348 @@ def get_template_repos_by_name(template_name, username, password, template_url):
...
+
+    # Verify all features are enabled
+    for feature_name in ["remote_management", "content", "analytics"]:
+        assert features[feature_name]["enabled"] is True, (
+            f"Feature '{feature_name}' should be enabled when explicitly set"
+        )
+        assert features[feature_name]["successful"] is True, (
+            f"Feature '{feature_name}' should be successful"
+        )
+
```
jirihnidek: What about adding special testing function for the case, when all features are disabled?

integration-tests/test_connect.py
```diff
@@ -527,3 +527,348 @@ def get_template_repos_by_name(template_name, username, password, template_url):
...
+    assert rhc.is_registered
+    assert not yggdrasil_service_is_active()
+
+    json_output = json.loads(result.stdout)
+    features = json_output["features"]
+
+    # Map CLI feature names to JSON feature names
+    feature_mapping = {
+        "remote-management": "remote_management",
```
jirihnidek: Oh, it is pity that we use `remote-management` on the CLi and `remote_management` in the JSON output. I am sorry for that. We cannot do anything about this now due to backward compatibility.

jirihnidek (changes requested):
Thanks for the PR. Overall it looks good. I have only some ideas how to make the code more flexible and readable.

integration-tests/test_connect.py
pbartiko: Understand. The reason was that all the scenarios in `test_connect_with_feature_enabled_disabled_combinations` has the yggdrasil service inactive. The all features enabled has also the yggdrasil active so I took it as a special case and made a separate test. But I agree we can do the same for the all features disabled.

PR #304, opened 2025-11-25
fix(test): Do not try to install insight-client on Fedora

jirihnidek:
* Do not try to install `insights-client` on Fedora, because the `required insights-core` RPM is not available there

### DIFF
--- a/systemtest/tests/integration/test.sh
+++ b/systemtest/tests/integration/test.sh
@@ -9,19 +9,62 @@ if [[ -f "/etc/os-release" ]]; then
   source /etc/os-release
 fi
 
+function mock_insights_client() {
+  # Create configuration directory, when it does not exists
+  if [[ ! -d "/etc/insights-client/" ]]; then
+    mkdir -p /etc/insights-client/
+  fi
+  # Create empty configuration file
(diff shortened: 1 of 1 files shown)


systemtest/tests/integration/test.sh
```diff
@@ -4,14 +4,52 @@ set -ux
...
   # Check for GitHub pull request ID and install build if needed.
   # This is for the downstream PR jobs.
   [ -z "${ghprbPullId+x}" ] || ./systemtest/copr-setup.sh
 
-  dnf --setopt install_weak_deps=False install -y \
-    podman git-core python3-pip python3-pytest logrotate insights-client
+  if [[ "${ID}" = "fedora" ]]; then
+    # Do not try to install insights-client on Fedora, because it cannot be installed there
+    dnf --setopt install_weak_deps=False install -y \
+      podman git-core python3-pip python3-pytest logrotate
```
m-horky: I'd move this out of the `if` statement, back to where it was, it happens in both branches.

m-horky (changes requested):
Great to see green CI results. I have two notes that are both related to maintenance, the logic itself looks fine.

PR #306, opened 2025-11-25
Use primarily activation key and org as authentication method

pbartiko:
As Satellite doesn't support basic auth for now, use primarily activation key + organization


jirihnidek (approved):
LGTM

PR #309, opened 2025-11-26
feat(collector): Add collector configuration parsing

pkoprda:
* Card ID: CCT-1696

This commit implements objects and methods required to parse collector configurations.

### DIFF
--- a/internal/collector/collector.go
+++ b/internal/collector/collector.go
@@ -1,0 +1,128 @@
+package collector
+
+import (
+	"fmt"
+	"log"
+
+	"github.com/BurntSushi/toml"
+)
+
+// ConfigDir is the default directory path where collector configuration files are stored.
+const ConfigDir = "/usr/lib/rhc/collector/"
+const defaultMetaType = "ingress"
+const defaultUser = "root"
+const defaultGroup = "root"
+
+// Config represents the configuration for a collector instance.
+type Config struct {
+	// ID is the unique identifier for the collector.
+	ID string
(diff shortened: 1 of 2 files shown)


jirihnidek (changes requested):
There is no comment nor doc string. Please add doc strings to every structure, every attribute and every method.

m-horky (changes requested):
In my eyes, this looks great! I left two small requests inline.

internal/collector/collector.go
```diff
@@ -0,0 +1,122 @@
...
+// configDto represents the structure of a TOML configuration file for parsing.
+type configDto struct {
+	Meta    *metaDto    `toml:"meta"`
+	Ingress *ingressDto `toml:"ingress"`
+}
+
+// metaDto represents the metadata section of a TOML configuration file.
+type metaDto struct {
+	Name    string `toml:"name"`
+	Feature string `toml:"feature"`
```
m-horky: As a optional field, this should be `*string`, with `omitempty` tag.

internal/collector/collector.go
```diff
@@ -0,0 +1,122 @@
...
+	Type    string `toml:"type"`
+}
+
+// ingressDto represents the ingress section of a TOML configuration file.
+type ingressDto struct {
+	User        string `toml:"user"`
+	Group       string `toml:"group"`
```
m-horky: Same as `feature`.

m-horky (changes requested):

PR #313, opened 2025-12-05
feat(collector): Add collector configuration validation

pkoprda:
* Card ID: CCT-1697

This change implements function GetCollectors() which returns list of available collectors from valid TOML files in ConfigDir.

### DIFF
--- a/internal/collector/collector_test.go
+++ b/internal/collector/collector_test.go
@@ -1,6 +1,8 @@
 package collector
 
 import (
+	"io/fs"
+	"strings"
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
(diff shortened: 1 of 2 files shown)


internal/collector/collector_test.go
```diff
@@ -1,11 +1,18 @@
...
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
 )
 
+type mockDirEntry struct {
+	name  string
+	isDir bool
+}
+
```
m-horky: Can we move this struct just above its methods? Or the other way around; just to be together.

m-horky (changes requested):
I left several minor inline requests, otherwise it looks fine.

m-horky (approved):
LGTM

PR #316, opened 2025-12-09
feat(logging): implement support for file-based logging

jlocash:
* Card ID: CCT-1565
  * Adds structured logging with `log/slog` to connect, disconnect, and status commands
  * Implement custom log handler which handles log formatting (and includes PID)
  * Reconfigures `log/slog` to use the custom handler and write to `/var/log/rhc/rhc.log`
  * Brings content of logs to parity with CLI output
  * Updated RPM spec to create log directory during install

### DIFF
--- a/data/logrotate.d/meson.build
+++ b/data/logrotate.d/meson.build
@@ -1,0 +1,4 @@
+install_data(
+    'rhc',
+    install_dir: join_paths(get_option('sysconfdir'), 'logrotate.d'),
+)
--- a/data/logrotate.d/rhc
+++ b/data/logrotate.d/rhc
@@ -1,0 +1,11 @@
+/var/log/rhc/*.log {
+    rotate 90
+    daily
+    compress
+    dateext
+    maxage 90
+    missingok
+    notifempty
+    copytruncate
+    create 640 root root
+}
--- a/data/meson.build
+++ b/data/meson.build
@@ -1,2 +1,3 @@
 subdir('completion')
+subdir('logrotate.d')
 subdir('systemd')
--- a/dist/srpm/rhc.spec.in
+++ b/dist/srpm/rhc.spec.in
@@ -167,6 +167,7 @@ fi
 %if %{with rhcd_compat}
 %{_unitdir}/yggdrasil.service.d/rhcd.conf
 %endif
+%config(noreplace) %{_sysconfdir}/logrotate.d/rhc
 %attr(0755,root,root) %dir %{_localstatedir}/log/rhc
 
 %changelog


cmd/rhc/connect_cmd.go
```diff
@@ -58,6 +58,20 @@ func (connectResult ConnectResult) Error() string {
 	return result
 }
 
+func (connectResult ConnectResult) errorMessages() map[string]string {
```
m-horky: I'd like to see a docstring here.

cmd/rhc/connect_cmd.go
```diff
@@ -150,17 +164,146 @@ func beforeConnectAction(ctx *cli.Context) error {
 	return nil
 }
 
+func connectRHSM(ctx *cli.Context, connectResult *ConnectResult) {
```
m-horky: I'm not really happy these new methods are mutating their argument. It would be cleaner to have some kind of action struct that has the `connectResult` as its own variable. I.e.

```go
func (s *ConnectAction) TryRegisterRHSM(ctx *cli.Context) {
    returnedMsg, err := rhsm.RegisterRHSM(ctx, features.ConnectFeature.Enabled)
    if err != nil {
        s.result.RHSMConnected = false
        ...
```

cmd/rhc/logging.go
```diff
@@ -1,10 +1,216 @@
...
+
+// groupOrAttrs holds either a group name or a list of slog.Attrs.
+type groupOrAttrs struct {
+	group string      // group name if non-empty
+	attrs []slog.Attr // attrs if non-empty
+}
+
+type PIDHandler struct {
+	opts PIDHandlerOptions
+	goas []groupOrAttrs
```
m-horky: This field name feels very confusing.

internal/logging/pid_handler.go
```diff
@@ -0,0 +1,194 @@
...
+type PIDHandler struct {
+	opts PIDHandlerOptions
+	// preAttrs holds the groups and attributes added via WithGroup and WithAttrs.
+	// These are prepended to each log record's attributes with appropriate group prefixes.
+	preAttrs []groupOrAttrs
+	mu       *sync.Mutex
+	out      io.Writer
+}
+
+// NewPIDHandler creates a new PIDHandler that writes to out wit/quh the given options.
```
m-horky: > NewPIDHandler creates a new PIDHandler that writes to **out wit/quh** the given options.

Typo.

cmd/rhc/interactive.go
```diff
@@ -27,23 +27,25 @@ func showTimeDuration(durations map[string]time.Duration) {
...
-				if logMsg.level >= conf.Config.LogLevel {
-					_, _ = fmt.Fprintf(w, "%v\t%v\t%v\n", logMsg.level, step, logMsg.message)
-				}
-			}
-			_ = w.Flush()
-			if hasPriorityErrors(errorMessages, slog.LevelError) {
-				return cli.Exit("", 1)
-			}
-		}
+func showErrorMessages(action string, errorMessages map[string]string) error {
```
m-horky: This method no longer filters for `LogMessage`'s `slog.Level`, so all messages get printed, even if they would have been `DEBUG` before. Is that intentional? I don't think we've ever produced anything that wasn't `ERROR`, but it is still a change in behavior on paper.

m-horky (changes requested):
I'd like to see the new methods documented. Public ones for sure, but I'd prefer if all of them had a docstring. I appreciate the work so far.

cmd/rhc/main.go
```diff
@@ -21,6 +21,8 @@ const (
 	cliAPIServer = "base-url"
 )
 
+var logCleanup func() error
```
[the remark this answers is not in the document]
m-horky: Thinking more about it -- no other logging library I know of requires a cleanup to be run. Look at what they are doing, and replicate that.

jlocash:
> When running the code, if subscription-manager D-Bus API returns an error, we still display success:
> 
> ```
>  [𐄂] Cannot connect to Red Hat Subscription Management
>   [𐄂] Skipping generation of Red Hat repository file
>   [𐄂] Skipping connection to Red Hat Insights
>   [𐄂] Skipping activation of yggdrasil service
> 
> Successfully connected to Red Hat!
> 
> Manage your connected systems: https://red.ht/connector
> 
> STEP       DURATION  
> rhsm       5.189s    
> insights   0s        
> yggdrasil  0s        
> 
> The following errors were encountered during connect:
> 
> STEP  ERROR  
> rhsm  cannot connect to Red Hat Subscription Management: error: Server error attempting a GET to /subscription/users/USER/owners returned status 403
> Forbidden: Invalid credentials for request.
> ```

@m-horky This was pre-existing behavior. I suggest we add two additional messages:

When `TryRegisterRHSM` fails: 
```
Connection to Red Hat failed. Unable to register system.

...
```

When `TryRegisterRHSM` succeeds, but one or more of the others fail:
```
Connected to Red Hat, but one or more features could not be enabled.

...
```

m-horky (commented):
It's a shame standard library's `slog.commonHandler` isn't available to us, the implementation of `slog.TextHandler` gets very elegant with it :(

To be honest, I'm afraid to take in 194 new lines that fully reimplement the stdlib handler, even if they have 623 lines of tests. I'm guessing the tests are autogenerated, but your commits don't include any sign-off or mention of AI tools, so it is hard to tell.

There are two things I've noticed when reading the [Go's README](https://github.com/golang/example/blob/master/slog-handler-guide/README.md) on implementing handlers:

- The handler doesn't process `r.PC` (source location information). We don't have it set up here, but I'm guessing we'd want that at some point.
- The use of `fmt.Appendf` causes an allocation for each modification of the log statement; the official example [recommends](https://github.com/golang/example/blob/master/slog-handler-guide/README.md#speed) using `append()` directly.

---

Given how soon this is supposed to be delivered, how about we solve the PID thing in some other way?

Line [main.go#144](https://github.com/RedHatInsights/rhc/blob/19c36edf889c45bc848aaa1c9d037608dc437496/cmd/rhc/main.go#L144) you updated already contains PID, which is mostly what we'd need from it (even if your design specifies it a bit differently).

I looked back and it was my suggestion to include it in each message and deal with concurrent executions (around November 3rd), but this PR feels like a lot of effort for an edge case that might not happen (we're not concurrency-aware anywhere within the program). Having PID included in the first log statement every time we start up is good enough I think.

For this time, I'd let the idea of our own handler go, and ensure we're certain the logging won't break at any point (even though I haven't discovered any obvious edge cases while reading the code).

What do you think?

m-horky:
> This was pre-existing behavior.

I'll file a card for that, let's not overload the content of this PR.

m-horky:
The main deliverable the product is tracking is "having file-based logs". We do not need to include PIDs at all, that came as a technical improvement idea. If we discover we need them later, we'll include them. The same applies for having ability to have custom format. Having a single log statement that logs the PID is good enough for me right now, we don't need to have it part of every line.

I'll mostly leave the evaluation up to you -- whether it is worth having this code in or not. To me, every line of code is a liability :) I'd like to let others express their opinions, maybe I'm just overly cautious. @jirihnidek, would you mind looking at it as well?

jirihnidek:
When you run `rhc status` as non-root user, then you get similar corrupted output, when the `/var/log/rhc/rhc.log` exists:

```console
jhnidek@localhost:~/$ rhc status
2026/01/08 13:44:23 WARN unable to open log file, falling back to stderr error="open /var/log/rhc/rhc.log: permission denied" path=/var/log/rhc/rhc.log
2026-01-08T13:44:23+01:00 PID=545857 INFO rhc started version="0.3.5" pid=545857
Connection status for thinkpad:

2026-01-08T13:44:23+01:00 PID=545857 INFO Checking system connection status
2026-01-08T13:44:23+01:00 PID=545857 INFO Not connected to Red Hat Subscription Management
 [ ] Not connected to Red Hat Subscription Management
2026-01-08T13:44:23+01:00 PID=545857 WARN Red Hat repository file not generated
  [ ] Content ... Red Hat repository file not generated
2026-01-08T13:44:23+01:00 PID=545857 INFO Not connected to Red Hat Insights
  [ ] Analytics ... Not connected to Red Hat Insights
2026-01-08T13:44:23+01:00 PID=545857 WARN The yggdrasil service is inactive but not loaded
  [ ] Remote Management ... The yggdrasil service is inactive but not loaded

```

This is regression!

If journald was used, then this would not be issue. I do not have advice how to solve this issue.

cmd/rhc/connect_cmd.go
jirihnidek: Please add doc string to all new functions.

jirihnidek:
When you connect and it is not possible to start `yggdrasil.service` (because `mosquitto.service` is not running and yggdrasil is trying to connect to it), then you get false positive result:

```console
root@localhost:/# rhc connect --username admin --password admin --organization donaldduck
Connecting thinkpad to Red Hat.

Features preferences: [✓]content, [✓]analytics, [✓]remote-management

  [✓] Analytics ... Connected to Red Hat Insights
  [✓] Remote Management ... Activated the yggdrasil service

```

jirihnidek:
When I run `rhc status` as non-root user, then it is still not possible to write to log file.

```console
jhnidek@localhost:~/path/to/rhc$ ./builddir/cmd/rhc/rhc status
2026/01/09 10:26:58 WARN unable to open log file error="open /var/local/log/rhc/rhc.log: permission denied" path=/var/local/log/rhc/rhc.log

  [✓] Analytics ... Connected to Red Hat Insights
  [ ] Remote Management ... The yggdrasil service is inactive but not loaded

```

The XDG specification https://specifications.freedesktop.org/basedir/latest/ says that client application should write log files somewhere in `$HOME/.local/state` ... it is default path that could be defined by `$XDG_STATE_HOME` environment variable, but this environment variable is not defined on Fedora nor RHEL. The directory `$HOME/.local/state` exists and we could create directory `$HOME/.local/state/rhc` and write there the `$HOME/.local/state/rhc/rhc.log` in such case. Such behavior should be documented in some man page.

cmd/rhc/disconnect_cmd.go
```diff
@@ -80,82 +85,100 @@ func disconnectService(disconnectResult *DisconnectResult, errorMessages *map[st
...
 		disconnectResult.YggdrasilStopped = false
 		disconnectResult.YggdrasilStoppedError = errMsg
+		slog.Error(errMsg)
 		ui.Printf(" [%v] %v\n", ui.Icons.Error, errMsg)
 	} else {
 		disconnectResult.YggdrasilStopped = true
-		ui.Printf(" [%v] Deactivated the %v service\n", ui.Icons.Ok, ServiceName)
+		infoMsg := fmt.Sprintf("Deactivated the %s service", ServiceName)
+		slog.Info(infoMsg)
+		ui.Printf(" [%v] %v\n", ui.Icons.Info, infoMsg)
```
jirihnidek: This icon `ui.Icons.Info` is confusing. We should use `ui.Icons.Ok` here as we did use it in the original code.

cmd/rhc/connect_cmd.go
```diff
@@ -212,148 +358,32 @@ func connectAction(ctx *cli.Context) error {
...
-				ServiceName,
-			)
-		}
-	}
+	/* 3. Start yggdrasil (rhcd) service */
+	start = time.Now()
+	connectResult.TryActivateServices()
+	durations[ServiceName] = time.Since(start)
+
+	slog.Info("Successfully connected to Red Hat")
```
jirihnidek: This should be conditional. We should print this message only, when `rhc connect` is successful.

cmd/rhc/disconnect_cmd.go
```diff
@@ -187,32 +212,35 @@ func disconnectAction(ctx *cli.Context) error {
...
 	durations["insights"] = time.Since(start)
 
 	/* 3. Unregister system from Red Hat Subscription Management */
 	start = time.Now()
-	_ = disconnectRHSM(&disconnectResult, &errorMessages)
+	_ = disconnectResult.TryUnregisterRHSM()
 	durations["rhsm"] = time.Since(start)
 
+	slog.Info("Successfully disconnected from Red Hat")
+	slog.Info("Manage your connected systems: https://red.ht/connector")
```
jirihnidek: This should not be printed to log file.

cmd/rhc/disconnect_cmd.go
jlocash: This was unintentional. Will fix

cmd/rhc/main.go
```diff
@@ -56,6 +62,33 @@ func configureUI(ctx *cli.Context) {
...
+	// Attempt to open the log file
+	// This file path typically resolves to /var/log/rhc/rhc.log
+	logFilePath := filepath.Join(LogDir, LongName, LongName+".log")
+	file, err := os.OpenFile(logFilePath, os.O_APPEND|os.O_WRONLY|os.O_CREATE, 0640)
+
+	var w io.Writer
+	if err != nil {
+		// Discard log messages if we can't open the log file
+		w = io.Discard
+		slog.Warn("unable to open log file", "error", err, "path", logFilePath)
```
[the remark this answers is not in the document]
jirihnidek: Oh, I overlooked this. When you call `slog.Warn("unable to open log file", "error", err, "path", logFilePath)`, then this message is printed to `stderr`. Thus, I thought that other messages will be also printed to stderr.

We should write here the information, that other messages will be dropped. Maybe, we should not write this warning as a log message. Maybe, we should write it using `ui.Printf()`. Otherwise, it looks like that other log messages will be printed also to `stderr`.

cmd/rhc/main.go
```diff
@@ -56,6 +62,33 @@ func configureUI(ctx *cli.Context) {
 	)
 }
 
+// configureFileLogging sets up file-based logging to the configured log file path.
+// If the log file can't be opened, it falls back to io.Discard, effectively ignoring all log messages.
+func configureFileLogging(logLevel slog.Leveler) {
```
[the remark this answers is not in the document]
jlocash: Implemented this as `ensureLogDirectory` in my latest changes. If root, it will create and return `/var/log/rhc`, or `~/.local/state/rhc` if non-root.

cmd/rhc/connect_cmd.go
```diff
@@ -58,6 +58,155 @@ func (connectResult ConnectResult) Error() string {
...
+func (connectResult *ConnectResult) TryRegisterInsightsClient() {
+	if !features.AnalyticsFeature.Enabled {
+		connectResult.Features.Analytics.Successful = false
+		slog.Info("Connecting to Red Hat Insights disabled (analytics feature disabled)")
+		ui.Printf("%s[ ] Analytics ... Connecting to Red Hat Insights disabled\n", ui.Indent.Medium)
+		return
+	}
+
+	if connectResult.RHSMConnectError != "" {
+		slog.Error("Skipping connection to Red Hat Insights (RHSM registration failed)", "rhsm_error", connectResult.RHSMConnectError)
```
jirihnidek: Again the severity should be lower, because real error happened earlier. The warning lever would be enough.

cmd/rhc/logging.go
```diff
@@ -1,10 +1,112 @@
...
-	message error
+// ensureLogDirectory ensures that the log directory exists and is writable by the current user.
+// If the directory doesn't exist, it is created.
+// If the program is running as root, the log directory is created under /var/log/rhc.
+// Otherwise, the log directory is created under ~/.local/state/rhc.
+func ensureLogDirectory() (string, error) {
+	isRootUser := os.Getuid() == 0
+
+	if isRootUser {
+		// This path resolves to /var/log/rhc
```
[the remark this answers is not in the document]
m-horky: @jirihnidek I see. We may as well hardcode `/var/log/rhc/` in constants.go directly, there's no reason to have any logic around `LocalStateDir="/var/"` on top if it.

Archana-PandeyM (commented):
LGTM.
Minor nit: rotate 90 + maxage 90 are redundant with **daily** rotation. Not an issue, just optional cleanup.

PR #318, opened 2025-12-11
test: update testware to run tests with proxy in stage

Archana-PandeyM:
When running tests in stage env downstream, we need the requests to go via an internal proxy if the test system is in aws  or in cloud in general. 
So modified the tmt prepare to detect the environment and set proxy in all configuration files that affects rhc functionality- rhsm.conf, insights-client.conf and yggdrasil service

### DIFF
--- a/systemtest/configure_proxy_stage.py
+++ b/systemtest/configure_proxy_stage.py
@@ -1,0 +1,141 @@
+#!/usr/bin/env python3
+"""Configure system-wide proxy settings when ENV_FOR_DYNACONF is 'stage'.
+
+Uses local settings.toml if present, otherwise downloads from SETTINGS_URL.
+"""
+
+import configparser
+import os
+import subprocess
+import sys
+import urllib.error
+import urllib.request
+from pathlib import Path
+
+try:
+    import tomllib  # Python 3.11+
+except ModuleNotFoundError:
+    import tomli as tomllib
+
+
+ENV_VAR = "ENV_FOR_DYNACONF"
+TARGET_ENV = "stage"
+
+RHSM_CONF = Path("/etc/rhsm/rhsm.conf")
+INSIGHTS_CONF = Path("/etc/insights-client/insights-client.conf")
+YGGDRASIL_OVERRIDE_DIR = Path("/etc/systemd/system/yggdrasil.service.d")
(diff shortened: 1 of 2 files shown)


Archana-PandeyM:
> This looks good!
> 
> only problem I see is that the sed commands might break in some cases. The sed on line 47 uses `|` as a delimiter, which is very unlikely to be in a proxy url so that one is relatively save, but the other seds (lines 38 and 48) use `/` as a delimiter, and it is very possible that a url could contain a `/` (eg `http://foo.bar/`).
> 
> So we have a few options to fix this:
> 
> 1. use a `|` for the other delimiters
> 2. use an escape function:
> 
> ```
> # Escape special characters for sed replacement strings
> escape_sed() {
>     printf '%s\n' "$1" | sed 's/[&/\|]/\\&/g'
> }
> 
> # Then use it when setting variables:
> host_safe=$(escape_sed_replacement "$host")
> proxy_url_safe=$(escape_sed_replacement "$proxy_url")
> ```
> 
> 3. or you can do this all in python which has a safe way to edit these files using `import configparser`

In our case there is no ''/" in config file for these proxy parameters, So it's probably fine, but to be consistent and safe I have decide to roll back all this to a python script (Initially I started with python script for ease of config parser and then I switched to bash, I dont know why. Probably I just wanted to use AI :) ).

systemtest/configure_proxy_stage.py
```diff
@@ -0,0 +1,135 @@
...
+YGGDRASIL_OVERRIDE_DIR = Path("/etc/systemd/system/yggdrasil.service.d")
+YGGDRASIL_OVERRIDE = YGGDRASIL_OVERRIDE_DIR / "proxy.conf"
+PROFILE_PROXY = Path("/etc/profile.d/proxy.sh")
+
+
+def load_proxy_settings() -> tuple[str, str, str]:
+    """Download and extract noauth_proxy host/port from [stage] section."""
+    settings_url = os.environ.get("SETTINGS_URL")
+    if not settings_url:
+        raise RuntimeError("SETTINGS_URL must be provided")
```
Lorquas: This might break the ability to run tmt locally. We should only check for settings_url if there isn't a local settings.toml file in the top level of the project.

Lorquas (approved):
LGTM

PR #319, opened 2025-12-12
feat(collector): Add data collection compression

pkoprda:
* Card ID: CCT-1701

Add GetArchive() and createArchive() functions to compress a directory into binary xz archive.

### DIFF
--- a/internal/collector/collector.go
+++ b/internal/collector/collector.go
@@ -2,7 +2,6 @@ package collector
 
 import (
 	"fmt"
-	"log"
 	"log/slog"
 	"os"
 	"os/exec"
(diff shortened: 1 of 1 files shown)


internal/collector/collector.go
```diff
@@ -86,6 +99,16 @@ func GetConfig(id string) (Config, error) {
 	return config, nil
 }
 
+// createArchive compresses a directory into a xz-compressed tar archive.
+// Returns an error if the tar command fails.
+func createArchive(archiveName, tmpDir string) error {
+	cmd := exec.Command("tar", "-cJf", archiveName, "-C", tmpDir, ".")
```
m-horky: This command saves the archive into the current directory. Instead, it should store it in a path passed as an additional argument (meaning we'll need to pass two directory values: one to store the data in, other containing the data to compress).

In production, the data store directory will likely be `/var/tmp/rhc/` (to mimic whatever insights-client is doing).

internal/collector/collector.go
```diff
@@ -86,6 +99,16 @@ func GetConfig(id string) (Config, error) {
...
+	if err := cmd.Run(); err != nil {
```
m-horky: Do we care about whatever `tar` may print to standard output and error? If so, `cmd.CombinedOutput()` or using `.StdinPipe()+.StderrPipe()` might provide ways to enure our log contains useful information.

m-horky (changes requested):
I left some minor requests inline. I think this PR will be ready to be merged after they are addressed.

PR #335, opened 2026-01-19
chore: Vendor packages on Fedora 

m-horky:
* Card ID: CCT-1886
* Card ID: CCT-1929

Since Fedora Rawhide/44+ now enforces Go dependency vendorization, update the .spec file to match those requirements.

Also, use the `%gocheck2` macro instead of deprecated `%gocheck`.

https://fedoraproject.org/wiki/Changes/GolangPackagesVendoredByDefault
https://docs.fedoraproject.org/en-US/packaging-guidelines/Golang/

This PR does _not_ switch Packit from using the dist/ version of .spec file.

### DIFF
--- a/rhc.spec
+++ b/rhc.spec
@@ -1,0 +1,118 @@
+# Run unit tests during package build
+%global with_check 1
+# Include rhcd->yggdrasil compatibility patch
+%global with_rhcd_compat 0
+
+%global goipath         github.com/redhatinsights/rhc
+Version:                0.3.8
+
+%gometa -L -f
+
+Name:           rhc
+Release:        %autorelease
+Summary:        Tool for registering to Red Hat services
+License:        Apache-2.0 AND BSD-2-Clause AND BSD-3-Clause AND GPL-3.0-only AND MIT
+URL:            %{gourl}
+Source0:        %{gosource}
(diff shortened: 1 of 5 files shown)


rhc.spec
```diff
@@ -0,0 +1,115 @@
...
+%go_vendor_license_install -c %{S:2}
+# Binaries
+install -m 0755 -vd                     %{buildroot}%{_bindir}
+install -m 0755 -vp _build/bin/*        %{buildroot}%{_bindir}/
+# Bash completion
+install -m 0755 -vd                     %{buildroot}%{bash_completions_dir}/
+install -m 0644 -vp data/completion/rhc.bash  %{buildroot}%{bash_completions_dir}/%{name}
+# Man page
+install -m 0755 -vd                     %{buildroot}%{_mandir}/man1
+install -m 0644 -vp rhc.1               %{buildroot}%{_mandir}/man1/rhc.1
```
jirihnidek: You should also install `doc/source/markdown/rhc-configure.8.md`

m-horky: I don't think we can do that yet, we don't have the logic to convert markdown to ruff. There is an internal epic tracking that: CCT-1847.

PR #336, opened 2026-01-20
Clean up constants.go

m-horky:
* Card ID: CCT-1927
* Card ID: CCT-1930

The variables defined in constants.go were inherited from yggdrasil
during the initial split (38a1dff) and were designed to support runtime
discovery of installation directories. We have no use for that.

ShortName or LongName were injected at compile time. Because the
original expectations no longer hold, we can simplify the code by
dropping them from the code and build system.

This Pull Request should not change any behavior or output of rhc, upstream nor downstream.


cmd/rhc/constants.go
```diff
@@ -26,93 +24,15 @@ var (
...
-var (
-	PrefixDir         string
-	BinDir            string
-	SbinDir           string
-	LibexecDir        string
-	DataDir           string
-	DatarootDir       string
-	ManDir            string
-	DocDir            string
-	SysconfDir        string
```
jirihnidek: This could be useful for `/etc/rhc/` ... We have hard-coded this path several times in the code. It is not great.

PR #343, opened 2026-02-04
feat(collector): Implement timer cache functionality

pkoprda:
* Card ID: CCT-1843

Implement timer cache functionality to track collector execution start or finish times. This commit includes two public API methods - one for reading the cache and the other one for writing to the cache.

### DIFF
--- a/internal/collector/collector.go
+++ b/internal/collector/collector.go
@@ -15,7 +15,7 @@ import (
 )
 
 // ConfigDir is the default directory path where collector configuration files are stored.
-const ConfigDir = "/usr/lib/rhc/collector/"
+const ConfigDir = "/usr/lib/rhc/collectors/"
 
 // TimerDir is the default directory path where information about collectors execution are stored.
 const TimerDir = "/var/cache/rhc/collectors/"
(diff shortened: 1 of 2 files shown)


internal/collector/collector.go
```diff
@@ -36,6 +38,18 @@ type Config struct {
 	ContentType string
 }
 
+// Timer represents the execution timing information for a collector.
+type Timer struct {
```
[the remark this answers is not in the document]
m-horky: Or perhaps to its own file completely. But we can split this file into several smaller ones once we get there.

internal/collector/collector.go
m-horky: I know the chance for exploitability is very low, but how about we do some sanitization before we read and write files that can be controlled from the outside?

We could pass the `id` parameter through `filepath.Base()` to get rid of weird non-id-like strings that would be attempting to do path traversal, but in general since we know the shape of collector IDs (`something.something.something...`), how about we return an error if a public method (the security boundary) receives an `id` that doesn't match regex `^[a-z]+(\.[a-z]+)*$`?

m-horky: This regex is relatively open, perhaps enforcing at least two or three `a-z` groups from the start would also help to guide the names for collectors as they get integrated.

m-horky (changes requested):
Looking good!

internal/collector/collector.go
m-horky: I'd like to keep the IDs lowercase, but I don't have anything against digits. `a-z0-9` works for me.

PR #345, opened 2026-02-06
feat(tests): Add integration tests for rhc connect with features config

DuckBoss:
This patch aims to add integration tests for the implementation of drop-in configuration file support for connectivity features. The tests cover cases ranging from default usage,
CLI/config features overriding, and edge cases with partial config files and feature dependency resolution.

Resolves: CCT-1973

AI Usage Disclosure: This commit contains test cases that were created with the assistance of AI tools.

### DIFF
--- a/integration-tests/conftest.py
+++ b/integration-tests/conftest.py
@@ -77,3 +77,78 @@ Environment=HTTP_PROXY={proxy_url}
 
     except Exception as e:
         logger.error(f"Error during yggdrasil proxy cleanup: {e}")
+
+
+@pytest.fixture(scope="function")
+def features_config_file():
+    """
+    Fixture to manage the RHC features drop-in configuration file.
+    Creates the config directory and file, and cleans up after test completion.
+
+    Usage:
+        def test_example(features_config_file):
+            features_config_file.write('features = { "content" = true }')
+            # ... run test ...
+            # Cleanup happens automatically after test
+
+    The fixture provides:
+        - write(content: str): Write TOML content to the config file
+        - remove(): Remove the config file if it exists
+        - path: The path to the config file
+        - dir: The path to the config directory
+    """
+    RHC_CONFIG_DIR = "/etc/rhc/config.toml.d"
+    RHC_FEATURES_CONFIG_FILE = f"{RHC_CONFIG_DIR}/01-features.toml"
+    class FeaturesConfigManager:
+        def __init__(self):
+            self.dir = RHC_CONFIG_DIR
+            self.path = RHC_FEATURES_CONFIG_FILE
+            self._dir_existed = os.path.exists(self.dir)
+            self._file_existed = os.path.exists(self.path)
+            self._original_content = None
+            if self._file_existed:
+                with open(self.path, "r") as f:
+                    self._original_content = f.read()
+
+        def write(self, content: str):
+            """Write the features configuration file with the given TOML content"""
+            os.makedirs(self.dir, exist_ok=True)
+            with open(self.path, "w") as f:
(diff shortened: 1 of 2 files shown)


integration-tests/test_connect.py
```diff
@@ -876,3 +876,436 @@ def test_connect_with_nonexistent_feature(external_candlepin, rhc, test_config,
...
+            {"content": True, "analytics": True, "remote_management": True},
+            False,
+        ),
+        # Config enables analytics/remote-management, CLI disables content - dependency conflict (exit 64)
+        (
+            'features = { "content" = true, "analytics" = true, "remote-management" = true }',
+            [],
+            ["content"],
+            None,  # No expected states - command should fail
+            True,
```
Archana-PandeyM:
```suggestion
            64,
```

integration-tests/test_connect.py
```diff
@@ -876,3 +876,436 @@ def test_connect_with_nonexistent_feature(external_candlepin, rhc, test_config,
...
+        ),
+        # Config enables remote-management, CLI disables analytics - dependency conflict (exit 64)
+        (
+            'features = { "content" = true, "analytics" = true, "remote-management" = true }',
+            [],
+            ["analytics"],
+            None,  # No expected states - command should fail
+            True,
```
Archana-PandeyM:
```suggestion
            64,
```

integration-tests/test_connect.py
```diff
@@ -876,3 +876,436 @@ def test_connect_with_nonexistent_feature(external_candlepin, rhc, test_config,
...
+        test_config, auth="activation-key", output_format="json"
+    )
+
+    # Add CLI flags to override config
+    for feature in cli_enable:
+        command_args.extend(["--enable-feature", feature])
+    for feature in cli_disable:
+        command_args.extend(["--disable-feature", feature])
+
+    command = ["connect"] + command_args
```
Archana-PandeyM: result = rhc.run("connect", *command_args, check=(expected_return_code == 0))

    assert result.returncode == expected_return_code, (
        f"Expected return code {expected_return_code}, but got {result.returncode}"
    )

    if expected_return_code != 0:
        # Dependency conflict: CLI disables a feature that config-enabled features depend on
        assert not rhc.is_registered, (
            "System should not be registered when CLI flag conflicts with config file"
        )
        return

    assert rhc.is_registered

    # Parse JSON output and verify feature states
    json_output = json.loads(result.stdout)
    features = json_output["features"]

    for feature_key, expected_enabled in expected_states.items():
        actual_enabled = features[feature_key]["enabled"]
        assert actual_enabled is expected_enabled, (
            f"Feature '{feature_key}' should have enabled={expected_enabled} "
            f"(CLI override), but got enabled={actual_enabled}"
        )

    # Verify yggdrasil service state
    if expected_states.get("remote_management", False):
        assert yggdrasil_service_is_active()
    else:
        assert not yggdrasil_service_is_active()

Archana-PandeyM: I am suggesting this above code block for rest of test code. I could not find a way to suggest multiline code .

Archana-PandeyM:
Thanks for the PR Jason. Overall it looks good.

Non-blocking Suggestion: split config-focused connect tests into a dedicated module to keep test_connect.py maintainable.
for example- Create integration-tests/test_connect_config_file.py 

I would like to hear your thoughts too.

Archana-PandeyM:
I think it would be better to add some negative tests to cover the malformed toml file, so that errors should not be silently ignored.
e.g.
1- Invalid TOML
2- Unknown feature key

PR #349, opened 2026-02-13
 chore: Remove meson from Packit

m-horky:
Use the new rhc.spec and drop Meson from build process even for Packit.

'packit build locally' seems to have been broken for some time, this patch fixes it.

Blocked by #335, needs to be rebased after that one is merged.


jirihnidek (dismissed):
LGTM

Note: It seems that it is not possible to run some actions after `rpmbuild` and reset `rhc.spec` (https://packit.dev/docs/configuration/actions). The `rhc.spec` is modified by `packit` after each run of `packit build`.

PR #401, opened 2026-04-13
Identify rhc while uploading to ingress

mjcr99:
## Description

Hi team, this PR adds the User-Agent header info when uploading to provide information about the origin.

This in one example of the expected User-Agent added info:
```console
rhc-collector/0.3.8 (triggered-by: Red Hat Lightspeed Advisor) rhel/10.1
```

## Testing:

To facilitate the manual test we can modify the following lane, to send the request to a local mocked server in order to inspect it:

```console
[vagrant@lv-rhel101 rhc]$ sed -i 's|https://cert.console.redhat.com|http://localhost:3000|' cmd/rhc-collector/main.go
```

Also we need to create a testing collector:
```console
[vagrant@lv-rhel101 rhc]$ mkdir -p /usr/lib/rhc/collectors
[vagrant@lv-rhel101 rhc]$ nano  /usr/lib/rhc/collectors/com.redhat.advisor.toml
[vagrant@lv-rhel101 rhc]$ cat /usr/lib/rhc/collectors/com.redhat.advisor.toml
[meta]
name = "Red Hat Lightspeed Advisor"
feature = "analytics"
type = "ingress"

[ingress]
user = "root"
group = "root"
content_type = "application/vnd.redhat.advisor.collection"
```

Then build the project:
```console
[vagrant@lv-rhel101 rhc]$ make build
...
```

In another terminal use the following python script to serve the requests and leave it running:

```python3
#!/usr/bin/env python3

from http.server import HTTPServer, BaseHTTPRequestHandler

class Handler(BaseHTTPRequestHandler):
    def do_POST(self):
        ua = self.headers.get("User-Agent", "")

        print("\n=== Incoming Request ===")
        print(f"User-Agent: {ua}")

if __name__ == "__main__":
    print("Mock server on http://localhost:3000")
    HTTPServer(("0.0.0.0", 3000), Handler).serve_forever()
```

```console
[vagrant@lv-rhel101 ~]$ python3 mock_ingress.py 
Mock server on http://localhost:3000
```

And in the previous terminal, execute the compiled `rhc-collector` binary:
```console
[vagrant@lv-rhel101 rhc]$ sudo ./rhc-collector com.redhat.advisor "insights-client --output-dir"
2026/04/13 15:05:26 INFO starting rhc-collector id=com.redhat.advisor
2026/04/13 15:05:26 INFO created temporary directory dir=/var/tmp/rhc/collector-1370553453
2026/04/13 15:05:26 INFO configuration of the collector config="{ID:com.redhat.advisor Name:Red Hat Lightspeed Advisor IsAnalyticsFeature:true User:root Group:root ContentType:application/vnd.redhat.advisor.collection}"
2026/04/13 15:05:39 INFO collector has ran successfully output="Starting to collect Insights data for lv-rhel101\nWriting RHSM facts to /etc/rhsm/facts/insights-client.facts ...\nCollected data copied to /var/tmp/rhc/collector-1370553453\n"
2026/04/13 15:05:40 INFO archive created path=/var/tmp/rhc/rhc-collector-20260413150539683.tar.xz
2026/04/13 15:05:40 INFO Uploading archive archive=/var/tmp/rhc/rhc-collector-20260413150539683.tar.xz url=http://localhost:3000
2026/04/13 15:05:40 INFO Successfully uploaded archive archive=/var/tmp/rhc/rhc-collector-20260413150539683.tar.xz
```

The server will show the following message:
```console
=== Incoming Request ===
User-Agent: rhc-collector/0.3.8 (triggered-by: Red Hat Lightspeed Advisor) rhel/10.1
```

### DIFF
--- a/cmd/rhc-collector/main.go
+++ b/cmd/rhc-collector/main.go
@@ -7,6 +7,7 @@ import (
 	"os/exec"
 
 	"github.com/redhatinsights/rhc/internal/collector"
+	httpapi "github.com/redhatinsights/rhc/internal/http"
 )
 
 // FIXME: Make these configurable (use the values from "rhc configure")
@@ -15,6 +16,11 @@ const (
 	clientCertPath = "/etc/pki/consumer/cert.pem"
 	clientKeyPath  = "/etc/pki/consumer/key.pem"
 	rhcTmpDir      = "/var/tmp/rhc"
+)
+
+var (
+	// Version is set at build time.
+	Version = "dev"
 )
 
 func main() {
@@ -48,7 +54,7 @@ func run(collectorId, command string) error {
 		return err
 	}
 	defer cleanup(archivePath)
-	if err = uploadArchive(archivePath, config.ContentType); err != nil {
+	if err = uploadArchive(archivePath, config); err != nil {
 		return err
 	}
 	return nil
@@ -107,17 +113,18 @@ func getArchivePath(tmpDir string) (string, error) {
 
 // uploadArchive uploads the created archive to Red Hat Hybrid Cloud Console.
 // Returns an error if the upload fails.
-func uploadArchive(archivePath, contentType string) error {
+func uploadArchive(archivePath string, collectorConfig collector.Config) error {
 	archive := collector.ArchiveDto{
 		Path:        archivePath,
-		ContentType: contentType,
+		ContentType: collectorConfig.ContentType,
 	}
 	serviceConfig := collector.ServiceConfig{
 		URL:            ingressUrl,
 		ClientCertPath: clientCertPath,
 		ClientKeyPath:  clientKeyPath,
 	}
-	if err := collector.UploadArchive(archive, serviceConfig); err != nil {
+	userAgent := httpapi.GetUserAgent("rhc-collector", Version, collectorConfig.ID)
+	if err := collector.UploadArchive(archive, serviceConfig, userAgent); err != nil {
 		slog.Error("failed to upload archive", "error", err)
 		return fmt.Errorf("failed to upload archive: %w", err)
 	}
--- a/internal/collector/ingress.go
+++ b/internal/collector/ingress.go
@@ -39,7 +39,7 @@ type ServiceConfig struct {
 }
 
 // UploadArchive uploads an archive file to the Red Hat Hybrid Cloud Console.
-func UploadArchive(archive ArchiveDto, config ServiceConfig) error {
+func UploadArchive(archive ArchiveDto, config ServiceConfig, userAgent string) error {
 	if err := validateArchive(archive); err != nil {
 		return err
 	}
@@ -54,7 +54,7 @@ func UploadArchive(archive ArchiveDto, config ServiceConfig) error {
 		return err
 	}
 	client := httpapi.NewHTTPClient(tlsConfig)
-	req, err := createUploadRequest(formData, config)
+	req, err := createUploadRequest(formData, config, userAgent)
 	if err != nil {
 		return err
 	}
@@ -165,8 +165,7 @@ func createMultipartForm(archive ArchiveDto) (multipartData, error) {
 
 // createUploadRequest creates an HTTP POST request for uploading multipart form data.
 // Returns an error if request creation fails.
-func createUploadRequest(formData multipartData, config ServiceConfig) (*http.Request, error) {
-	// FIXME: Add proper User-Agent header to identify the rhc
+func createUploadRequest(formData multipartData, config ServiceConfig, userAgent string) (*http.Request, error) {
 	req, err := http.NewRequest("POST", config.URL, formData.Buffer)
 	if err != nil {
 		slog.Error("Failed to create request", "error", err)
@@ -174,6 +173,8 @@ func createUploadRequest(formData multipartData, config ServiceConfig) (*http.Re
 	}
 	req.Header.Set("Content-Type", formData.ContentType)
 	req.Header.Set("Accept", "application/json")
+	req.Header.Set("User-Agent", userAgent)
+
 	return req, nil
 }
 
--- a/internal/http/useragent.go
+++ b/internal/http/useragent.go
@@ -1,0 +1,108 @@
+package httpapi
+
+import (
+	"fmt"
+	"log/slog"
+	"os"
+	"strings"
+	"sync"
+)
+
+const (
+	// unknownOS is returned when OS information cannot be determined.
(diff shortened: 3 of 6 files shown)


internal/collector/ingress.go
```diff
@@ -38,8 +38,37 @@ type ServiceConfig struct {
...
+	// TriggeredBy is the name/ID of the data collector or Varlink caller
+	TriggeredBy string
+}
+
+// NewCollectorCallerInfo creates a CallerInfo for a collector-triggered upload.
+func NewCollectorCallerInfo(collectorConfig Config, componentVersion string) CallerInfo {
+	return CallerInfo{
+		ComponentName:    "rhc-collector",
+		ComponentVersion: componentVersion,
+		TriggeredBy:      collectorConfig.Name,
```
m-horky: Since the collector names may contain spaces and may change over time, we should pass in ID and not Name.

cmd/rhc-collector/main.go
```diff
@@ -17,6 +17,11 @@ const (
 	rhcTmpDir      = "/var/tmp/rhc"
 )
 
+var (
+	// Version is set at build time.
+	Version = "dev"
+)
```
m-horky, line 23: I find it unfortunate we have to redefine Version in every package. I don't have any better solution though.

m-horky (changes requested):
I have some notes after the first quick look. Non-code note: one of the commit messages is referencing a wrong old ticket. And in general I'd appreciate describing why each commit is being made in its message.

internal/http/useragent.go
```diff
@@ -0,0 +1,130 @@
...
+		env[key] = value
+	}
+
+	if err := scanner.Err(); err != nil {
+		slog.Debug("Failed to scan os-release file", "path", path, "error", err)
+		return nil, fmt.Errorf("failed to scan os-release file: %w", err)
+	}
+
+	return env, nil
+}
```
m-horky, line 111: This is a lot of code to read two values. Can you instead create something more generic? We could later reuse it if we had to:

```go
func loadEnv(path string) (map[string]string, error) {
    data, _ = os.ReadFile(path)
    env := make(map[string]string)
    for _, line := range strings.Split(string(data), "\n") {
        // clean up (trim spaces, skip empty and comment lines)
        // split on the first =, save as env[k]=v
    return env, nil
}
```

That method would not work for actual shell files with multiline strings and so on, but we know the format of os-release is much more conservative.

internal/http/useragent.go
```diff
@@ -0,0 +1,130 @@
...
+//   - ComponentVersion: build-time version string injected via ldflags
+//   - TriggeredBy: validated collector ID or Varlink method name
+type UserAgentInfo struct {
+	// ComponentName is the name of the rhc component (e.g., "rhc-collector", "rhc-server")
+	ComponentName string
+	// ComponentVersion is the version of the rhc component
+	ComponentVersion string
+	// TriggeredBy is the name/ID of the data collector or varlink caller
+	TriggeredBy string
+}
```
m-horky: This API doesn't feel ergonomic -- it has internal state, but its user has to provide all the state, devaluating it to a pure wrapper that formats it into a string. Why not use a method?

```go
func getUserAgent(component, version, caller string) string {
    release, _ := loadEnv("/usr/lib/os-release")
    return fmt.Sprinf("%s/%s (triggered-by: %s) %s", ...)
}
```

^ That's a simplified example, I like your idea of caching the OS information.

m-horky (changes requested):

internal/http/useragent.go
mjcr99: Addressed, I have simplified it.

internal/http/useragent.go
mjcr99: Good point, addressed!

internal/http/useragent.go
```diff
@@ -0,0 +1,111 @@
...
+		version,
+		triggeredBy,
+		osVersion,
+	)
+}
+
+// getOSVersion reads the OS version from /etc/os-release.
+// Returns ID/VERSION_ID (e.g., "rhel/9.3", "fedora/39", "centos/9").
+// The result is cached to avoid repeated file I/O.
+func getOSVersion() string {
```
m-horky, line 51: Strictly speaking, we're getting both name and version of the OS, not just the version. How about `getOSIdentifier` or something like that?

internal/http/useragent.go
```diff
@@ -0,0 +1,111 @@
...
+	for _, line := range strings.Split(string(data), "\n") {
+		// Trim whitespace
+		line = strings.TrimSpace(line)
+
+		// Skip empty lines and comments
+		if line == "" || strings.HasPrefix(line, "#") {
+			continue
+		}
+
+		// Split on the first =
```
m-horky, line 98: We may drop these comments, I think the code is pretty self-explanatory. What do you think?

m-horky (changes requested):
Looking good! My notes are very minor.

internal/http/useragent.go
mjcr99: Sure, makes sense.

internal/http/useragent.go
mjcr99: You are right, removed.

internal/collector/ingress.go
```diff
@@ -38,8 +38,37 @@ type ServiceConfig struct {
...
+
+// NewVarlinkCallerInfo creates a CallerInfo for a Varlink API-triggered upload.
+func NewVarlinkCallerInfo(varlinkCaller string, componentVersion string) CallerInfo {
+	return CallerInfo{
+		ComponentName:    "rhc-server",
+		ComponentVersion: componentVersion,
+		TriggeredBy:      varlinkCaller,
+	}
+}
+
```
m-horky: Do we need this object at all? Can't we construct the user-agent string using the method `httpapi.GetUserAgent` immediately, instead of using this CallerInfo object as a proxy? I'm not convinced we're gaining anything with this abstraction.

m-horky (changes requested):

internal/collector/ingress.go
mjcr99: hmmm, yeah, I think you are right, it was intended to better adapt for the two different formats that there may be with rhc-collector and rhc-server inputs, but yeah, this is overkill.

m-horky (approved):
I have not reviewed unit tests for time capacity reasons, I trust you with them. LGTM on the production code, it looks much cleaner! Merging.

PR #403, opened 2026-04-14
feat(feature): Add feature management system

m-horky:
* Card ID: CCT-2169

Add a new 'pkg/feature' package providing unified interface for managing feature levels. Features may declare dependencies and can be enabled and disabled programatically.

Since the number of features is expected to stay quite small, I have reached a conclusion that an implementation of a DAG or otherwise "smart" resolution of what depends on what would only make things complicated.

The number of features might grow in the future, but they will stay sortable and without cycles; hardcoding their dependencies feels like the simplest way to keep it small and maintainable.

### DIFF
--- a/pkg/feature/analytics.go
+++ b/pkg/feature/analytics.go
@@ -1,0 +1,36 @@
+package feature
+
+import (
+	"github.com/redhatinsights/rhc/internal/datacollection"
+)
+
+// Analytics implements IFeature.
+type Analytics struct{}
+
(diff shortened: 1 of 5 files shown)


mjcr99 (approved):
Thanks @m-horky! 
I'm approving this PR because this commit is a building block for a more complex feature; it's simple and straightforward, and the actual underlying logic isn't reflected yet, so I don't have any objections at this point.

PR #415, opened 2026-04-16
feat(feature): Add feature management CLI

m-horky:
* Card ID: CCT-2169

This patch implements 'feature status/enable/disable' commands, supporting both registered and unregistered systems.

Currently, it is not possible to change the feature level through 'rhc' itself: users may manually run 'subscription-manager', 'insights-client' and 'systemctl enable/disable yggdrasil', but that exposes implementation details of rhc.

---

I'm not sure this is the best way to split pre-/post-registration methods. The conditional approach works, but it creates additional context load on the developers. It also does not handle JSON formatting at all. We might need more aggressive refactoring to achieve that, though I'm not sure whether we should block this feature on that or not.

### DIFF
--- a/cmd/rhc/configure_features_cmd.go
+++ b/cmd/rhc/configure_features_cmd.go
@@ -1,0 +1,314 @@
+package main
+
+import (
+	"fmt"
+	"log/slog"
+	"os"
+	"text/tabwriter"
+
+	"github.com/urfave/cli/v2"
+
+	"github.com/redhatinsights/rhc/internal/rhsm"
+	"github.com/redhatinsights/rhc/pkg/feature"
+	"github.com/redhatinsights/rhc/pkg/feature/prefcache"
+)
+
+// TODO Handle machine-readable output by always returning a DTO from business logic;
+//  a place here should only act as a presentation layer.
+
+// TODO All methods should return 'cli.ExitCoder' instead of plain 'error'
+
+// TODO Use ui.Icons.Ok when we have UTF-8 capable tabwriter
+
+// beforeFeaturesStatusAction validates inputs before executing the status action.
+func beforeFeaturesStatusAction(ctx *cli.Context) error {
+	err := checkFormatFlag(ctx)
+	if err != nil {
+		return err
+	}
+	configureUI(ctx)
+	return checkForUnknownArgs(ctx)
+}
+
+// featuresStatusAction displays the current status or preferences of all features.
+func featuresStatusAction(ctx *cli.Context) error {
+	logCommandStart(ctx)
+	isRegistered, err := rhsm.IsRHSMRegistered()
+	if err != nil {
+		return err
+	}
+
+	if isRegistered {
+		return featuresStatusActionRegistered(ctx)
+	}
+	return featuresStatusActionNotRegistered(ctx)
+}
+
+func featuresStatusActionNotRegistered(_ *cli.Context) error {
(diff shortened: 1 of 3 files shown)


cmd/rhc/configure_features_cmd.go
```diff
@@ -0,0 +1,328 @@
...
+
+// featuresStatusAction displays the current status or preferences of all features.
+func featuresStatusAction(ctx *cli.Context) error {
+	logCommandStart(ctx)
+	isRegistered, err := rhsm.IsRHSMRegistered()
+	if err != nil {
+		return err
+	}
+
+	switch isRegistered {
```
mjcr99: This works perfectly, but using a switch for this simple conditional seems a bit overengineered for me.

Same for the other used switchs.

cmd/rhc/configure_features_cmd.go
```diff
@@ -0,0 +1,328 @@
...
+
+// featuresEnableActionNotRegistered handles enabling a feature on a non-registered system.
+func featuresEnableActionNotRegistered(ctx *cli.Context, targetName string) error {
+	cache, err := prefcache.LoadCache(ConnectFeaturesPrefsPath)
+	if err != nil {
+		return cli.Exit(fmt.Sprintf("error: failed to load feature preferences: %v", err), ExitCodeSoftware)
+	}
+
+	target, err := feature.Get(targetName)
+	if err != nil {
```
mjcr99: I think this check is redundant, since we already checked it in `beforeFeaturesEnableAction`, I would suggest using MustGet() since validation should already have happened.

Same thought for the other cases where a previous check is done in the `before...` function.

cmd/rhc/configure_features_cmd.go
```diff
@@ -0,0 +1,328 @@
...
+	{
+		enabled, err := cache.Get(targetName)
+		if err != nil {
+			return cli.Exit(fmt.Sprintf("error: failed to get feature preference: %v", err), ExitCodeSoftware)
+		}
+		if !enabled {
+			fmt.Printf("During registration, '%s' will be enabled.\n", targetName)
+			if err = cache.Set(targetName, true); err != nil {
+				return cli.Exit(fmt.Sprintf("error: failed to update preference: %v", err), ExitCodeSoftware)
+			}
```
mjcr99, line 154: In case `enabled == true` we could let the user know with an else block like:
``` 
} else {                                                                                                                                                                                                                                   
    fmt.Printf("Feature '%s' is already set to be enabled during registration.\n", targetName)
}
```

mjcr99 (changes requested):
Thanks for the PR @m-horky! So far, I have some comments.

cmd/rhc/configure_features_cmd.go
m-horky: Hehe, I tried it out for aesthetic reasons. I'll change it back to if-else blocks.

cmd/rhc/configure_features_cmd.go
m-horky: I'm not sure this provides any valuable information. I wanted to focus on the _diff_ of that action, not whole system state.

cmd/rhc/configure_features_cmd.go
m-horky: Good point, addressed.

mjcr99 (approved):
LGTM

PR #416, opened 2026-04-20
feat(feature): Use preference cache for 'connect' 

m-horky:
* Card ID: [CCT-2169](https://redhat.atlassian.net/browse/CCT-2169)

First commit adds a warning icon the UI can display. Some situations cannot be expressed with only Ok, Info and Error: a feature level being skipped during 'connect' because its dependency failed fits none of those categories.

Second commit updates 'rhc connect' code to use preference feature cache and the new feature resolution logic. When 'rhc configure features' is run on a system that is not registered, a preference cache is written to the filesystem. Later, during 'connect', this cache is loaded and used as if '--enable-feature' or '--disable-feature' have been passed in to the command. If these flags are passed in, the preference cache is ignored and a notice is displayed to the user.

Third commit does clean-up and adds TODOs and comments.

---

To test this PR, perform various combinations of 'rhc configure features' followed by 'rhc connect'. Whatever 'rhc configure features status' shows before registration should be applied during registration. If `--enable|disable-feature` flag is passed during registration, anything defined with 'rhc configure' is ignored.

I'm aware of the size and amount of individual changes in the second commit, but I did not find a better way to split it up even more. I think the best way to review the functionality is to look at it as a black-box, without trying to mentally parse the differences in behavior.

[CCT-2169]: https://redhat.atlassian.net/browse/CCT-2169?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


cmd/rhc/connect_cmd.go
```diff
@@ -114,31 +115,14 @@ func (connectResult *ConnectResult) TryRegisterRHSM(ctx *cli.Context) {
...
 	slog.Info("Connecting to Red Hat Lightspeed")
 	err := ui.Spinner(datacollection.RegisterInsightsClient, ui.Indent.Medium, "Connecting to Red Hat Lightspeed (formerly Insights)...")
 	if err != nil {
 		connectResult.Features.Analytics.Successful = false
-		connectResult.Features.Analytics.Error = fmt.Sprintf("cannot connect to Red Hat Lightspeed (formerly Insights): %v", err)
+		connectResult.Features.Analytics.Error = fmt.Sprintf("cannot connect to Red Hat Lightspeed: %v", err)
 		slog.Error(fmt.Sprintf("cannot connect to Red Hat Lightspeed: %v", err))
 		ui.Printf(
-			"%s[%v] Analytics ... Cannot connect to Red Hat Lightspeed (formerly Insights)\n",
+			"%s[%v] Analytics ... Cannot connect to Red Hat Lightspeed\n",
```
pkoprda, line 125: Why are you removing `(formerly Insights)`? Wouldn't be that confusing for some customers?

cmd/rhc/connect_cmd.go
```diff
@@ -202,21 +158,73 @@ func (connectResult *ConnectResult) TryActivateServices() {
...
 }
 
-// beforeConnectAction ensures that user has supplied correct CLI options
-// and there is no conflict between them. When there is anything wrong,
-// then this function will invoke cli.Exit() with an appropriate error
-// message and error code. The exit codes are defined in the
-// constants.go module
+// checkFeatureFlags validates --enable-feature and --disable-feature flag combinations.
+// Returns an error if the combination is invalid.
+func checkFeatureFlags(toEnable, toDisable []string) error {
```
pkoprda, line 163: This method has too many for loops. I understand that the iteration may not be 1000, but I think it may be simplified. What about an approach using lookups instead of comparisons in for loops?
```go
enabledSet := make(map[string]bool)
disabledSet := make(map[string]bool)

for _, f := range toEnable {
  if disabledSet[f] {
    return fmt.Error()
  }
  enabledSet[f] = true
}

// similar for `range toDisable`

// check dependencies
for _, e := range toEnable {
    f, err := feature.Get(e)
    if err != nil {
        return err
    }
    for _, dep := range f.Requires() {
        if disabledSet[dep] {
            return fmt.Errorf("invalid combination: enable '%s', disable '%s'", e, dep)
        }
    }
}

// check reverse dependencies
```

cmd/rhc/connect_cmd.go
```diff
@@ -296,106 +329,139 @@ func beforeConnectAction(ctx *cli.Context) error {
...
 }
 
-// connectAction tries to register system against Red Hat Subscription Management,
-// gather the profile information that the system will configure
-// connect system to Red Hat Lightspeed, and it also tries to start rhcd service
+// connectAction manages 'rhc connect' steps:
+// first we register against Red Hat Subscription Management,
+// then we enable data collection for Red Hat Lightspeed services,
+// then we start remote management service yggdrasil.
 func connectAction(ctx *cli.Context) error {
```
pkoprda, line 336: I am a little worried, this method has too many lines and the complexity is just rising, I've headache from just understanding what this method do :(. But I don't want to block the PR on this, can you at least add some `TODO refactoring` comment for this method?

pkoprda (changes requested):
Thanks for the PR, I have left some comments below.

cmd/rhc/connect_cmd.go
m-horky: I don't like it and I think this change was a mistake, but it was done by an accident I think.

cmd/rhc/connect_cmd.go
m-horky: I'm quite unhappy with it as well. I've added a FIXME, and I'd like to create a downstream card as well. We have multiple TODOs and FIXMEs at the code now, we may go through and do it at once later.

cmd/rhc/connect_cmd.go
m-horky: Addressed. I'm not sure it is worth it, O(n³) isn't slow for single digits of items, but I am not opposed. It might be making the code slightly less readable, I've added unit tests for this method.

pkoprda (changes requested):
Thanks for the updates.

I see some test failures:
```
FAILED integration-tests/test_connect.py::test_connect_with_single_feature_disabled[content]
FAILED integration-tests/test_connect.py::test_connect_with_feature_enabled_disabled_combinations[enabled_features6-disabled_features6-None-True]
FAILED integration-tests/test_connect.py::test_connect_with_nonexistent_feature[--enable-feature]
FAILED integration-tests/test_connect.py::test_connect_with_nonexistent_feature[--disable-feature]
```

Also I've noticed that after using `rhc disconnect`, the preference status is changed from `disabled` to `enable` for every feature. Is that expected?
```
# rhc configure features status
Not connected to Red Hat.

FEATURE            PREFERENCE  DESCRIPTION
content            enable      Red Hat content management
analytics          enable      Red Hat Lightspeed data collection
remote-management  enable      Red Hat Lightspeed remote management
# rhc configure features disable analytics
During registration, 'remote-management' will not be enabled (depends on 'analytics').
During registration, 'analytics' will not be enabled.
# rhc configure features status                                                     
Not connected to Red Hat.

FEATURE            PREFERENCE  DESCRIPTION
content            enable      Red Hat content management
analytics          skip        Red Hat Lightspeed data collection
remote-management  skip        Red Hat Lightspeed remote management
# rhc connect
Connecting kvm-03-guest10.lab.eng.rdu2.dc.redhat.com to Red Hat. Enabled features: content.
This might take some time.

Username: pkoprda_ethel
Password: 

  [●] Analytics ... Skipped
  [●] Remote Management ... Skipped

# rhc configure features status
Connected to Red Hat.

FEATURE            STATE     DESCRIPTION
content            enabled   Red Hat content management
analytics          disabled  Red Hat Lightspeed data collection
remote-management  disabled  Red Hat Lightspeed remote management
# rhc disconnect               
Disconnecting kvm-03-guest10.lab.eng.rdu2.dc.redhat.com from Red Hat.

 [●] The yggdrasil service is already inactive
 [●] Already disconnected from Red Hat Lightspeed (formerly Insights)
 [✓] Disconnected from Red Hat Subscription Management
# rhc configure features status
Not connected to Red Hat.

FEATURE            PREFERENCE  DESCRIPTION
content            enable      Red Hat content management
analytics          enable      Red Hat Lightspeed data collection
remote-management  enable      Red Hat Lightspeed remote management
```

m-horky:
> I see some test failures

Good point, fixed.

> I've noticed that after using rhc disconnect, the preference status is changed from disabled to enable for every feature.

Yeah, this is part of the design. It might not be obvious this way, but the other would also be complicated: let's say you run 'rhc configure features disable remote-management, 'rhc connect', 'rhc configure features enable remote-management', 'rhc disconnect', 'rhc connect'. Should yggdrasil be enabled or not?

We decided to go with the simple-to-document way of "you always have to do the configuration before you connect, no matter the previous state."

pkoprda (approved):
Alright then, looks good to me, approving. However I hope the code will be refactored at some point.

m-horky:
The RHEL CI failing on `test_logging.py::test_log_connect_with_disabled_feature` is being addressed in https://github.com/RedHatInsights/pytest-client-tools/pull/50. Merging.

PR #417, opened 2026-04-20
Add socket activation tests for Varlink API

mjcr99:
* Card ID: CCT-1756

## Description

Hi team, this PR updates the `integration-tests/test_collector_api.py` to use the new reorganized utils for `systemctl` and `varlinkctl`, and also implements new tests to verify that an FD3 socket activation boots `rhc-server` on a `varlink` call.

### DIFF
--- a/integration-tests/test_socket_activated_service.py
+++ b/integration-tests/test_socket_activated_service.py
@@ -1,0 +1,87 @@
+"""
+:casecomponent: rhc
+:requirement: RHSS-XXXXX
+:subsystemteam: rhel-sst-csi-client-tools
+:caseautomation: Automated
+:upstream: Yes
+"""
+
+import pytest
+
+from utils.varlink import run_varlinkctl
+from utils.systemctl import (
+    is_service_active,
+    is_socket_enabled,
+    stop_service,
+    enable_and_start_socket,
+    disable_and_stop_socket,
+)
+
+
+@pytest.fixture(scope="module")
+def fd3_socket_setup():
+    """
+    Fixture to ensure rhc-server.socket is enabled for FD3 socket activation tests.
+    Stops the service if running to test auto-boot behavior.
+    """
+    socket_name = "rhc-server.socket"
+    service_name = "rhc-server.service"
+
+    socket_was_enabled = is_socket_enabled(socket_name)
+
+    if not socket_was_enabled:
+        enable_and_start_socket(socket_name)
+
+    if is_service_active(service_name):
+        stop_service(service_name)
+
+    yield {
+        "socket_name": socket_name,
+        "service_name": service_name,
+        "socket_was_enabled": socket_was_enabled,
+    }
+
+    if not socket_was_enabled:
+        disable_and_stop_socket(socket_name)
+
+
+@pytest.mark.tier1
+def test_fd3_socket_activation(fd3_socket_setup):
+    """
+    :id: d1e2f3a4-b5c6-7890-def1-23456789abcd
+    :title: Verify FD3 socket activation boots rhc-server on varlink call
+    :reference: https://redhat.atlassian.net/browse/CCT-1756
+    :description:
+        Test that when rhc-server.service is not running but rhc-server.socket
+        is enabled, making a varlink call triggers systemd socket activation,
+        automatically boots the service, and returns a correct response.
+    :tags: Tier 1
+    :steps:
+        1. Ensure rhc-server.service is stopped
+        2. Ensure rhc-server.socket is enabled
+        3. Verify service is not active before the call
+        4. Make a varlink call (com.redhat.rhc.collector.List)
+        5. Verify the service becomes active (auto-booted)
+        6. Verify the response is correct and well-formed
+    :expectedresults:
+        1. Service is stopped successfully
+        2. Socket is enabled
+        3. Service is not active before varlink call
+        4. Varlink call succeeds
+        5. Service becomes active after the call
+        6. Response contains the expected "collectors" field
+    """
+    service_name = fd3_socket_setup["service_name"]
+
+    assert not is_service_active(
+        service_name
+    ), f"Service {service_name} should not be active before FD3 call"
+
+    response = run_varlinkctl("com.redhat.rhc.collector.List")
+
+    assert "collectors" in response, "Response should contain 'collectors' field"
+    assert isinstance(response["collectors"], list), "'collectors' should be a list"
+
+    assert is_service_active(
+        service_name
+    ), f"Service {service_name} should be active after FD3 socket activation"
--- a/integration-tests/utils/systemctl.py
+++ b/integration-tests/utils/systemctl.py
@@ -1,0 +1,74 @@
+"""
+Systemctl-related utility functions for integration tests.
+
+This module provides helper functions for managing systemd services, sockets,
+and units using systemctl commands.
+"""
+
+import subprocess
+
+
+def is_service_active(service_name: str) -> bool:
+    """
+    Check if a systemd service is active.
+
+    :param service_name: Name of the systemd service
+    :return: True if the service is active, False otherwise
+    """
+    result = subprocess.run(
+        ["systemctl", "is-active", service_name],
+        capture_output=True,
+    )
+    return result.returncode == 0
+
+
+def stop_service(service_name: str) -> None:
+    """
+    Stop a systemd service.
+
+    :param service_name: Name of the systemd service to stop
+    """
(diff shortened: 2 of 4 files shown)


integration-tests/utils/systemctl.py
```diff
@@ -0,0 +1,116 @@
...
+Systemctl-related utility functions for integration tests.
+
+This module provides helper functions for managing systemd services, sockets,
+and units using systemctl commands.
+"""
+
+import subprocess
+
+
+def is_service_active(service_name):
```
m-horky: Could you type hint the methods?

integration-tests/test_socket_activated_service.py
```diff
@@ -0,0 +1,98 @@
...
+        3. Service is not active before varlink call
+        4. Varlink call succeeds
+        5. Service becomes active after the call
+        6. Response contains expected "collectors" field
+    """
+    service_name = fd3_socket_setup["service_name"]
+
+    # Ensure service is stopped before test
+    if is_service_active(service_name):
+        stop_service(service_name)
```
m-horky: The fixture already does this. Why are we doing it twice?

integration-tests/test_socket_activated_service.py
```diff
@@ -0,0 +1,98 @@
...
+    response = run_varlinkctl("com.redhat.rhc.collector.List")
+
+    # Verify response structure is correct
+    assert "collectors" in response, "Response should contain 'collectors' field"
+    assert isinstance(response["collectors"], list), "'collectors' should be a list"
+
+    # Verify the service became active due to socket activation
+    assert is_service_active(
+        service_name
+    ), f"Service {service_name} should be active after FD3 socket activation"
```
m-horky: This assert isn't part of the steps/results you put into the method docstring. The assert itself is correct, but is there a reason to make this assertion? Are there any product or technical requirements that say the service should stay running?

integration-tests/test_socket_activated_service.py
```diff
@@ -0,0 +1,98 @@
...
+    Fixture to ensure rhc-server.socket is enabled for FD3 socket activation tests.
+    Stops the service if running to test auto-boot behavior.
+    """
+    socket_name = "rhc-server.socket"
+    service_name = "rhc-server.service"
+
+    # Check initial states
+    socket_was_enabled = is_socket_enabled(socket_name)
+
+    # Enable and start the socket if not already enabled
```
m-horky: These comments feel like a noise, the code feels self-documenting...

m-horky (dismissed):

integration-tests/utils/systemctl.py
mjcr99: Sure, good suggestion!

integration-tests/test_socket_activated_service.py
mjcr99: You are right, it's redundant.

integration-tests/test_socket_activated_service.py
mjcr99: It represents point 5, both in `steps` and `expected results`. 
It's not explicitly stated to make this assertion, but I think it makes sense to assert it since getting a correct answer to a request should imply the service has been correctly booted.

integration-tests/test_socket_activated_service.py
mjcr99: Addressed.

integration-tests/test_socket_activated_service.py
```diff
@@ -0,0 +1,86 @@
...
+    if not socket_was_enabled:
+        disable_and_stop_socket(socket_name)
+
+
+@pytest.mark.tier1
+def test_fd3_socket_activation(fd3_socket_setup):
+    """
+    :id: d1e2f3a4-b5c6-7890-def1-23456789abcd
+    :title: Verify FD3 socket activation boots rhc-server on varlink call
+    :description:
```
pkoprda, line 54: Can you add Jira reference CCT-1756 to this docstring? E.g., see an example in insights-client: https://github.com/RedHatInsights/insights-client/blob/4a0d9720be5b5bb3f05a002ea688d63545672449/integration-tests/test_compliance.py#L43

integration-tests/utils/systemctl.py
```diff
@@ -0,0 +1,116 @@
...
+    """
+    Disable a systemd socket.
+
+    :param socket_name: Name of the systemd socket
+    """
+    subprocess.run(
+        ["systemctl", "disable", socket_name],
+        check=False,
+        capture_output=True,
+    )
```
pkoprda: Why `check=false`? Also `check=false` is default assigned value for `subprocess.run()`, so it doesn't make sense to have it here

integration-tests/utils/varlink.py
```diff
@@ -0,0 +1,38 @@
...
+from typing import Optional, Union
+
+
+def run_varlinkctl(
+    method: str,
+    params: Optional[dict] = None,
+    check: bool = True
+) -> Union[dict, subprocess.CompletedProcess]:
+    """
+    Helper function to call varlinkctl and return parsed JSON response.
```
pkoprda: Missing article
```suggestion
    Helper function to call varlinkctl and return a parsed JSON response.
```

integration-tests/test_socket_activated_service.py
```diff
@@ -0,0 +1,86 @@
...
+        4. Make a varlink call (com.redhat.rhc.collector.List)
+        5. Verify the service becomes active (auto-booted)
+        6. Verify the response is correct and well-formed
+    :expectedresults:
+        1. Service is stopped successfully
+        2. Socket is enabled
+        3. Service is not active before varlink call
+        4. Varlink call succeeds
+        5. Service becomes active after the call
+        6. Response contains expected "collectors" field
```
pkoprda: Missing article
```suggestion
        6. Response contains the expected "collectors" field
```

integration-tests/utils/systemctl.py
pkoprda, line 1: Some of the `subprocess.run()` in this file has `text` parameter set on, some of them not. I find this a bit confusing, can you clarify why some of them have this parameter set on and some of them not?

pkoprda (changes requested):
Thanks for the PR, I left some comments below

integration-tests/test_socket_activated_service.py
mjcr99: Addressed

integration-tests/utils/systemctl.py
mjcr99: Good point, you are right, addressed.

integration-tests/utils/systemctl.py
mjcr99: You are right, since none of these functions actually read the output content (they only check return codes or rely on check=True for exceptions), we don't need text=True at all.

pkoprda, line 1: Are the methods `start_service()`, `enable_socket()` and `disable_socket()` even used somewhere?

integration-tests/utils/systemctl.py
```diff
@@ -0,0 +1,112 @@
...
+
+import subprocess
+
+
+def is_service_active(service_name: str) -> bool:
+    """
+    Check if a systemd service is active.
+
+    :param service_name: Name of the systemd service
+    :return: True if service is active, False otherwise
```
pkoprda:
```suggestion
    :return: True if the service is active, False otherwise
```

integration-tests/utils/systemctl.py
```diff
@@ -0,0 +1,112 @@
...
+        capture_output=True,
+    )
+
+
+def is_socket_enabled(socket_name: str) -> bool:
+    """
+    Check if a systemd socket is enabled.
+
+    :param socket_name: Name of the systemd socket
+    :return: True if socket is enabled, False otherwise
```
pkoprda:
```suggestion
    :return: True if the socket is enabled, False otherwise
```

pkoprda (changes requested):

integration-tests/utils/systemctl.py
mjcr99: Addressed

integration-tests/utils/systemctl.py
mjcr99: Addressed

integration-tests/utils/systemctl.py
mjcr99: Nop, they were used in a previous commit version, but not now, removed.

pkoprda (approved):
LGTM, thanks for the updates, merging

PR #420, opened 2026-04-21
[0.2] fix(ci): Rebase GitHub Actions workflows

m-horky:
- Constrain workflow permissions
- Use latest Go version for linting
- Instead of multiple hardcoded Go versions, use "stable"

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,6 +1,11 @@
 name: Lint
 
-on: pull_request
+on:
+  pull_request:
+  workflow_dispatch:
+
+permissions:
+  contents: read
 
 jobs:
   lint-commits:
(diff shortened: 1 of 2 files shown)


.github/workflows/lint.yml
```diff
@@ -19,34 +24,28 @@ jobs:
...
           version: "latest"
           args: "--verbose"
-      - name: Install shellcheck
+
+      - name: "Install shellcheck"
         run: sudo apt install shellcheck
-      - name: Check shell bash scripts
+      - name: "Run shellcheck"
         run: shellcheck ./rhc.bash
-  lint-language:
```
mjcr99, line 43: Why do we want to remove this job?

mjcr99 (changes requested):

.github/workflows/lint.yml
m-horky: It's being removed by mistake, I didn't realize we have it and use it in main as well.

mjcr99 (approved):
LGTM

PR #422, opened 2026-04-23
refactor: standardize error handling

Peto-RH:
CardID: CCT-2168

Replaced magic number exit codes with named constants and normalized error message formatting.

### DIFF
--- a/cmd/rhc-collector/main.go
+++ b/cmd/rhc-collector/main.go
@@ -22,13 +22,13 @@ const (
 func main() {
 	if len(os.Args) <= 2 {
 		slog.Error("usage: rhc-collector COLLECTOR-ID COMMAND")
-		os.Exit(1)
+		os.Exit(ExitCodeUsage)
 	}
 	collectorId, command := os.Args[1], os.Args[2]
 	slog.Info("starting rhc-collector", slog.String("id", collectorId))
 	if err := run(collectorId, command); err != nil {
 		slog.Error("rhc-collector exited with error", "error", err)
-		os.Exit(1)
+		os.Exit(ExitCodeErr)
 	}
 }
 
--- a/cmd/rhc-collector/constants.go
+++ b/cmd/rhc-collector/constants.go
@@ -1,0 +1,6 @@
+package main
+
+const (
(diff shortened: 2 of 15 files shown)


Peto-RH:
Note for reviewer: 
It might make more sense having error constants file directly in `/cmd` dir since `rhc` and `rhc-collector` and `rhc-server` share those constants

cmd/rhc-collector/main.go
```diff
@@ -16,6 +16,10 @@ const (
 	clientCertPath = "/etc/pki/consumer/cert.pem"
 	clientKeyPath  = "/etc/pki/consumer/key.pem"
 	rhcTmpDir      = "/var/tmp/rhc"
+
+	// Exit codes
+	ExitCodeErr   = 1  // generic error
+	ExitCodeUsage = 64 // command line usage error
```
pkoprda: Why introduce new constants? Why not to use already defined constants from [constants.go](https://github.com/RedHatInsights/rhc/blob/main/cmd/rhc/constants.go)?
Also as you already suggested, it would be better to move `constants.go` into `cmd/` directory.

pkoprda (changes requested):
There are still some pending `error: ` prefixes in `cmd/rhc/connect_cmd.go` (probably re-introduced by #416) and also `server error:` prefix in [cmd/rhc-server/main.go](https://github.com/RedHatInsights/rhc/blob/57eeec3a3ffa952b848cf6ada8cb09a9bb896ce3/cmd/rhc-server/main.go#L44).

cmd/rhc-collector/main.go
Peto-RH: I put it into `cmd/constants` directory if that is also okay

Peto-RH: I moved just the shared error messages. The rest of the file (specific to `cmd/rhc`) I left in the original file

pkoprda: Is that `cmd/constants/` directory really necessary?

pkoprda (approved):
LGTM, thanks!

PR #327, opened 2026-01-07
feat(ui): Add UTF-8 aware string handling for table display

cnsnyder:
- Implemented displayWidth and truncateString functions to handle UTF-8 characters correctly.
- Enhanced PrintTable function to support proper column alignment and truncation based on terminal width, ensuring better display of strings in tables.

At the moment this does not change any used functionality in rhc, hence why it is in draft. After upcoming work to read from the config files for user preferences the functions herein (along with a few extra bits to tie things together) will implement the "rhc configure features show" command.

Opening this PR to get early feedback on the approach and results of this approach so as to make best use of time.

The following script in a single go file can be used to demonstrate the additions to ui.go:

```go
package main

import (
	"fmt"
	"strings"
	"unicode/utf8"
)

func displayWidth(s string) int {
	return utf8.RuneCountInString(s)
}

func truncateString(s string, maxWidth int) string {
	if displayWidth(s) <= maxWidth {
		return s
	}
	targetWidth := maxWidth - 3
	if targetWidth < 0 {
		targetWidth = 0
	}
	width := 0
	truncated := strings.Builder{}
	for _, r := range s {
		runeWidth := displayWidth(string(r))
		if width+runeWidth > targetWidth {
			break
		}
		truncated.WriteRune(r)
		width += runeWidth
	}
	return truncated.String() + "..."
}

func PrintTable(table [][]string, sep string, termWidth int) {
	if len(table) == 0 || len(table[0]) == 0 {
		return
	}
	sepWidth := displayWidth(sep)
	columnWidths := make([]int, len(table[0]))
	for _, row := range table {
		for col, cell := range row {
			cellWidth := displayWidth(cell)
			if cellWidth > columnWidths[col] {
				columnWidths[col] = cellWidth
			}
		}
	}
	tableOutput := ""
	for _, row := range table {
		rowString := ""
		for col, cell := range row {
			rowString += cell
			if col < len(row)-1 {
				cellWidth := displayWidth(cell)
				padding := columnWidths[col] - cellWidth + sepWidth
				if padding > 0 {
					rowString += strings.Repeat(" ", padding)
				}
				rowString += sep
			}
		}
		if displayWidth(rowString) > termWidth {
			rowString = truncateString(rowString, termWidth)
		}
		tableOutput += rowString + "\n"
	}
	fmt.Printf("%s", tableOutput)
}

func main() {
	fmt.Println("Test 1: Basic table with UTF-8")
	table1 := [][]string{
[...]

PR #425, opened 2026-04-29
feat(cli): rhc collector CLI

pkoprda:
* Card ID: CCT-1845

Add comprehensive collector CLI interface with commands:
- rhc collector info <ID> - Display collector information
- rhc collector list - List available collectors
- rhc collector timers - List collector timers
- rhc collector enable [--now] <ID> - Enable timer-based collection
- rhc collector disable [--now] <ID> - Disable timer-based collection

An example of how to test this PR:
```
# systemctl enable rhc-server.socket rhc-server.service
# systemctl start rhc-server.socket rhc-server.service
# mkdir -p /usr/lib/rhc/collectors/
# cat /usr/lib/rhc/collectors/com.redhat.advisor.toml
[meta]
name = "Red Hat Lightspeed Advisor"
feature = "analytics"
type = "ingress"

[ingress]
user = "root"
group = "root"
content_type = "application/vnd.redhat.advisor.collection"
# mkdir -p /var/cache/rhc/collectors/
# cat /var/cache/rhc/collectors/com.redhat.advisor.json
{
  "last_started": {"timestamp": 1777461050},
  "last_finished": {"timestamp": 1777461150, "exit_code": 0}
}

# rhc collector
NAME:
   rhc collector - Collect data for analysis

USAGE:
   rhc collector COMMAND [command options]

COMMANDS:
   info     Display collector information
   list     List available collectors
   timers   List collector timers
   enable   Enable timer-based collection
   disable  Disable timer-based collection
   help, h  Shows a list of commands or help for one command

OPTIONS:
   --help, -h  show help
# rhc collector list
ID                  NAME
com.redhat.advisor  Red Hat Lightspeed Advisor
# rhc collector info com.redhat.advisor
Name:      Red Hat Lightspeed Advisor
Feature:   analytics

Last run:  Wed 2026-04-29 07:12 EDT (2h 48m ago)
Next run:  -

Config:   /usr/lib/rhc/collectors/com.redhat.advisor.toml
Service:  rhc-collector-com.redhat.advisor.service
Timer:    rhc-collector-com.redhat.advisor.timer
# rhc collector timers
ID                  LAST    NEXT
com.redhat.advisor  2h 48m  -

Hint: Run 'rhc collector info COLLECTOR' to show more details.
```

### DIFF
--- a/cmd/rhc/collector_cmd.go
+++ b/cmd/rhc/collector_cmd.go
@@ -1,0 +1,205 @@
+package main
+
+import (
+	"encoding/json"
+	"fmt"
+	"log/slog"
+	"net"
+	"strings"
+
+	"github.com/emersion/go-varlink"
+	"github.com/redhatinsights/rhc/internal/collector"
+	"github.com/redhatinsights/rhc/internal/ui"
+	"github.com/redhatinsights/rhc/varlink/collectorapi"
+	"github.com/urfave/cli/v2"
+)
+
+const rhcServerSocket = "/run/rhc/com.redhat.rhc"
+
+// newCollectorClient creates a varlink client for the collector API.
+func newCollectorClient() (*collectorapi.Client, func(), error) {
+	conn, err := net.Dial("unix", rhcServerSocket)
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to connect to rhc-server: %w", err)
+	}
+	varlinkClient := varlink.NewClient(conn)
+	client := &collectorapi.Client{Client: varlinkClient}
+	cleanup := func() {
+		err = varlinkClient.Close()
+		if err != nil {
+			slog.Debug("failed to close varlink client", "error", err)
+			return
+		}
+	}
+	return client, cleanup, nil
+}
+
+// beforeCollectorInfoAction validates the collector info command arguments and configuration.
+func beforeCollectorInfoAction(ctx *cli.Context) error {
+	return validateCollectorCommand(ctx, true, true)
+}
+
+// collectorInfoAction retrieves and displays information for a specific collector.
+func collectorInfoAction(ctx *cli.Context) error {
+	logCommandStart(ctx)
+	client, cleanup, err := newCollectorClient()
+	if err != nil {
+		return cli.Exit(fmt.Sprintf("failed to connect to rhc-server: %v", err), ExitCodeErr)
+	}
+	defer cleanup()
+	response, err := client.Info(&collectorapi.InfoIn{Id: ctx.Args().First()})
+	if err != nil {
+		return cli.Exit(fmt.Sprintf("failed to get collector info: %v", err), ExitCodeErr)
+	}
+	ui.PrintCollectorInfo(&response.Info)
+	return nil
+}
+
+// beforeCollectorListAction validates the collector list command configuration.
+func beforeCollectorListAction(ctx *cli.Context) error {
+	return validateCollectorCommand(ctx, false, true)
+}
+
+// collectorListAction retrieves and displays a list of all available collectors.
+func collectorListAction(ctx *cli.Context) error {
+	logCommandStart(ctx)
+
+	client, cleanup, err := newCollectorClient()
+	if err != nil {
+		return cli.Exit(fmt.Sprintf("failed to connect to rhc-server: %v", err), ExitCodeErr)
+	}
+	defer cleanup()
+
+	response, err := client.List(&collectorapi.ListIn{})
+	if err != nil {
+		return cli.Exit(fmt.Sprintf("failed to list collectors: %v", err), ExitCodeErr)
+	}
+
+	if ui.IsOutputMachineReadable() {
+		if len(response.Collectors) == 0 {
+			fmt.Println("{}")
+			return nil
+		}
+		jsonData, err := json.Marshal(response.Collectors)
+		if err != nil {
+			return cli.Exit(fmt.Sprintf("failed to marshal collectors: %v", err), ExitCodeErr)
+		}
+		fmt.Println(string(jsonData))
+		return nil
+	}
+
+	var rows [][]string
+	for _, info := range response.Collectors {
+		rows = append(rows, []string{info.Id, info.Name})
+	}
+
+	headers := []string{"ID", "NAME"}
+	ui.PrintTable(headers, rows)
+	return nil
+}
+
+// beforeCollectorTimersAction validates the collector timers command configuration.
+func beforeCollectorTimersAction(ctx *cli.Context) error {
+	return validateCollectorCommand(ctx, false, true)
+}
+
+// collectorTimersAction retrieves and displays timer information for all collectors.
+func collectorTimersAction(ctx *cli.Context) error {
+	logCommandStart(ctx)
+
+	client, cleanup, err := newCollectorClient()
+	if err != nil {
+		return cli.Exit(fmt.Sprintf("failed to connect to rhc-server: %v", err), ExitCodeErr)
+	}
+	defer cleanup()
+
+	response, err := client.List(&collectorapi.ListIn{})
+	if err != nil {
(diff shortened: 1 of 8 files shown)


internal/ui/collector.go
```diff
@@ -0,0 +1,161 @@
...
+		if i == len(cells)-1 {
+			fmt.Printf("%s\n", cell)
+		} else {
+			fmt.Printf("%-*s  ", colWidths[i], cell)
+		}
+	}
+}
+
+// PrintTable prints data in a table format with dynamic column widths.
+// headers are the column headers, rows contains the data for each row.
```
m-horky: nit: `rows contain`

internal/ui/collector.go
```diff
@@ -0,0 +1,161 @@
...
+	if info.LastRun != nil {
+		lastRunTime := time.Unix(int64(*info.LastRun), 0)
+		relativeTime := formatRelativeTime(time.Since(lastRunTime))
+		fmt.Printf("Last run:  %s (%s ago)\n", lastRunTime.Format(timeFormat), relativeTime)
+	} else {
+		fmt.Printf("Last run:  -\n")
+	}
+	if info.NextRun != nil {
+		nextRunTime := time.Unix(int64(*info.NextRun), 0)
+		relativeTime := formatRelativeTime(time.Until(nextRunTime))
```
m-horky, line 122: > `formatRelativeTime` converts negative durations to positive ones, but looking at the usage:
> - `time.Until(nextRunTime)` - negative for overdue scheduled times
>
> For overdue "next run" times, showing a positive duration without indicating it's overdue is misleading. It should probably show something like "overdue by 2h" instead of just "2h"

-- Claude Sonnet 4

Do we want to address this somehow? The subjective likelihood of this feels low, I'm not sure if systemd could _not_ start the execution.

`time.Until(lastRunTime)` would only be relative if the system's clock rolled back, like in a misconfigured NTP situation. I am not sure if we want to deal with that use-case either, making an assumption that "time always goes forward" does not seem insane to me.

cc @pholica

internal/ui/collector.go
```diff
@@ -0,0 +1,161 @@
...
+			fmt.Printf("%s\n", cell)
+		} else {
+			fmt.Printf("%-*s  ", colWidths[i], cell)
+		}
+	}
+}
+
+// PrintTable prints data in a table format with dynamic column widths.
+// headers are the column headers, rows contains the data for each row.
+func PrintTable(headers []string, rows [][]string) {
```
m-horky, line 47: We already use `text/tabwriter` to display tables. Does this approach have any significant benefits? I know `test/tabwriter` doesn't handle UTF-8 well, some characters take two graphical characters to draw.

There was an attempt to solve this in #327, but we never got it into mergable state.

internal/collector/collector.go
```diff
@@ -346,3 +348,26 @@ func newConfig(id string, dto *configDto) (Config, error) {
 		ContentType:        dto.Ingress.ContentType,
 	}, nil
 }
+
+// ValidateCollectorAndConnect validates that the collector exists and systemd is available,
+// then establishes a systemd connection for timer operations.
+// Returns the systemd connection, timer unit name, and any validation errors.
+func ValidateCollectorAndConnect(collectorID string) (*systemd.Conn, string, error) {
+	if !systemd.IsSystemdAvailable() {
+		return nil, "", fmt.Errorf("automatic execution requires systemd, which is not present in your environment, please use 'rhc collector run' instead")
```
m-horky: You aren't specifying `run` subcommand in the cmd/ code. I'm not sure we need and want it, can we simplify this error message, and keep just the "automatic execution requires systemd" part?

cmd/rhc/collector_cmd.go
```diff
@@ -0,0 +1,207 @@
...
+	collectorId := ctx.Args().First()
+	nowFlag := ctx.Bool("now")
+
+	conn, timerName, err := collector.ValidateCollectorAndConnect(collectorId)
+	if err != nil {
+		return cli.Exit(fmt.Sprintf("%v", err), 1)
+	}
+	defer conn.Close()
+
+	err = conn.EnableUnit(timerName, nowFlag, false)
```
m-horky: Here, we're enabling the timer. Should we pass `true` instead of `nowFlag`? Now the timer will only get enabled after reboot.

The `--now` flag is supposed to start the service immediately, not the timer, right?

m-horky (changes requested):
I have not discovered any significant issues, overall the code feels solid and considering the state of the codebase well structured.

UX notes:

- With no data collector, `list` outputs `failed to list collectors: EOF`. Could we display some better message, like "Error: No data collectors are available"?
- `rhc collector timers` does not display the value of NEXT, even when the timer is enabled. It may be because we aren't passing `--all` flag to `systemctl list-timers`.

More inline.

internal/ui/collector.go
pkoprda: I tried a working approach for your suggestion, but in my eyes it was a bit overengineering. I get your point, but I would rather get some user feedback if the users are okay with the current approach and if not we can change that.

internal/ui/collector.go
pkoprda: Right, thanks for the suggestion, fixed

internal/collector/collector.go
pkoprda: `run` subcommand is not part of the collector, it was copy-pasted from the design document. I fixed the message so it doesn't mention `run` subcommand.

cmd/rhc/collector_cmd.go
```diff
@@ -0,0 +1,232 @@
...
+		if err != nil {
+			slog.Debug("failed to close varlink client", "error", err)
+			return
+		}
+	}
+	return client, cleanup, nil
+}
+
+// beforeCollectorInfoAction validates the collector info command arguments and configuration.
+func beforeCollectorInfoAction(ctx *cli.Context) error {
```
mjcr99, line 38: All `beforeCollector*Action()` functions follow very similar patterns. Is there any way to extract the common logic and reuse it?

internal/ui/collector.go
m-horky: Can you create a Jira task for addressing or ignoring this? We could validate with PM outside of this pull request, and it would stay somewhere in the history.

internal/ui/collector.go
```diff
@@ -0,0 +1,164 @@
...
+
+	fmt.Printf("Config:   %s\n", info.ConfigPath)
+	fmt.Printf("Service:  %s\n", info.ServiceName)
+	fmt.Printf("Timer:    %s\n", info.TimerName)
+}
+
+// PrintCollectorTimers formats multiple CollectorInfo structs into a table showing timing information.
+// For machine-readable output, it returns JSON array.
+// For human-readable output, it prints a table.
+func PrintCollectorTimers(infos []*collectorapi.CollectorInfo) {
```
m-horky, line 136: I'm not 100% happy with this solution, I'd have preferred if JSON printing happened elsewhere, but I don't want to block the PR on this. Consider it a nit.

m-horky (dismissed):
Nothing stands out, you have my ACK. I'm leaving the resolution of the rest of my comments up to you.

mjcr99 (commented):
Thanks for the PR @pkoprda! LGTM, I have a nit suggestion.

internal/ui/collector.go
pkoprda: Sure, created CCT-2382

internal/ui/collector.go
pkoprda: Addressed, I added a method `printMachineReadable()` which marshals data to JSON and then prints it to stdout.

cmd/rhc/collector_cmd.go
pkoprda: Addressed in https://github.com/RedHatInsights/rhc/pull/425/changes/35ccf3c8703ca26256b46c04520c325b6c490645

mjcr99 (approved):
LGTM

PR #426, opened 2026-04-29
fix(test): do not treat dependency skips as fatal errors

pkoprda:
When features are explicitly disabled with --disable-feature, dependent features are skipped but this should not cause the connect command to fail. This commit fixes the test_log_connect_with_disabled_feature test which expects connect to succeed when content is disabled.


m-horky (approved):
ACK

PR #427, opened 2026-05-05
tests: clean up rhc worker pid files after each integration test

Archana-PandeyM:
**What-**

- Add a per-test teardown hook in integration-tests/conftest.py to clean stale worker PID files under /var/run/rhc/workers.
- Remove all *.pid files after each test run instead of targeting only one specific PID file.

**Why-**
- Some tests can leave stale worker PID files, which may affect subsequent tests and create flaky behavior(known issue on yggdrasil 0.2.6).
This change improves test isolation and makes integration runs more reliable.

**Ref - CCT-1066** - A fix was provided in yggdrasil -[0.2.7.2](https://github.com/RedHatInsights/yggdrasil/releases/tag/0.2.7.2)

### DIFF
--- a/integration-tests/conftest.py
+++ b/integration-tests/conftest.py
@@ -3,8 +3,10 @@ import subprocess
 import logging
 import distro
 import os
+import time
 
 logger = logging.getLogger(__name__)
+WORKER_PID_DIR = "/var/run/rhc/workers"
 
 # default values so import-time access does not crash
 pytest.rhel_version_tuple = (0, 0)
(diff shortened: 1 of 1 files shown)


integration-tests/conftest.py
```diff
@@ -35,6 +37,38 @@ def pytest_configure(config):
...
+        for name in names:
+            if not name.endswith(".pid"):
+                continue
+            pid_path = os.path.join(WORKER_PID_DIR, name)
+            try:
+                os.remove(pid_path)
+                removed_any = True
+            except FileNotFoundError:
+                # Another process removed it between listdir and remove.
+                pass
```
pkoprda: Can you add here some debug log?

pkoprda (changes requested):
Looks good, I have just one request below.

Lint CI job is failing because the commit does not follow conventional commit specification. Prefix should be `test:` not `tests:`. Also could you add an explanation why are you doing this change to the commit description? Thanks!

Archana-PandeyM:
content template tests failed due to full deprecation of basic auth, I will move to token based auth in another PR.

pkoprda (approved):
LGTM

zpetrace (approved):
Yeah, LGTM, thanks @Archana-PandeyM !

PR #428, opened 2026-05-06
refactor(version): centralize version handling in `pkg/version`

mjcr99:
* Card ID: CCT-2392

## Description

Hi team,

This Pr moves the version variable from cmd/rhc, cmd/rhc-server, and cmd/rhc-collector into a shared `pkg/version` package. This establishes the pkg/ business layer as part of the ADR-011 three-layer architecture transformation.

### DIFF
--- a/pkg/version/version.go
+++ b/pkg/version/version.go
@@ -1,0 +1,5 @@
+package version
+
+// Version is the version string, set at build time via ldflags.
+// Example: go build -ldflags "-X github.com/redhatinsights/rhc/pkg/version.Version=1.2.3"
+var Version = "dev"
--- a/Makefile
+++ b/Makefile
@@ -2,13 +2,15 @@
 .SHELLFLAGS := -e -c
 
 VERSION := $(shell rpmspec rhc.spec --query --queryformat '%{version}')
+LDFLAGS := -ldflags "-X github.com/redhatinsights/rhc/pkg/version.Version=$(VERSION)"
+GO_BUILD := go build $(LDFLAGS)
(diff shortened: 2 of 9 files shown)


pkg/version/version.go
```diff
@@ -0,0 +1,10 @@
+package version
+
+// Version is the version string, set at build time via ldflags.
+// Example: go build -ldflags "-X github.com/redhatinsights/rhc/pkg/version.Version=1.2.3"
+var Version = "dev"
+
+// Get returns the current version string.
+func Get() string {
+	return Version
+}
```
m-horky: This should either be `version` (unexported) with a `Get()` method, or `Version` (exported) with nothing else. This creates a weird duplication.

I'm in favor of dropping the `Get()`, personally.

m-horky (changes requested):
Could you update both the .spec files, please? Before we drop the old one, we should keep this consistent.

pkg/version/version.go
mjcr99: Yep, you are right, addressed!

m-horky (approved):
LGTM

RHEL tests seem to be unrelated:

```
E           requests.exceptions.HTTPError: 401 Client Error: Unauthorized for url: https://console.redhat.com/api/content-sources/v1.0//templates/
=========================== short test summary info ============================
FAILED integration-tests/test_connect.py::test_connect_with_content_template[basic]
FAILED integration-tests/test_connect.py::test_connect_with_content_template[activation-key]
```

PR #430, opened 2026-05-07
chore: migrate urfave cli to v3

Peto-RH:
* Card ID: CCT-2122

Update rhc to use urfave/cli v3 while preserving existing command behavior.


Peto-RH:
Note for reviewer: I removed yaml license from `go-vendor-tools.toml` as `cli/v3` no longer needs that yaml dependency

cmd/rhc/main.go
```diff
@@ -398,14 +401,14 @@ func main() {
...
-	app.BashComplete = BashComplete
+	app.EnableShellCompletion = true
+	app.ShellComplete = ShellComplete
 	app.Action = mainAction
 	app.Before = beforeAction
 	app.After = afterAction
 	app.ExitErrHandler = exitErrHandler
 
-	if err := app.Run(os.Args); err != nil {
+	if err := app.Run(context.Background(), os.Args); err != nil {
```
mjcr99: Nit:

With this approach, if a user presses `Ctrl+C` during registration, they cannot currently cancel, and the process will continue. I'd like to consider signal handling, so context cancellation would propagate and allow cleanup.

Maybe something like:
```go
  import (
      "os/signal" 
      "syscall"
  )  

  ctx, stop := signal.NotifyContext(context.Background(), os.Interrupt, syscall.SIGTERM)
  defer stop() 
  
  if err := app.Run(ctx, os.Args); err != nil {
      slog.Error(err.Error())
  }
```

cmd/rhc/main.go
```diff
@@ -43,61 +46,49 @@ func mainAction(c *cli.Context) error {
...
 
 // beforeAction is triggered before other actions are triggered
-func beforeAction(c *cli.Context) error {
+func beforeAction(ctx context.Context, cmd *cli.Command) (context.Context, error) {
 	// check if --log-level was set via command line
 	var logLevelSrc string
-	if c.IsSet(cliLogLevel) {
+	if cmd.IsSet(cliLogLevel) {
 		logLevelSrc = "command line"
 	}
```
mjcr99, line 76: Before these changes, we were verifying the config file in the `beforeaction` section. Should we consider it?

mjcr99 (changes requested):
GJ @Peto-RH, I leave you a couple of comments.
I would also consider referring to the Card ID in the second commit.

cmd/rhc/main.go
Peto-RH: Hi, good question. I was also brainstorming about this. 
The 2 functions `NewTomlSourceFromFile` and `ApplyInputSourceValues` do not exist anymore in `v3`, and `v3` unfortunately does not have a 1-to-1 solution for this. There are basically 2 problems:
1. How do we keep validating the config file before every command action runs? (this is the one you are probably asking about)
2. How do we propagate errors and return an error code upon parsing failure?

There are multiple approaches:
1. Do it the `v3` idiomatic way. This is the approach I currently proposed with the new commit. 
  - That means - do not validate the config file in beforeAction (validate only in `main()`) 
  - Do not report or catch parsing errors.
2. Set the `URFAVE_CLI_TRACING=on` env variable (for example in the .spec file). With this variable, the logs will show that the parsing failed, but this will not affect the program.
3. Add some custom functionality to validate the toml. I am doubtful about this one because it would require some 3rd party lib to be imported. This would probably look something like this.
```
import "github.com/pelletier/go-toml"

func beforeAction(ctx context.Context, cmd *cli.Command) (context.Context, error) {
    // some code ...
    configPath := cmd.String("config")
    if configPath != "" {
    if _, err := os.Stat(configPath); err == nil {
	    // File exists, validate it's parseable TOML
	    if _, err := toml.LoadFile(configPath); err != nil {
		    return ctx, fmt.Errorf("invalid config file %s: %w", configPath, err)
	    }
    }
    // some more code ...
}
```

What do you think should be our approach to this? Do you have any other ideas?

FYI, the only toml-connected function in `v3` is the `TOML` function that we are using in the new version, for example here [`altsrctoml.TOML(cliCertFile, configSource)`](https://github.com/RedHatInsights/rhc/pull/430/changes#:~:text=altsrctoml.TOML(cliKeyFile%2C%20configSource)%2C). 
[Here is the documentation for toml in v3](https://pkg.go.dev/github.com/urfave/cli-altsrc/v3@v3.1.0/toml)

pkoprda (changes requested):
Please resolve conflicts

cmd/rhc/main.go
mjcr99: Hi @Peto-RH,
In my opinion, option 3 is the best approach, it would keep the consistency with the current implementation,  failing fast before command execution and providing clear error messages.

pkoprda (approved):
LGTM, thanks!

mjcr99 (approved):

PR #436, opened 2026-05-12
test: Switch from basic auth to token based auth

Archana-PandeyM:
The Red Hat Console content-sources templates API no longer works with HTTP basic auth. The test_connect_with_content_template flow prefetches expected repository names from that API using basic auth.

This work updates the integration harness to use token based auth(via service account)
Obtain an access token from the SSO token endpoint, then call the templates API with Authorization: Bearer.

### DIFF
--- a/integration-tests/test_connect.py
+++ b/integration-tests/test_connect.py
@@ -8,21 +8,22 @@
 
 import contextlib
 import json
-import time
 import logging
+import time
+
 import pytest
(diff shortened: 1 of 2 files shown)


m-horky (approved):
It seems we are requesting new bearer token every time, instead of reusing it. It makes the code simpler, I don't think I have any concerns. ACK

zpetrace (approved):
LGTM, thanks!

PR #439, opened 2026-05-14
feat(operations): add Feature enum and FeatureStatus operation

mjcr99:
## Description

This PR creates pkg/operations/ with typed Feature enum and FeatureStatus function implementing dependency injection via FeatureChecker interface.

### DIFF
--- a/pkg/operations/feature.go
+++ b/pkg/operations/feature.go
@@ -1,0 +1,52 @@
+package operations
+
+import "fmt"
+
+// Feature represents a system feature that can be enabled or disabled.
+type Feature int
+
+const (
+	// Analytics represents Red Hat Lightspeed data collection.
+	Analytics Feature = iota
+	// Content represents Red Hat content management.
+	Content
+	// RemoteManagement represents Red Hat Lightspeed remote management.
+	RemoteManagement
+)
+
+// FeatureOperationOptions contains options for feature operations.
+// Currently holds only the Feature identifier, but can be extended
+// in the future to include additional operation-specific parameters.
+type FeatureOperationOptions struct {
+	Feature Feature
+}
+
+// String returns the string representation of the feature.
+// The returned string matches the feature ID used in commands and configuration.
+func (f Feature) String() string {
+	switch f {
+	case Analytics:
+		return "analytics"
+	case Content:
+		return "content"
+	case RemoteManagement:
+		return "remote-management"
+	default:
+		return fmt.Sprintf("unknown(%d)", int(f))
+	}
+}
+
+// ParseFeature converts a string feature name to a Feature enum value.
+// Returns an error if the string does not match any known feature.
+func ParseFeature(s string) (Feature, error) {
+	switch s {
+	case "analytics":
+		return Analytics, nil
+	case "content":
+		return Content, nil
+	case "remote-management":
+		return RemoteManagement, nil
+	default:
+		return 0, fmt.Errorf("unknown feature: %s", s)
+	}
+}
--- a/pkg/operations/feature_status.go
+++ b/pkg/operations/feature_status.go
@@ -1,0 +1,37 @@
+package operations
+
+import (
+	"fmt"
+
+	"github.com/redhatinsights/rhc/internal/datacollection"
+	"github.com/redhatinsights/rhc/internal/remotemanagement"
+	"github.com/redhatinsights/rhc/internal/rhsm"
+)
+
+// FeatureStatusResult represents the result of checking a feature's status.
+type FeatureStatusResult struct {
+	Enabled bool
+	Err     error
+}
+
+// FeatureStatus checks the enabled status of a feature by calling the
+// appropriate internal/* infrastructure function based on the feature type.
+// This function provides the required layer boundary between cmd/ and internal/*
+// as cmd/ cannot import internal/* directly per ADR-011 layering rules.
+func FeatureStatus(opts FeatureOperationOptions) FeatureStatusResult {
+	var enabled bool
+	var err error
+
+	switch opts.Feature {
+	case Analytics:
+		enabled, err = datacollection.InsightsClientIsRegistered()
+	case Content:
+		enabled, err = rhsm.IsContentManagementEnabled()
+	case RemoteManagement:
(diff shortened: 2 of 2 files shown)


pkg/operations/feature_status.go
```diff
@@ -0,0 +1,64 @@
...
+type FeatureStatusDeps struct {
+	Analytics        FeatureChecker // Checks analytics (insights-client) registration status
+	Content          FeatureChecker // Checks content management (rhsm) enabled status
+	RemoteManagement FeatureChecker // Checks remote management (yggdrasil) service status
+}
+
+// FeatureStatus checks the enabled status of the given features.
+// It returns a slice of results, one for each requested feature.
+// Dependencies are injected via the deps parameter.
+func FeatureStatus(features []Feature, deps FeatureStatusDeps) []FeatureStatusResult {
```
m-horky: This method is mixing singular (`FeatureStatus`) with plural (`[]`). Which of these options do we expect to need?

Personally, I have a slight preference for singular, but I could be convinced otherwise. We could drop the `FeatureStatusDeps` for that, and we could rely on a `checker FeatureChecker` passed in. But I can be convinced otherwise.

m-horky (changes requested):
Could you expand the commit message a bit? Imagine someone five years from now looking back, would they understand what is the reason we are adding the code like this?

I have one code-related note, otherwise it is looking good.

pkg/operations/feature_status.go
mjcr99: You are right, it is not very consistent. I agree that the singular approach is more convenient and also simplifies the implementation.

pkg/operations/feature_status.go
```diff
@@ -0,0 +1,34 @@
...
+type FeatureStatusResult struct {
+	Feature Feature
+	Enabled bool
+	Err     error
+}
+
+// FeatureStatus checks the enabled status of a single feature.
+// It accepts a feature and its corresponding checker via dependency injection.
+// Returns a result containing the feature, its enabled state, and any error.
+func FeatureStatus(feature Feature, checker FeatureChecker) FeatureStatusResult {
```
m-horky: That `feature` argument is only a passthrough now. Do we need to have the field in the `FeatureStatusResult`? How about

```go
type FeatureStatusResult struct {Enabled: bool, Error: error}
```

Since this is a layer API, do we want to wrap the input with an `Options`?

```go
type FeatureStatusOptions struct {Feature: Feature}
func FeatureStatus(opts FeatureStatusOptions) FeatureStatusResult {}
```

Without injections, we may query for sub-man, insights-client and yggdrasil directly. Let's match on the `opts.Feature` 🤔

pkg/operations/feature_status.go
```diff
@@ -0,0 +1,34 @@
...
+	result := FeatureStatusResult{
+		Feature: feature,
+	}
+
+	if checker == nil {
+		result.Err = fmt.Errorf("feature checker not provided for %s", feature)
+		return result
+	}
```
m-horky: It would be easy to blame this on AI, but I didn't manage to review the proposal correctly. Let's not do injections on this layer, we shouldn't need checker at all. Let's keep injections for `internal/` packages (maybe D-Bus could be testable with those).

pkg/operations/feature_status_test.go
m-horky, line 1: With the changes I am proposing, we might not need any unit tests. It's up to you.

m-horky (changes requested):
Simplify all the things! :)

pkg/operations/feature_status_test.go
mjcr99: So far, I have removed the tests.

pkg/operations/feature_status.go
mjcr99: I created this extendible type:
```go
type FeatureOperationOptions struct {
	Feature Feature
}
```
In the `rhc/pkg/operations/feature.go` file, as a DTO for all the operations we want to implement related to the features.

pkg/operations/feature.go
```diff
@@ -0,0 +1,52 @@
...
+	// Content represents Red Hat content management.
+	Content
+	// RemoteManagement represents Red Hat Lightspeed remote management.
+	RemoteManagement
+)
+
+// FeatureOperationOptions contains options for feature operations.
+// Currently holds only the Feature identifier, but can be extended
+// in the future to include additional operation-specific parameters.
+type FeatureOperationOptions struct {
```
m-horky, line 20: Could we keep the location of `*Options` and `*Result` next to each other? There's nothing wrong with this, but I would expect this object just above `FeatureStatusResult`.

m-horky (changes requested):
I like the shape of it a lot!

pkg/operations/feature.go
mjcr99: I placed it in the `pkg/operations/feature.go` because I think this `*Options` type, would be used by almost all the operations, but the `FeatureStatusResult` would be only handled by the `pkg/operations/feature_status.go` file.

m-horky: Fair. Let's go with it for now.

m-horky (approved):

pkg/operations/feature.go
```diff
@@ -0,0 +1,52 @@
...
+type Feature int
+
+const (
+	// Analytics represents Red Hat Lightspeed data collection.
+	Analytics Feature = iota
+	// Content represents Red Hat content management.
+	Content
+	// RemoteManagement represents Red Hat Lightspeed remote management.
+	RemoteManagement
+)
```
m-horky, line 15: Nit: Should this exist in `pkg/feature/` or `internal/feature/` instead of here? I am not sure this enum should be in operations.

My ACK on this PR is staying valid.

mjcr99: Hmm, good question, for me, both places are valid. `pkg/feature/` makes sense to me because it is going to be used in the feature folder, but `internal/feature/` is also a valid place because the type represents values extracted directly from the system.

I would rather leave it as it is for simplicity so far.

m-horky: Sounds good.

PR #442, opened 2026-05-15
fix(collector): Improve warning message invalid collector config

stephenbourke:
* Card ID: CCT-2239

Improves the warning message when unsupported collector configuration feature values are provided.

- Update warning level log message to debug level and made message clearer when an unsupported collector configuration feature value is used

### DIFF
--- a/internal/collector/collector.go
+++ b/internal/collector/collector.go
@@ -334,15 +334,15 @@ func newConfig(id string, dto *configDto) (Config, error) {
 		return Config{}, fmt.Errorf("invalid config: ingress.content_type is required")
 	}
 
-	// Emit warning if meta.feature is present but not 'analytics'
+	// Emit debug log message if meta.feature is present but not 'analytics'
 	if dto.Meta.Feature != nil && *dto.Meta.Feature != "analytics" {
-		slog.Warn("Unexpected meta.feature value", "actual", *dto.Meta.Feature, "expected", "analytics")
+		slog.Debug("Unsupported collector config feature value ignored", "feature", *dto.Meta.Feature, "supported", "analytics")
 	}
 
 	return Config{
 		ID:                 id,
 		Name:               dto.Meta.Name,
-		IsAnalyticsFeature: dto.Meta.Feature == nil || *dto.Meta.Feature == "analytics",
+		IsAnalyticsFeature: dto.Meta.Feature != nil && *dto.Meta.Feature == "analytics",
 		User:               user,
 		Group:              group,
 		ContentType:        dto.Ingress.ContentType,
--- a/internal/collector/collector_test.go
+++ b/internal/collector/collector_test.go
@@ -112,7 +112,7 @@ func TestNewConfig(t *testing.T) {
 			want: Config{
 				ID:                 "test.nil.feature",
 				Name:               "Test nil feature",
-				IsAnalyticsFeature: true,
+				IsAnalyticsFeature: false,
 				User:               "root",
 				Group:              "root",
 				ContentType:        "application/vnd.redhat.advisor.collection",


internal/collector/collector.go
```diff
@@ -334,15 +337,15 @@ func newConfig(id string, dto *configDto) (Config, error) {
...
+	// Emit warning if meta.feature is present but not a supported feature
+	if dto.Meta.Feature != nil && !slices.Contains(supportedFeatures, *dto.Meta.Feature) {
+		slog.Warn("Unsupported collector config feature value ignored", "feature", *dto.Meta.Feature, "supported", strings.Join(supportedFeatures, ", "))
 	}
 
 	return Config{
 		ID:                 id,
 		Name:               dto.Meta.Name,
-		IsAnalyticsFeature: dto.Meta.Feature == nil || *dto.Meta.Feature == "analytics",
+		IsSupportedFeature: dto.Meta.Feature == nil || slices.Contains(supportedFeatures, *dto.Meta.Feature),
```
m-horky: It feels like the original line was buggy: `dto.Meta.Feature == nil` would return `true`, evaluating as "this data collector implements the analytics feature", which is incorrect. We don't have any collectors to verify the behavior against, I'm not surprised it wasn't caught yet.

internal/collector/collector.go
```diff
@@ -34,8 +35,8 @@ type Config struct {
 	ID string
 	// Name is the human-readable name of the collector.
 	Name string
-	// IsAnalyticsFeature indicates whether collector provides analytics functionality.
-	IsAnalyticsFeature bool
+	// IsSupportedFeature indicates whether collector provides supported functionality, currently only analytics.
+	IsSupportedFeature bool
 	// User specifies the system user under which the collector should run.
```
m-horky, line 39: I don't think this is what we'll need long-term. The `IsAnalyticsFeature` is used to gate the enablement of analytics (see the `cmd/rhc-server/collector.go`), with this new logic any supported feature would trigger that.

We can either make this field carry a `string` (or perhaps a `Feature` once #439 gets merged), or we could keep this named `Analytics` and only change the logging.

internal/collector/collector.go
```diff
@@ -34,8 +35,8 @@ type Config struct {
 	ID string
 	// Name is the human-readable name of the collector.
 	Name string
-	// IsAnalyticsFeature indicates whether collector provides analytics functionality.
-	IsAnalyticsFeature bool
+	// IsSupportedFeature indicates whether collector provides supported functionality, currently only analytics.
+	IsSupportedFeature bool
```
m-horky: This value is used to get the enablement of `analytics` feature, this would enable it every time the config contained a valid feature name.

We could make this either a `string` (or a `Feature` once #439 is merged), or we could keep this a bool and limit the change to only cover the validation and logging.

internal/collector/collector.go
```diff
@@ -334,15 +337,15 @@ func newConfig(id string, dto *configDto) (Config, error) {
 		return Config{}, fmt.Errorf("invalid config: ingress.content_type is required")
 	}
 
-	// Emit warning if meta.feature is present but not 'analytics'
-	if dto.Meta.Feature != nil && *dto.Meta.Feature != "analytics" {
-		slog.Warn("Unexpected meta.feature value", "actual", *dto.Meta.Feature, "expected", "analytics")
+	// Emit warning if meta.feature is present but not a supported feature
+	if dto.Meta.Feature != nil && !slices.Contains(supportedFeatures, *dto.Meta.Feature) {
+		slog.Warn("Unsupported collector config feature value ignored", "feature", *dto.Meta.Feature, "supported", strings.Join(supportedFeatures, ", "))
```
m-horky: We should drop this to a `.Debug()`. This log message isn't actionable to the user, invalid feature name is a bug in the data collector, the user can't fix this.

m-horky (dismissed):
Thanks for the patch, I have a few notes.

Could you please drop that merge commit from this PR, and rebase instead?

m-horky (commented):
Looks good! Can you please squash those commits? We like our git history clean. LGTM otherwise.

m-horky (approved):
LGTM

PR #444, opened 2026-05-25
chore: Remove useless CLI option from unit file

jirihnidek:
* The rhc-server ignores all CLI options. It is useless to try to run rhc-server with some CLI options in systemd unit file


m-horky (approved):
ACK, thanks!

PR #450, opened 2026-05-26
fix: Implement JSON format output for configure status command

stephenbourke:
Card ID : CCT-2460

   - Implement JSON format output support for configure status command
   - Add JSON and tablular print util functions moved to internal/ui package
   - Enable and amend integration tests

### DIFF
--- a/cmd/rhc/configure_features_cmd.go
+++ b/cmd/rhc/configure_features_cmd.go
@@ -4,16 +4,45 @@ import (
 	"context"
 	"fmt"
 	"log/slog"
-	"os"
-	"text/tabwriter"
 
 	"github.com/urfave/cli/v3"
 
 	"github.com/redhatinsights/rhc/internal/rhsm"
+	"github.com/redhatinsights/rhc/internal/ui"
 	"github.com/redhatinsights/rhc/pkg/exitcode"
 	"github.com/redhatinsights/rhc/pkg/feature"
 	"github.com/redhatinsights/rhc/pkg/feature/prefcache"
 )
+
+type ConfigureFeatureStatus struct {
+	Preference  string `json:"preference,omitempty"`
+	Enabled     *bool  `json:"enabled,omitempty"`
+	Description string `json:"description,omitempty"`
+	Error       string `json:"error,omitempty"`
+}
+
+type ConfigureFeaturesStatus struct {
+	Connected bool `json:"connected"`
+	Features  struct {
+		Content          ConfigureFeatureStatus `json:"content"`
+		Analytics        ConfigureFeatureStatus `json:"analytics"`
+		RemoteManagement ConfigureFeatureStatus `json:"remote_management"`
+	} `json:"features"`
+	returnCode int
+}
+
+func (status *ConfigureFeaturesStatus) setFeatureResult(featureID string, result ConfigureFeatureStatus) {
+	switch featureID {
+	case "content":
+		status.Features.Content = result
+	case "analytics":
+		status.Features.Analytics = result
+	case "remote-management":
+		status.Features.RemoteManagement = result
+	default:
+		slog.Warn("unknown feature id for configure features status", "id", featureID)
+	}
+}
 
 // TODO Handle machine-readable output by always returning a DTO from business logic;
 //  a place here should only act as a presentation layer.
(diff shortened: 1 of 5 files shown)


TESTING.md
```diff
@@ -19,7 +19,7 @@ You will have to install required RPMs and Python packages:
 
 ```
 dnf -y install python3-pip python3-pytest tmt+all
-sudo pip -r install integration-tests/requirements.txt
+sudo pip install -r integration-tests/requirements.txt
```
pkoprda: Why this change?

integration-tests/test_configure_feature.py
```diff
@@ -46,9 +42,15 @@ def _assert_configure_features_status_json_shape(data: dict):
...
-        assert isinstance(v["enabled"], bool)
+    if data["connected"]:
+        for _k, v in feats.items():
+            assert set(v.keys()) <= {"enabled", "error"}
+            if "error" not in v:
+                assert isinstance(v["enabled"], bool)
+    else:
+        for _k, v in feats.items():
+            assert set(v.keys()) == {"description", "preference"}
+            assert v["preference"] in ("enable", "skip")
```
pkoprda, line 63: The test `test_configure_features_status_json_when_connected` is failing because of this change. Please, fix it

cmd/rhc/status_cmd.go
```diff
@@ -221,14 +220,6 @@ type SystemStatus struct {
 
 // printJSONStatus tries to print the system status as JSON to stdout.
 // When marshaling of systemStatus fails, then error is returned
-func printJSONStatus(systemStatus *SystemStatus) error {
```
pkoprda, line 224: You removed the method, but you didn't remove this comment

cmd/rhc/util.go
```diff
@@ -129,3 +130,12 @@ func validateCollectorCommand(ctx *cli.Context, requiresCollectorID, requiresFor
 	configureUI(ctx)
 	return nil
 }
+
+func PrintJSON(v any) error {
```
pkoprda: Missing docstring for this method

pkoprda: Also food for thought: wouldn't it make more sense to have this method in `ui` package?

internal/ui/ui.go
```diff
@@ -104,6 +116,19 @@ func Printf(
 	fmt.Printf(format, a...)
 }
 
+// FPrintf acts as a no-op if the output is machine-readable.
```
pkoprda:
```suggestion
// Fprintf acts as a no-op if the output is machine-readable.
```

cmd/rhc/configure_features_cmd.go
```diff
@@ -45,15 +76,46 @@ func featuresStatusAction(ctx *cli.Context) error {
...
+				err = cli.Exit("", exitcode.Err)
+			}
+		}(&configureFeaturesStatus)
+	}
+
+	ui.Println("Not connected to Red Hat.")
+	ui.Println("")
 	w := tabwriter.NewWriter(os.Stdout, 0, 0, 2, ' ', 0)
-	_, _ = fmt.Fprintf(w, "FEATURE\tPREFERENCE\tDESCRIPTION\n")
+	ui.Fprintf(w, "FEATURE\tPREFERENCE\tDESCRIPTION\n")
```
pkoprda: I don't think this is a good approach. If `ui.Fprintf()` does an early return (`ui.IsOutputMachineReadable()` is true) and writes nothing to the `tabwriter`, then at the bottom of this method `w.Flush()` is still called.

cmd/rhc/configure_features_cmd.go
```diff
@@ -63,26 +125,78 @@ func featuresStatusActionNotRegistered(_ *cli.Context) error {
...
+				err = cli.Exit("", exitcode.Err)
+			}
+		}(&configureFeaturesStatus)
+	}
+
+	ui.Println("Connected to Red Hat.")
+	ui.Println("")
 	w := tabwriter.NewWriter(os.Stdout, 0, 0, 2, ' ', 0)
-	_, _ = fmt.Fprintf(w, "FEATURE\tSTATE\tDESCRIPTION\n")
+	ui.Fprintf(w, "FEATURE\tSTATE\tDESCRIPTION\n")
```
pkoprda: Ditto

cmd/rhc/configure_features_cmd.go
```diff
@@ -63,26 +125,78 @@ func featuresStatusActionNotRegistered(_ *cli.Context) error {
...
+	format := ctx.String("format")
+	switch format {
+	case "json":
+		machineReadablePrintFunc = PrintJSON
+	default:
+		break
+	}
+
+	if ui.IsOutputMachineReadable() {
+		defer func(configureFeaturesStatus *ConfigureFeaturesStatus) {
```
pkoprda: I don't feel good about this `defer func()`. Seems to me like an anti pattern in go, but I don't know how to improve it :(.

Shouldn't have it some named return?

pkoprda (changes requested):
Thanks for the change, I had left some comments bellow.

TESTING.md
Archana-PandeyM: -r is an option for install, not for pip itself.  :+1: 
can be fixed in another PR

pkoprda: Sure, can you just create a separate commit for it in this PR? No need to create a new PR

cmd/rhc/configure_features_cmd.go
stephenbourke: I think you're right on this. It's better to have the writer instantiated and flushed only when text output is required.

cmd/rhc/configure_features_cmd.go
stephenbourke: I don't really like it either. I didn't want to do too much of a rewrite on the logic already there, but it may need a rethink.

stephenbourke: In the latest push, I've removed the defer and hopefully it's a slightly easier to read implementation.

TESTING.md
stephenbourke: This file needs a few changes to it. So perhaps a separate PR will be better.

pkoprda (changes requested):
Looking good, please, resolve conflicts.

pkoprda (approved):
LGTM, thanks!

PR #455, opened 2026-05-28
fix(collector): show human friendly output when no collector is found

lprchal:
* Card ID: CCT-2531


pkoprda (approved):
Hi :) thanks for the PR, approved!

PR #456, opened 2026-05-28
fix: Start rhc-server.socket, when rhc package is installed

jirihnidek:
* Not sure about it, because it break some policy
  * The service should not be started, when rpm is installed as part of installation script, but we will only instruct systemd to start to listen on one another socket.
* The `rhc-server.socket` has to be "started" to be able to call varlink API using given `/run/rhc/com.redhat.rhc` socket. The enabling the systemd `rhc-server.socket` is not enough despite the socket file `/run/rhc/com.redhat.rhc` is created by systemd. It seems that there is no macro for this.

### DIFF
--- a/rhc.spec
+++ b/rhc.spec
@@ -96,6 +96,9 @@ install -m 0644 -vp %{buildroot}%{_unitdir}/yggdrasil.service.d/rhcd.conf %{buil
 %post
 %systemd_post rhc-canonical-facts.timer
 %systemd_post rhc-server.socket
+if [ "$1" -eq 1 ] ; then
+    systemctl start rhc-server.socket >/dev/null 2>&1 || :
+fi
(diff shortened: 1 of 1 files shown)


jirihnidek:
https://fedoraproject.org/wiki/Packaging:Systemd#Why_don't_we.... says:

> Installations can be in changeroots, in an installer context, or in other situations where you don't want the services autostarted.

m-horky (commented):
I'm not sure we _need_ this... rhc comes pre-installed on recent RHEL systems, so when you boot it up, it should already be present. On Fedora, where it is optional we might want to consider this... or we can improve the error message and say "Hey, you need to run systemctl before this".

jirihnidek:
Closing, because it breaks rules.

PR #459, opened 2026-06-01
test: move collector integration tests to tier2

Archana-PandeyM:
Reclassify the Collector API and Socket Activation integration tests from Tier 1 to Tier 2.

Rationale: Only Tier 1 jobs are used for gating. Features that are currently untested, unstable, or known to be flaky should not be classified as Tier 1.

We might do a reclassification effort afterwards for all existing tests. This PR is solely for collector tests and rhc collector is in initial implementation phase.


pkoprda (approved):
ACK

PR #460, opened 2026-06-01
fix: show helpful error when rhc-server.socket is not available

pkoprda:
* Card ID: CCT-2530

When rhc collector commands fail to connect to the Unix socket, provide clear instructions to restart the socket unit instead of a generic connection error.

### DIFF
--- a/cmd/rhc/collector_cmd.go
+++ b/cmd/rhc/collector_cmd.go
@@ -3,10 +3,13 @@ package main
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"log/slog"
 	"net"
+	"os"
 	"strings"
+	"syscall"
 
 	"github.com/emersion/go-varlink"
 
@@ -23,7 +26,12 @@ const rhcServerSocket = "/run/rhc/com.redhat.rhc"
 func newCollectorClient() (*collectorapi.Client, func(), error) {
 	conn, err := net.Dial("unix", rhcServerSocket)
 	if err != nil {
-		return nil, nil, fmt.Errorf("failed to connect to rhc-server: %w", err)
+		slog.Error("failed to connect to rhc-server", "error", err)
+		var pathErr *os.PathError
+		if errors.As(err, &pathErr) && errors.Is(pathErr.Err, syscall.ENOENT) {
+			return nil, nil, fmt.Errorf("rhc-server.socket is not available. Try: systemctl restart rhc-server.socket")
+		}
+		return nil, nil, err
 	}
 	varlinkClient := varlink.NewClient(conn)
 	client := &collectorapi.Client{Client: varlinkClient}


Archana-PandeyM:
This is a nice improvement!

One concern: this now returns the same “socket is not enabled” message for all net.Dial failures, which can hide other root causes like permission.
Could we make this conditional (e.g., only for missing socket) and keep %w for everything else?

Example- 
```
[root@kvm-03-guest14 ~]# su testuser
[testuser@kvm-03-guest14 root]$ rhc collector list
failed to connect to rhc-server: rhc-server.socket is not enabled. Enable it with: systemctl enable --now rhc-server.socket
[testuser@kvm-03-guest14 root]$ 
```
The actual error is getting masked-
`"failed to connect to rhc-server: failed to connect to rhc-server: dial unix /run/rhc/com.redhat.rhc: connect: permission denied"`

jirihnidek:
I believe that the UX could be improved, because majority of users will want to have the `rhc-server.socket` running.

Why you just don't start the `rhc-server.socket` and print the message that the `rhc-server.socket` was started? The `rhc` has to be run as a root user and the `rhc` has rights to do that.

If user do not want to have `rhc-server.server` running and automatically started by `rhc`, then user can always mask the `rhc-server.socket` using `systemctl mask rhc-server.socket`.

pkoprda:
> Why you just don't start the `rhc-server.socket` and print the message that the `rhc-server.socket` was started? The `rhc` has to be run as a root user and the `rhc` has rights to do that.

I don't think `rhc` should alter the system's state or enable a systemd unit behind the scenes, because that violates Principle of Least Privilege. Providing clear instructions is the standard UX pattern here.

mjcr99 (commented):
Tested:
```console
[vagrant@lv-rhel101-go-lab rhc]$ ls /run/rhc/com.redhat.rhc
ls: cannot access '/run/rhc/com.redhat.rhc': No such file or directory
[vagrant@lv-rhel101-go-lab rhc]$ sudo rhc collector list
failed to connect to rhc-server: rhc-server.socket is not enabled. Enable it with: systemctl enable --now rhc-server.socket
```

Enabled the socket following the instructions, and works fine:
```console
[vagrant@lv-rhel101-go-lab rhc]$ sudo systemctl enable --now rhc-server.socket
Created symlink '/etc/systemd/system/sockets.target.wants/rhc-server.socket' → '/usr/lib/systemd/system/rhc-server.socket'.
[vagrant@lv-rhel101-go-lab rhc]$ sudo rhc collector list
No data collectors available.
```

More extreme case:
```console
[vagrant@lv-rhel101-go-lab rhc]$ sudo rm /run/rhc/com.redhat.rhc
[vagrant@lv-rhel101-go-lab rhc]$ sudo rhc collector list
failed to connect to rhc-server: rhc-server.socket is not enabled. Enable it with: systemctl enable --now rhc-server.socket
[vagrant@lv-rhel101-go-lab rhc]$ sudo systemctl enable --now rhc-server.socket
[vagrant@lv-rhel101-go-lab rhc]$ sudo rhc collector list
failed to connect to rhc-server: rhc-server.socket is not enabled. Enable it with: systemctl enable --now rhc-server.socket
```

Under a scenario where the `/run/rhc/com.redhat.rhc` has been somehow accidentally removed, the provided instruction to solve it would not work. Do we want to consider it?

pkoprda:
> Under a scenario where the `/run/rhc/com.redhat.rhc` has been somehow accidentally removed, the provided instruction to solve it would not work. Do we want to consider it?

I can change the error message from:
```
failed to connect to rhc-server: rhc-server.socket is not enabled. Enable it with: systemctl enable --now rhc-server.socket
```
to:
```
failed to connect to rhc-server: rhc-server.socket is not available. Try: systemctl restart rhc-server.socket
```

@mjcr99 what do you think?

mjcr99:
Sounds good to me! I think this approach is more appropriate; in fact, the case covered here is more about not having the `rhc-server.socket` started or in a consistent state, rather than about having it enabled.

mjcr99 (dismissed):
LGTM

jirihnidek:
> > Why you just don't start the `rhc-server.socket` and print the message that the `rhc-server.socket` was started? The `rhc` has to be run as a root user and the `rhc` has rights to do that.
> 
> I don't think `rhc` should alter the system's state or enable a systemd unit behind the scenes, because that violates Principle of Least Privilege. Providing clear instructions is the standard UX pattern here.

Well, you start `yggdrasil.service`, when you run `rhc connect`. Why is this different? And I am not telling that you should do it behind the scene. If the `rhc-connect.socket` was started, then you should print appropriate message to console.

pkoprda:
> > > Why you just don't start the `rhc-server.socket` and print the message that the `rhc-server.socket` was started? The `rhc` has to be run as a root user and the `rhc` has rights to do that.
> > 
> > 
> > I don't think `rhc` should alter the system's state or enable a systemd unit behind the scenes, because that violates Principle of Least Privilege. Providing clear instructions is the standard UX pattern here.
> 
> Well, you start `yggdrasil.service`, when you run `rhc connect`. Why is this different? And I am not telling that you should do it behind the scene. If the `rhc-connect.socket` was started, then you should print appropriate message to console.

That's fair point, but there's important difference: `rhc connect` is explicitly about establishing a connection, so starting services is part of its core function. `rhc collector` commands are read-only command (except `rhc collector enable/disable`) - users expect it to display information, not modify system state. 
Overall, I would prefer to gather feedback from users and then we can modify the code if necessary.

jirihnidek:
> That's fair point, but there's important difference: `rhc connect` is explicitly about establishing a connection, so starting services is part of its core function. `rhc collector` commands are read-only command (except `rhc collector enable/disable`) - users expect it to display information, not modify system state.
> Overall, I would prefer to gather feedback from users and then we can modify the code if necessary.

The `rhc-server.socket` is enabled by `rhc-*.rpm` during installation or update. Thus, the `rhc-server.socket` will be started anyway after the next reboot of the system. It would be OK to start the `rhc-server.socket` from rhc to provided needed functionality.

BTW: when you run `systemctl start rhc-server.socket`, then it will only instruct systemd to listen on `/run/rhc/com.redhat.rhc` socket. It will not start a new process. It will not consume more system resources.

m-horky:
> I don't think rhc should alter the system's state or enable a systemd unit behind the scenes, because that violates Principle of Least Privilege. Providing clear instructions is the standard UX pattern here.

I agree with this. On all systemd-capable hosts (everything but OCI containers), the socket should already be available due to the `.preset` file. If it is not, it means either the user has upgraded rhc to a Varlink-capable version, or there are problems with the daemon.

If the user has upgraded rhc, a system restart is required (RHEL 10 -> 11), or they want to use the tech preview of data collectors (and one more step in this tech preview is valid ask, the PM told me).

If the service is not listening because it has failed to start, there are other issues with the system that have to be solved first, and starting the service won't likely start because the service could crash again.

I'd like to be more explicit about the lifecycle of the daemon.

cmd/rhc/collector_cmd.go
```diff
@@ -23,7 +23,10 @@ const rhcServerSocket = "/run/rhc/com.redhat.rhc"
 func newCollectorClient() (*collectorapi.Client, func(), error) {
 	conn, err := net.Dial("unix", rhcServerSocket)
 	if err != nil {
-		return nil, nil, fmt.Errorf("failed to connect to rhc-server: %w", err)
+		if strings.Contains(err.Error(), "no such file or directory") {
+			return nil, nil, fmt.Errorf("rhc-server.socket is not available. Try: systemctl restart rhc-server.socket")
+		}
```
m-horky: Can you make sure we log this error at Error level? I'd like to keep the fact this connection failed in the log.

m-horky (changes requested):
I've got one request to log the error. The solution is fine otherwise.

cmd/rhc/collector_cmd.go
```diff
@@ -23,7 +23,11 @@ const rhcServerSocket = "/run/rhc/com.redhat.rhc"
 func newCollectorClient() (*collectorapi.Client, func(), error) {
 	conn, err := net.Dial("unix", rhcServerSocket)
 	if err != nil {
-		return nil, nil, fmt.Errorf("failed to connect to rhc-server: %w", err)
+		slog.Error("failed to connect to rhc-server", "error", err)
+		if strings.Contains(err.Error(), "no such file or directory") {
```
m-horky: AI-based suggestion: String matching can be fragile, would this work instead?

```go
var pathErr *os.PathError                                                                                                              
if errors.As(err, &pathErr) && errors.Is(pathErr.Err, syscall.ENOENT) {
```

There are no tests for this, I can't automatically validate one or the other.

m-horky (approved):
ACK

jirihnidek:
> If the service is not listening because it has failed to start, there are other issues with the system that have to be solved first, and starting the service won't likely start because the service could crash again.

It is not about the service (`rhc-server.service`) not listening. It is about systemd not listening on the socket (`rhc-server.socket`). When you run `systemctl start rhc-server.socket`, then you do not start the `rhc-server.service` itself. You only instruct systemd to start listening on the socket file (`/run/rhc/com.redhat.rhsm`). Thus, `systemctl start rhc-server.socket` could fail only in a situation, when there is a high severity bug in the core of systemd, which is very unlikely.

PR #461, opened 2026-06-02
fix(test): Correct rhc collector unit tests

m-horky:
* Card ID: CCT-2520

### DIFF
--- a/cmd/rhc-collector/rhc-collector_test.go
+++ b/cmd/rhc-collector/rhc-collector_test.go
@@ -49,6 +49,8 @@ func TestCreateTmpDir(t *testing.T) {
 	})
 }
 
+// TestGetConfig verifies that getConfig correctly validates the collector ID
+// before attempting to read from the filesystem.
 func TestGetConfig(t *testing.T) {
 	// Note: Can't test happy path without mocking filesystem or root permissions.
 	// This wrapper around collector.GetConfig() reads from /usr/lib/rhc/collectors/.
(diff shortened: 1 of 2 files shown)


cmd/rhc-collector/rhc-collector_test.go
pkoprda, line 13: I see that this test case still uses defer cleanup. Do we want to also change that so it uses `t.Cleanup()`?

pkoprda, line 1: `TestGetArchive()` has nice docstring, would it be possible to add docstring to rest of the test cases?

pkoprda (changes requested):
Looks good, I have two requests below

cmd/rhc-collector/rhc-collector_test.go
m-horky: `t.Cleanup` is only useful in test setup functions, which have to be cleaned up after the test runs. Since these defers exist in the method itself, it should be safe to keep it as such.

We could do it for consistency, though.

pkoprda: Let's keep it how it is for now

pkoprda (approved):
LGTM, thanks!

pkoprda (changes requested):
Cannot merge, the commit https://github.com/RedHatInsights/rhc/pull/461/changes/fa7a74e1c302befd3738fad0a87ff29e7b8cc425 needs to be signed.
> Base branch requires signed commits. Rebase merges cannot be automatically signed by GitHub

m-horky:
Blocked until #473 is merged.

pkoprda:
#472 and #473 have been merged, resolve conflicts in this PR. I am guessing this PR should be just about docstring update

PR #467, opened 2026-06-04
fix: create /var/tmp/rhc if missing

Archana-PandeyM:
* Card Id: CCT-2534

Create /var/tmp/rhc with root-only permissions before creating the collector-specific temporary directory. This prevents rhc-collector from failing when the parent directory does not already exist.

### DIFF
--- a/cmd/rhc-collector/main.go
+++ b/cmd/rhc-collector/main.go
@@ -57,9 +57,38 @@ func run(collectorId, command string) error {
 	return nil
 }
 
-// createTmpDir creates a temporary directory for collector output in rhcTmpDir.
-// Returns the directory path or an error if creation fails.
+// createTmpDir ensures rhcTmpDir exists with root-only permissions (0700)
+// and creates a collector-specific temporary directory inside it. If the
+// parent directory exists with different permissions, they are reset to
+// 0700. Returns the temporary directory path or an error if any step fails.
 func createTmpDir() (string, error) {
+	// Ensure the parent directory exists
+	if err := os.MkdirAll(rhcTmpDir, 0700); err != nil {
+		slog.Error("failed to create rhc temporary directory", "error", err)
+		return "", fmt.Errorf("failed to create rhc temporary directory: %w", err)
+	}
+
+	// Verify permissions and fix if necessary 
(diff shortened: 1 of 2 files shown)


cmd/rhc-collector/main.go
```diff
@@ -60,6 +60,11 @@ func run(collectorId, command string) error {
 // createTmpDir creates a temporary directory for collector output in rhcTmpDir.
 // Returns the directory path or an error if creation fails.
 func createTmpDir() (string, error) {
+	if err := os.MkdirAll(rhcTmpDir, 0700); err != nil {
```
stephenbourke, line 66: I don't think this will change the permissions on an already existing directory. Do we want to cover this case?

cmd/rhc-collector/main.go
```diff
@@ -60,6 +60,11 @@ func run(collectorId, command string) error {
 // createTmpDir creates a temporary directory for collector output in rhcTmpDir.
```
stephenbourke: You could perhaps update the docstring to expand more on what the function now provides based on your change.

cmd/rhc-collector/main.go
Archana-PandeyM: I intentionally kept it this way. As per acceptance criteria on the ticket, we want to create the directory if it does not exist. os.MkdirAll() does not alter permissions on existing dir.

stephenbourke (approved):
LGTM. Just needs a rebase.

cgwalters:
This ended up breaking bootc because our container export flow requires that all content have a SELinux label defined. See https://github.com/bootc-dev/bootc/pull/2395 (which I ended up rolling into another PR).

A few things here.

First, hardcoded file names in `/tmp` and `/var/tmp` are just a bad idea. They are at best a DoS vector because there's a chance for e.g. unprivileged other code to create the dirs first.

You added a tmpfiles.d config, which is fine.

But why do you have a hardcoded tmpdir at all? You should have a randomly named tmpdir like `rhc-XXXXXX` created by a sane language tmpdir API instead, and it should not be owned by the RPM package and not created by systemd-tmpfiles.

PR #468, opened 2026-06-08
fix: Allow multiple arguments for configure features enable/disable

stephenbourke:
* Card Id: CCT-2459
    
     - Allow content, analytics and remote-management arguments
       to be specified together for configure features enable/disable
     - Add integration tests to test multiple arguments on configure
       features enable/disable


integration-tests/__pycache__/__init__.cpython-314.pyc
pkoprda, line 1: These `.pyc` files shouldn't be included

integration-tests/test_configure_feature.py
```diff
@@ -709,3 +709,173 @@ def test_configure_features_disable_idempotent_json(rhc):
...
+        feature names in one command and update the preference cache accordingly.
+    :tags: Tier 1
+    :steps:
+        1. Disconnect the host.
+        4. Run ``rhc configure features enable content analytics remote-management``.
+        5. Verify all features show preference ``enable``.
+    :expectedresults:
+        1. Host is disconnected.
+        2. Disable exits 0.
+        5. JSON status shows ``enable`` for each feature.
```
pkoprda: This numbering seems wrong

integration-tests/test_configure_feature.py
```diff
@@ -709,3 +709,173 @@ def test_configure_features_disable_idempotent_json(rhc):
...
+    :id: f4a5b6c7-d8e9-4f0a-bcde-1234567890ab
+    :title: ``configure features`` disable accepts multiple FEATURE arguments
+    :description:
+        On a connected host, ``disable`` accepts one or more
+        feature names in one command and update the live feature state accordingly.
+    :tags: Tier 1
+    :steps:
+        1. Connect the host.
+        2. Run ``rhc configure features disable content analytics remote-management``.
+        4. Verify all features show live state ``disabled`` is False.
```
pkoprda: Ditto

pkoprda (dismissed):
Left some initial comments below.

Also please fix the typo in the commit `mulitple` -> `multiple`

integration-tests/__pycache__/__init__.cpython-314.pyc
stephenbourke: Ah they are not currently in the gitignore. Thanks!

cmd/rhc/configure_features_cmd.go
```diff
@@ -173,65 +173,72 @@ func beforeFeaturesEnableAction(ctx context.Context, cmd *cli.Command) (context.
 	}
 	configureUI(cmd)
 
-	if cmd.Args().Len() != 1 {
-		return ctx, cli.Exit("this command requires a single FEATURE argument", exitcode.Usage)
+	if cmd.Args().Len() == 0 || cmd.Args().Len() > 3 {
```
mjcr99: I would use `feature.All()` instead of  `3`.

cmd/rhc/configure_features_cmd.go
```diff
@@ -242,39 +249,40 @@ func featuresEnableActionNotRegistered(_ context.Context, _ *cli.Command, target
...
-		if err != nil {
-			return cli.Exit(fmt.Sprintf("failed to check status of target feature '%s': %v", targetName, err), exitcode.Software)
-		}
-		if featureEnabled {
-			slog.Debug("feature is already enabled", "feature", targetName)
-			return nil
+func featuresEnableActionRegistered(_ context.Context, _ *cli.Command, targetNames []string) error {
+	for _, targetName := range targetNames {
+		target := feature.MustGet(targetName)
+		// enable required feature
```
mjcr99: Let's keep consistency with plurals in the comments, I would use:
```go
// enable required features
// disable dependent features
```

integration-tests/test_configure_feature.py
```diff
@@ -709,3 +709,173 @@ def test_configure_features_disable_idempotent_json(rhc):
         ).stdout
     )
     assert a == b
+
+
+@pytest.mark.tier1
+def test_configure_features_enable_multiple_features_disconnected(external_candlepin, rhc, test_config):
+    """
+    :id: f4a5b6c7-d8e9-4f0a-bcde-1234567890ab
```
mjcr99: The ID of the test is the same as the second one, could we change it?

integration-tests/test_configure_feature.py
```diff
@@ -709,3 +709,173 @@ def test_configure_features_disable_idempotent_json(rhc):
...
+    :title: ``configure features`` enable accepts multiple FEATURE arguments
+    :description:
+        On a disconnected host, ``enable`` accepts one or more
+        feature names in one command and update the preference cache accordingly.
```
mjcr99: Should here be updates instead of update?

mjcr99 (dismissed):
Thanks @stephenbourke!

I have some comments.
* I think we should consider splitting the integration tests implementation into a different commit. The commit message does not reference the integration tests implementation it contains.
* Also, I think there is a typo in the commit message `remote-managment` please fix it.

stephenbourke:
> Thanks @stephenbourke!
> 
> I have some comments.
> 
>     * I think we should consider splitting the integration tests implementation into a different commit. The commit message does not reference the integration tests implementation it contains.
> 
>     * Also, I think there is a typo in the commit message `remote-managment` please fix it.

Many thanks Manu. I've updated the PR based on your feedback.

cmd/rhc/configure_features_cmd.go
```diff
@@ -242,39 +250,40 @@ func featuresEnableActionNotRegistered(_ context.Context, _ *cli.Command, target
...
-			return cli.Exit(fmt.Sprintf("failed to enable target feature '%s': %v", targetName, err), exitcode.Software)
+		// enable target features
+		{
+			featureEnabled, err := target.IsEnabled()
+			if err != nil {
+				return cli.Exit(fmt.Sprintf("failed to check status of target feature '%s': %v", targetName, err), exitcode.Software)
+			}
+			if featureEnabled {
+				slog.Debug("feature is already enabled", "feature", targetName)
+				continue
```
pkoprda, line 280: Nit: use `return` instead of `continue`

pkoprda (commented):
Works like a charm, I have just comment.

cmd/rhc/configure_features_cmd.go
stephenbourke: We are looping through each feature argument. If it is already enabled/disabled we want to continue the loop and move onto the next feature target rather than bailing.

pkoprda: Make sense

pkoprda (approved):
LGTM

mjcr99 (approved):
GJ @stephenbourke!
LGTM.

PR #471, opened 2026-06-08
fix: invoke WriteTimerCache during collector execution

zpetrace:
This change updates `executeCollector` to:
- Capture the `startTime` and `endTime` around the command execution.
- Safely extract the process exit code via `exec.ExitError` type assertion.
- Call `collector.WriteTimerCache` to write the execution info to the filesystem
cache at `/var/cache/rhc/collectors/COLLECTOR-ID.json`.

The work is done for CCT-2537

### DIFF
--- a/cmd/rhc-collector/main.go
+++ b/cmd/rhc-collector/main.go
@@ -5,6 +5,7 @@ import (
 	"log/slog"
 	"os"
 	"os/exec"
+	"time"
 
 	"github.com/redhatinsights/rhc/internal/collector"
 	httpapi "github.com/redhatinsights/rhc/internal/http"
@@ -34,16 +35,17 @@ func main() {
 }
 
 func run(collectorId, command string) error {
+	config, err := getConfig(collectorId)
+	if err != nil {
+		return err
+	}
 	tmpDir, err := createTmpDir()
 	if err != nil {
 		return err
 	}
 	defer cleanup(tmpDir)
-	config, err := getConfig(collectorId)
-	if err != nil {
-		return err
-	}
-	if err = executeCollector(command, tmpDir); err != nil {
+
+	if err = executeCollector(collectorId, command, tmpDir); err != nil {
 		return err
 	}
 	archivePath, err := getArchivePath(tmpDir)
@@ -112,15 +114,44 @@ func getConfig(collectorId string) (collector.Config, error) {
 
 // executeCollector runs the specified collector command and stores output in tmpDir.
 // Returns an error if the command execution fails.
-func executeCollector(command, tmpDir string) error {
+func executeCollector(collectorId, command, tmpDir string) error {
 	// TODO: Run collector as specific user and group (defined in config of collector)
 	cmd := exec.Command("/bin/sh", "-c", fmt.Sprintf("%s %q", command, tmpDir))
 	cmd.Dir = tmpDir
+
+	// Capture start/end time and execute the command
+	startTime := time.Now()
 	output, err := cmd.CombinedOutput()
+	endTime := time.Now()
+
+	exitCode := 0
+	if err != nil {
+		if exitErr, ok := err.(*exec.ExitError); ok {
+			exitCode = exitErr.ExitCode()
+		} else {
+			exitCode = 1
+		}
+	}
+
+	// Timer payload and writing to cache
+	timerPayload := collector.Timer{
+		ID:           collectorId,
+		LastStarted:  startTime,
+		LastFinished: endTime,
+		ExitCode:     exitCode,
+	}
+	cacheErr := collector.WriteTimerCache(collectorId, timerPayload)
+	if cacheErr != nil {
+		slog.Error("Failed to write timer cache", "error", cacheErr)
+	}
 	if err != nil {
 		slog.Error("failed to execute collector", "error", err, "output", string(output))
 		return fmt.Errorf("failed to execute collector: %w", err)
 	}
+	if cacheErr != nil {
+		return fmt.Errorf("failed to write timer cache: %w", cacheErr)
+	}
+
 	slog.Info("collector has ran successfully", "output", string(output))
 	return nil
 }
--- a/integration-tests/test_collector_api.py
+++ b/integration-tests/test_collector_api.py
@@ -16,6 +16,9 @@ import textwrap
 from utils.varlink import run_varlinkctl
 from utils.systemctl import is_service_active
 
+RHC_COLLECTOR = "/usr/libexec/rhc/rhc-collector"
+TIMER_CACHE_DIR = "/var/cache/rhc/collectors"
+
 
 @pytest.fixture(scope="module")
 def rhc_server_socket():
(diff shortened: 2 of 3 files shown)


internal/collector/collector.go
```diff
@@ -156,7 +156,8 @@ func ReadTimerCache(id string) (*Timer, error) {
 }
 
 // WriteTimerCache saves timer data to the cache for the specified collector ID.
-func WriteTimerCache(id string, timer Timer) error {
+// If dir is empty, TimerDir is used.
+func WriteTimerCache(id string, timer Timer, dir string) error {
```
stephenbourke: Is there an advantage to adding a "dir" parameter to this function's signature if production always uses the hard coded location?

If it is, i think the loadTimerFromFile function would have to allow for a "dir" parameter too so it could read back from a previously written location. It would also require a means to determine the previously written directory location.

It may be a simpler implementation to keep the cache directory hard coded if specifying the directory is not strictly required.

zpetrace: Well it was mainly added for unit testing as production writes to /var/cache/rhc/collectors/ and on GitHub this path is not writeable so it was failing at runtime. 

But I also understand why that would be easier. So I will revert the dir parameter, add a skip to the failing TestExecuteCollector if the path is not writeable. Timer cache writes are now covered by the integration test I have added anyway, which runs rhc-collector against the real cache location.

integration-tests/test_collector_api.py
```diff
@@ -242,6 +245,50 @@ def test_collector_info_with_timer_cache(
...
+    os.makedirs(TIMER_CACHE_DIR, exist_ok=True)
+    if os.path.exists(cache_path):
+        os.remove(cache_path)
+
+    subprocess.run(
+        [RHC_COLLECTOR, collector_id, "/bin/true"],
+        capture_output=True,
+        check=False,
+    )
+
```
Archana-PandeyM: since check=False is used here,  I would suggest asserting return-code.

Archana-PandeyM (changes requested):

integration-tests/test_collector_api.py
```diff
@@ -242,6 +245,62 @@ def test_collector_info_with_timer_cache(
...
+    """
+    collector_id = collector_config["id"]
+    cache_path = os.path.join(TIMER_CACHE_DIR, f"{collector_id}.json")
+
+    os.makedirs("/var/tmp/rhc", exist_ok=True)
+    os.makedirs(TIMER_CACHE_DIR, exist_ok=True)
+    if os.path.exists(cache_path):
+        os.remove(cache_path)
+
+    if not rhc.is_registered:
```
Archana-PandeyM: I think the test would always skip here, system is not registered yet.

zpetrace: Oh you're right, I wanted to skip the test at first, then created an assert further in the code and forgot to delete this skip. My bad.

zpetrace:
My manual test of the change:

```
[root@e364f140-97fe-4e5e-b681-9c77d3602fcf ~]# rpm -qa | grep rhc
rhc-debugsource-0.3.9-1.20260610111131857527.pr471.36.g27575bb.el10.x86_64
rhc-debuginfo-0.3.9-1.20260610111131857527.pr471.36.g27575bb.el10.x86_64
rhc-0.3.9-1.20260610111131857527.pr471.36.g27575bb.el10.x86_64
```
I have created a test collector, so the upload itself will fail, but the timestamps are captured:

```
[root@e364f140-97fe-4e5e-b681-9c77d3602fcf ~]# /usr/libexec/rhc/rhc-collector test.collector /bin/true
2026/06/10 12:31:02 INFO starting rhc-collector id=test.collector
....
2026/06/10 12:31:02 ERROR rhc-collector exited with error error="failed to upload archive: upload to https://cert.console.redhat.com/api/test/v1/upload failed with status code: 415"

[root@e364f140-97fe-4e5e-b681-9c77d3602fcf ~]# cat /var/cache/rhc/collectors/test.collector.json
{"last_started":{"timestamp":1781094662},"last_finished":{"timestamp":1781094662,"exit_code":0}}


[root@e364f140-97fe-4e5e-b681-9c77d3602fcf ~]# rhc collector info test.collector
Name:      Test Collector
Feature:   analytics

Last run:  Wed 2026-06-10 12:31 UTC (1m ago)
Next run:  -

Config:   /usr/lib/rhc/collectors/test.collector.toml
Service:  rhc-collector-test.collector.service
Timer:    rhc-collector-test.collector.timer

```
Due to the command failing, both timestamps are of same value as the command started and finished in the same second.

stephenbourke (dismissed):
LGTM

mjcr99 (dismissed):

stephenbourke:
> > My manual test of the change:
> > ```
> > [root@e364f140-97fe-4e5e-b681-9c77d3602fcf ~]# rpm -qa | grep rhc
> > rhc-debugsource-0.3.9-1.20260610111131857527.pr471.36.g27575bb.el10.x86_64
> > rhc-debuginfo-0.3.9-1.20260610111131857527.pr471.36.g27575bb.el10.x86_64
> > rhc-0.3.9-1.20260610111131857527.pr471.36.g27575bb.el10.x86_64
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I have created a test collector, so the upload itself will fail, but the timestamps are captured:
> > ```
> > [root@e364f140-97fe-4e5e-b681-9c77d3602fcf ~]# /usr/libexec/rhc/rhc-collector test.collector /bin/true
> > 2026/06/10 12:31:02 INFO starting rhc-collector id=test.collector
> > ....
> > 2026/06/10 12:31:02 ERROR rhc-collector exited with error error="failed to upload archive: upload to https://cert.console.redhat.com/api/test/v1/upload failed with status code: 415"
> > 
> > [root@e364f140-97fe-4e5e-b681-9c77d3602fcf ~]# cat /var/cache/rhc/collectors/test.collector.json
> > {"last_started":{"timestamp":1781094662},"last_finished":{"timestamp":1781094662,"exit_code":0}}
> > 
> > 
> > [root@e364f140-97fe-4e5e-b681-9c77d3602fcf ~]# rhc collector info test.collector
> > Name:      Test Collector
> > Feature:   analytics
> > 
> > Last run:  Wed 2026-06-10 12:31 UTC (1m ago)
> > Next run:  -
> > 
> > Config:   /usr/lib/rhc/collectors/test.collector.toml
> > Service:  rhc-collector-test.collector.service
> > Timer:    rhc-collector-test.collector.timer
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Due to the command failing, both timestamps are of same value as the command started and finished in the same second.

Apologies, I accidentally hit the close PR button.  :fearful: 

I think having a sleep in the command to show the timer reflects the command execution time regardless of pass/failure might help demonstrate the implementation?

`/usr/libexec/rhc/rhc-collector test.collector "sleep 2; /bin/true"`

`{"last_started":{"timestamp":1781097910},"last_finished":{"timestamp":1781097912,"exit_code":0}}`

integration-tests/test_collector_api.py
```diff
@@ -242,6 +245,59 @@ def test_collector_info_with_timer_cache(
...
+    assert os.path.exists(cache_path), "Timer cache file should be created"
+
+    with open(cache_path) as cache_file:
+        cache = json.load(cache_file)
+
+    assert "last_started" in cache
+    assert "last_finished" in cache
+    assert cache["last_finished"]["exit_code"] == 0
+
+    os.remove(cache_path)
```
Archana-PandeyM: cleanup is not guaranteed on failure,  I suggest to wrap the test body in a try/finally and always remove cache_path in finally.

Archana-PandeyM (changes requested):

integration-tests/test_collector_api.py
```diff
@@ -242,6 +245,59 @@ def test_collector_info_with_timer_cache(
...
+        [RHC_COLLECTOR, collector_id, "/bin/true"],
+        capture_output=True,
+        check=False,
+    )
+    # If the system is not registered with RHSM, the collector should fail with exit code 1
+    if not rhc.is_registered:
+        assert collector_run.returncode == 1
+    else:
+        # If the system is registered with RHSM, the collector should succeed with exit code 0
+        assert collector_run.returncode == 0, (
```
Archana-PandeyM: Since this test itself never registers rhc, this else block is dead code.

Archana-PandeyM (changes requested):

cmd/rhc-collector/main.go
```diff
@@ -112,15 +114,43 @@ func getConfig(collectorId string) (collector.Config, error) {
...
+	timerPayload := collector.Timer{
+		ID:           collectorId,
+		LastStarted:  startTime,
+		LastFinished: endTime,
+		ExitCode:     exitCode,
+	}
+	if cacheErr := collector.WriteTimerCache(collectorId, timerPayload); cacheErr != nil {
+		slog.Error("Failed to write timer cache", "error", cacheErr)
+		return fmt.Errorf("failed to write timer cache: %w", cacheErr)
+	}
```
Archana-PandeyM: @stephenbourke @mjcr99 Hi, OOC-
One concern with the current implementation: if collector.WriteTimerCache(...) fails, we return that error immediately, which can mask the original collector execution error.

I support writing timer cache on both success/failure, but current code may hide the real collector failure when cache write also fails. Can we keep execution error as primary and treat cache-write error as secondary (or warning)?

stephenbourke: It's a good spot. The function should defer return until both error state conditions are determined and log where applicable.

cmd/rhc-collector/main.go
```diff
@@ -112,15 +114,44 @@ func getConfig(collectorId string) (collector.Config, error) {
...
+	cacheErr := collector.WriteTimerCache(collectorId, timerPayload)
+	if cacheErr != nil {
+		slog.Error("Failed to write timer cache", "error", cacheErr)
+	}
 	if err != nil {
 		slog.Error("failed to execute collector", "error", err, "output", string(output))
 		return fmt.Errorf("failed to execute collector: %w", err)
 	}
+	if cacheErr != nil {
+		return fmt.Errorf("failed to write timer cache: %w", cacheErr)
```
stephenbourke, line 152: Should we negate the rest of the collector's execution (archive, upload) if we fail to write to the timer cache?

zpetrace: @stephenbourke I don't think a timer cache write failure should stop archive creation or upload IMO. The timer cache is local metadata used only to populate last_run in the collector Info API. The server already treats it as optional.

stephenbourke: Cool. I think you can remove this block so as not to error out early in `run`.

Archana-PandeyM (approved):
LGTM!

PR #472, opened 2026-06-08
fix(test): Do not delete archive before it is used

m-horky:
* Card ID: CCT-2520

The `defer os.Remove(archivePath)` ran when the helper function returned, deleting the archive before uploadArchive could use the returned path.

- Replaced with `t.Cleanup(...)` which runs after the whole test completes.
- Switched the output directory to `t.TempDir()`.

Due to the recent GitHub repository configuration changes, this PR implements a portion of #461.


pkoprda (approved):
ACK

PR #473, opened 2026-06-08
fix(test): Correct assumptions in TestGetArchive

m-horky:
* Card ID: CCT-2520

The tests called `getArchivePath(tmpDir)`, which internally defaults the output directory to `/var/tmp/rhc/`. In Fedora mock/koji environment this directory might not exist, resulting in tar failing to write.

Instead, we write into a temporary directory that gets cleaned when the test finishes.

The test was also renamed as TestGetArchive to reflect the method it is testing.

Due to the recent GitHub repository configuration changes, this PR implements a portion of https://github.com/RedHatInsights/rhc/pull/461.

### DIFF
--- a/cmd/rhc-collector/rhc-collector_test.go
+++ b/cmd/rhc-collector/rhc-collector_test.go
@@ -116,54 +116,53 @@ func TestExecuteCollector(t *testing.T) {
 	})
 }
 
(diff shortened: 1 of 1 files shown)


pkoprda (approved):
ACK

PR #474, opened 2026-06-08
feat: A new Varlink interface com.redhat.rhsm (one socket)

jirihnidek:
* Card ID: CCT-2348
* Added a new interface file .varlink for `com.redhat.rhsm.testing` interface, new rhsmapi also contains generat.go and generated go file with generated Go code for given Varlink interface
* The interface contains two methods:
  * `IsRegistered()`
  * `Ping()`
* It uses golang module: `github.com/jirihnidek/rhsm2`
* Updated `go.mod` and `go.sum`


cmd/rhc-server/rhsm.go
m-horky, line 1: I am not convinced about the placement of this file. It contains business logic, it might need to be placed somewhere in `pkg/` instead. The whole code is somehow in the middle of a refactor currently (and will be until CCT-2399 is resolved), I can imagine letting this pass for now until the rest has been cleaned up.

cmd/rhc-server/rhsm.go
```diff
@@ -0,0 +1,56 @@
+package main
+
+import (
+	"github.com/jirihnidek/rhsm2"
+)
+
+var AppName = "rhc-server"
```
m-horky: I think this should be a `const appName`, we're not using it anywhere else.

cmd/rhc-server/rhc-server.go
```diff
@@ -66,3 +70,48 @@ func (b *Backend) Info(in *collectorapi.InfoIn) (*collectorapi.InfoOut, error) {
...
+	}
+	return &rhsmapi.PingOut{Status: status}, nil
+}
+
+// IsRegistered checks if the system is registered with RHSM.
+func (b *Backend) IsRegistered(in *rhsmapi.IsRegisteredIn) (*rhsmapi.IsRegisteredOut, error) {
+	registered, err := IsSystemRegistered()
+	if err != nil {
+		// When it is not possible to determine registration status, then log the reason
+		// and return false
```
m-horky, line 112: Should we return a Varlink error object instead? What could be the reason for not being able to obtain the status? I'd like to be more explicit about the values we return, this is misleading.

varlink/rhsmapi/com.redhat.rhsm.testing.varlink
```diff
@@ -0,0 +1,28 @@
...
+# Generic error related to the server response, when the server is not available
+# or the server response the exit code that is not successful
+error FailedServerResponse(message: string)
+
+# Check if the candlepin server is available
+method Ping(
+    metadata: ?Metadata
+) -> (
+    status: object
+)
```
m-horky, line 26: Could you please document what this `object` is? I.e. "raw JSON output of Candlepin's `METHOD /foo`".

m-horky (changes requested):
Thanks. I don't have any major concerns, for now this should work.

In the commit message, could you mention that this Varlink API status is "testing" and is not meant for external use?

cmd/rhc-server/rhc-server.go
jirihnidek: We should not return error object from this method. I did some experiments with this idea and it had zero benefit for client tool. This method does not communicate with candlepin server at all. It only tries to read consumer certificate. The fact that you are not able to read consumer object as a root means that the system is not registered in 99.999% of cases. Thus, we should return only `true` or `false`.

There could be some theoretical corner cases that root is not able to read consumer certificate, because admin did some non-sense like `chmod 000 /etc/pki/consumer/cert.pem`, but for this reason there is debug log to help with troubleshooting such issue. We should not complicate our API for these rare corner cases.

cmd/rhc-server/rhsm.go
jirihnidek: It is pity that you cannot create pointer on const. Shouldn't be this defined in something like `constants.go`? It has potential to be used outside `rhsm.go`.

cmd/rhc-server/rhsm.go
jirihnidek: There are two `rhsm.go` files in the source code of `rhc`. It is not perfect, because the first file contains client code for D-Bus communication with `rhsm.service` and the second file contains server code of Varlink. It could be source of confusion.

jlocash: This seems like a side effect of having on a single `Backend` that provides all methods for all interfaces.  There should really be one backend per interface. Then the rhsm backend implementation can live in `internal/rhsm`

cmd/rhc-server/rhc-server.go
```diff
@@ -1,12 +1,16 @@
 package main
 
 import (
+	"encoding/json"
+	"errors"
 	"fmt"
 	"log/slog"
 
+	"github.com/jirihnidek/rhsm2"
```
jlocash, line 9: Is there a reason this code lives in a personal repo vs in this repo under `pkg/` or in another repo under the org?

jlocash (commented):
Thanks for the PR! I left a few comments

cmd/rhc-server/rhsm.go
m-horky: For now I'd keep it as local and private variable. If there's a need in the future, we'll clean it up.

cmd/rhc-server/rhc-server.go
jirihnidek: Faster development process.

jirihnidek:
@jlocash GitHub UI has some problems today. It wasn't possible to reply directly to your comment. Thus, I am doing it in this way.

> This seems like a side effect of having on a single Backend that provides all methods for all interfaces. There should really be one backend per interface. Then the rhsm backend implementation can live in internal/rhsm

All existing interfaces use the same backend. The `com.redhat.rhc.collector` and `com.redhat.rhc.collector` uses `Backend`. I just followed existing pattern. If somebody will want to do refactoring in the future, then I am not against, but it does not belong to this PR. BTW: I used another backend for two socket approach #438, because it was necessary, but this approach with two sockets was refused.

varlink/rhsmapi/com.redhat.rhsm.testing.varlink
jirihnidek: Fixed

cmd/rhc-server/rhc-server.go
jirihnidek: BTW: It is testing interface. When it will be need to return some error in the future, then we could change it in the future. I would keep it simple as it was approved in the design document.

cmd/rhc-server/rhsm.go
jlocash: @m-horky I'd like to know your thoughts on this suggestion.

cmd/rhc-server/rhc-server.go
m-horky: Good point. Yeah okay, it's fine in that case.

jlocash:
@jirihnidek Yeah I was mainly mentioning it as a reason the file seems out of place. You're right that this PR is not the place to make the change I suggested. Not a blocker

cmd/rhc-server/rhsm.go
m-horky: > There are two rhsm.go files in the source code of rhc.

The one under internal/ could be renamed to `identity.go`, for example. There's already `content.go` next to it.

> There should really be one backend per interface.

I think that would make sense, yeah. The [library example](https://github.com/emersion/go-varlink/blob/9c0e29254dd5f6738a130c529a20fbecf576905a/example/main.go#L53-L54) defines a separate backend for each namespace.

> Then the rhsm backend implementation can live in internal/rhsm

I don't love that. Though we haven't really started properly, all presentation layer code (which Varlink is) should stay in `cmd/`. Look up the [ARCHITECTURE.md](https://github.com/RedHatInsights/rhc/blob/main/doc/ARCHITECTURE.md), and we can chat about it outside of GitHub as well, but we need to create more separation between the layers of the code, not tangle them even more.

cmd/rhc-server/rhc-server.go
```diff
@@ -66,3 +70,48 @@ func (b *Backend) Info(in *collectorapi.InfoIn) (*collectorapi.InfoOut, error) {
 
 	return &collectorapi.InfoOut{Info: *info}, nil
 }
+
+// Ping checks the status of the RHSM server.
```
m-horky, line 74: I'd like to make sure we split this off and have a separate backend for each namespace, this will break as soon as two .varlink files define the same method or object.

I created downstream tracker CCT-2601 for that.

m-horky (approved):
ACK, no major concerns. Let's iterate as we go.

PR #475, opened 2026-06-08
fix(docs): Update docstrings

m-horky:
GitHub doesn't like force-pushing to rebase if the force-pushed history contains unsigned commits. Opening as the last part of #461.


pkoprda (approved):
ACK

PR #480, opened 2026-06-09
fix: Unify packaging across distributions

m-horky:
* Card ID: CCT-2265

go-vendor-tools isn't available for CentOS Stream or RHEL. Apparently,
the COPR for it is internal and is not expected to be used by 3rd
parties. The package itself is present in Fedora only.

rhc.spec:
- Bump the Epoch to 1 to match downstream.
- Enable rhcd compatibility on CentOS Stream and RHEL. This includes
  the compatibility shim for systemd, config file migration, and
  SELinux' semanage calls. The config file migration is corrected to
  be split between %pre and %post (on upgrade, the source config might
  already be gone).
- Do not pass '-f' to %gometa: we don't need to use the future
  version of the list of supported architectures, we're tied to RHEL.
- Add %if statements to separate what Fedora and RHEL do in terms of
  managing downstream builds and vendored dependency handling.
- Use %gocheck instead of %gocheck2 macro that is provided by
  go-vendor-tools and isn't available on CentOS Stream. Since we don't
  use submodules, there shouldn't be any behavioral difference we'd
  rely on.
- Recommends: insights-client on Fedora, Requires: on RHEL.
- Include LICENSE in the files
- Make %goprep work with %autopatch and document how we handle multiple
  source tarballs.

Makefile:
- Add archive-deps-fedora and archive-deps-rhel targets. Fedora one
  is a wrapper for go_vendor_archive, RHEL uses tar directly.

.packit.yaml:
- srpm_build_deps only installs packages needed by 'make srpm' and
  omits dependencies needed by Meson.
- Use go-vendor-tools if installed (on Fedora), fallback to
  self-managed vendor archive creation based on what go-vendor-tools
  do themselves to achieve/approach reproducibility.
- Since the targets: have been changed in the past to support
  go-vendor-tools that only existed in EPEL, they have been reverted
  to their original format.

rpmbuild.yml, CONTRIBUTING.md:
- We do not need the COPR.

dist/:
- With rhc.spec supporting both Fedora and RHEL, the other downstream
  spec file could be dropped.


.github/workflows/rpmbuild.yml
```diff
@@ -27,14 +27,10 @@ jobs:
...
     steps:
-      - name: "Add COPR repository (Go Vendor Tools)"
-        run: |
-          dnf copr enable -y @go-sig/go-vendor-tools-dev
-
       - name: "Install basic packages for 'dnf builddep'"
         run: |
           dnf --setopt install_weak_deps=False install -y \
-            git rpm-build go-vendor-tools go-rpm-macros
+            git rpm-build go-vendor-tools askalono-cli go-rpm-macros
```
jirihnidek: Why did you add `askalono-cli` here? It is still required in the `rhc.spec` (`BuildRequires:  askalono-cli`). Thus, it is automatically installed, when you run: `dnf --setopt install_weak_deps=False builddep -y rhc.spec` in the step "Install required packages".

.github/workflows/rpmbuild.yml
```diff
@@ -47,12 +43,11 @@ jobs:
         run: |
           dnf --setopt install_weak_deps=False builddep -y rhc.spec
 
-      - name: "Build srpm"
+      - name: "Build RPM artifacts"
```
jirihnidek, line 46: I had building of rpm in separate step, because only RPM artifacts are uploaded in the step "Upload RPM artifacts". It looks like that all artifacts will be uploaded now, but it is not true. Only `*.rpm` files are uploaded. No archive is uploaded and no `*.srpm` is uploaded. This change could give false expectation for users in the future. It is minor thing.  No blocker.

.packit.yaml
jirihnidek, line 1: Two files with instructions how to build `rhc`... I wish to be able to remove this file in the future or at least not use it for every PR.

Makefile
```diff
@@ -15,14 +15,29 @@ build:
...
+		go.mod go.sum vendor/
+
 .PHONY: srpm
-srpm: archive
+srpm:
 	rpmbuild --define "_sourcedir $$(pwd)" -bs rhc.spec
 
 .PHONY: rpm
-rpm: archive
+rpm: srpm
```
jirihnidek, line 37: This is not correct.  Try to run two commands: `make clean`  and then `make rpm`. It will fail, because there are missing files:

```
RPM build errors:
    Bad file: /home/jhnidek/github/redhat_insights/rhc/rhc-0.3.9.tar.gz: No such file or directory
    Bad file: /home/jhnidek/github/redhat_insights/rhc/rhc-0.3.9-vendor.tar.bz2: No such file or directory
```

You should be able to run `make rpm` with the fresh git checkout.

It seems that the `make rpm` needs to depend on `make archive-deps-fedora` on Fedora. It also needs to depend on `make archive-deps-rhel` on RHEL. This is not great and it does not have simple solution.

Do we really need `go-vendor-tools`? Cannot we just use the same approach that we use for RHEL on Fedora?

rhc.spec
```diff
@@ -37,24 +58,35 @@ Client tool to register Fedora, CentOS Stream or Red Hat Enterprise Linux
...
 %prep
-%goprep -A
-%setup -q -T -D -a1 %{forgesetupargs}
+# Unpack Source0 and set up the Go build directory. Since -k is not passed in,
+# the vendor/ directory from the tarball is explicitly deleted.
+%goprep
+# Unpack Source1 into the build tree, providing the vendor/ directory.
+%setup -q -T -D -a1 -n %{name}-%{version}
+# Apply patches, if present.
+%autopatch -p1
```
jirihnidek, line 67: Can we suppress warnings that no patch is available?

```
warning: /home/jhnidek/github/redhat_insights/rhc/rhc.spec line 67: autopatch: no matching patches in range
```

Makefile
jirihnidek: Note: The change should be something like:

```
rpm: archive-deps-fedora archive
```

BTW: There is some bug in GitHub. It is not possible to edit pending comment without loosing markdown formatting. :-/

rhc.spec
jirihnidek, line 1: After `make rpm` I can see:

```
RPM build warnings:
    /home/jhnidek/github/redhat_insights/rhc/rhc.spec line 67: autopatch: no matching patches in range
    File listed twice: /usr/share/licenses/rhc
    File listed twice: /usr/share/licenses/rhc/LICENSE
```

rhc.spec
```diff
@@ -37,24 +58,35 @@ Client tool to register Fedora, CentOS Stream or Red Hat Enterprise Linux
...
 
 %generate_buildrequires
+%if 0%{?fedora}
+# Generate data for the licence check go-vendor-tools provides.
 %go_vendor_license_buildrequires -c %{S:2}
+%endif
 
 %build
-export GO_LDFLAGS="-X github.com/redhatinsights/rhc/pkg/version.Version=%{version} -X main.ServiceName=yggdrasil"
```
jirihnidek, line 47: Just question: Why was `-X main.ServiceName=yggdrasil"` removed? Is it because the "yggdrasil" is default value and RHEL10+ will use only yggdrasil?

jirihnidek (changes requested):
I have some comments, questions and I believe that I found some issues.

rhc.spec
m-horky: There is no such variable as ServiceName, it was likely dropped without this LDFLAG being deleted. Should I move it to a separate commit/PR?

Makefile
m-horky: > It seems that the make rpm needs to depend on make archive-deps-fedora on Fedora. It also needs to depend on make archive-deps-rhel on RHEL. This is not great and it does not have simple solution.

I am not happy about it either.

> Do we really need go-vendor-tools? Cannot we just use the same approach that we use for RHEL on Fedora?

It will work, I'm unsure whether it will be acceptable from the distribution point of view.

.packit.yaml
m-horky: Until we have an alternative way to trigger integration tests, we have to keep Packit. I'm totally onboard with doing it differently in the future.

.github/workflows/rpmbuild.yml
m-horky: Only the files in `rpms/` are archived, as per the "Collect RPM artifacts" step just below. Yeah, this isn't necessary, but I felt like they could be merged together, which commands we run can be treated as an implementation detail.

rhc.spec
m-horky: I don't think we can suppress it. We can remove that line, and apply it back when we have patches downstream. It's just a warning, it shouldn't matter. I don't have a strong opinion, we could drop it.

rhc.spec
jirihnidek: No, it is OK. I was just currious.

rhc.spec
jirihnidek: What about this:

https://rpm-software-management.github.io/rpm/manual/autosetup.html

```
-q don’t warn if there are no matching patches
```

I tried it and it works as expected.

jirihnidek (changes requested):
Thanks for updates. I have one final requests that can suppress unnecessary warning.

rhc.spec
m-horky: I should have read the docs... Thanks, updated.

m-horky: It's not supported on RHEL 10 yet. I have removed it again, the if statement in the spec file didn't feel worth it.

jirihnidek (dismissed):
LGTM

m-horky:
> Failed to load packit config file

_sigh_ `packit config validate .` claims `.packit.yaml is valid and ready to be used`.

Edit: https://github.com/packit/packit/issues/1554

jirihnidek (approved):
LGTM

PR #483, opened 2026-06-10
fix(docs): Update testing readme for local testing

stephenbourke:
- Update unofficial Candlepin image location
 - Update insights-client mock to reflect what is included in
    systemtest/tests/test.sh
 - Fix pip install command
 - Ignore pycache and artifact files left from integration tests

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,5 @@ settings.toml
 .secrets.toml
 /rhc*.tar.*
 /vendor/
+/artifacts/
+__pycache__/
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -15,3 +15,5 @@ settings.toml
 .secrets.toml
 /rhc*.tar.*
 /vendor/
+artifacts/
```
Archana-PandeyM: LGTM! 
If the target is only top-level test output, use /artifacts/ instead.

Archana-PandeyM (approved):

PR #484, opened 2026-06-11
chore: Drop com.redhat.rhc.internal Varlink API

m-horky:
* Card ID: CCT-2601

The com.redhat.rhc.internal namespace was originally introduced to act as a target for the Varlink server development. With two other namespaces existing now, it is no longer required.


mjcr99 (approved):
LGTM, thanks!

PR #485, opened 2026-06-15
fix(rhc-collector): Fix failing on timer cache write errors

zpetrace:
Timer cache writes are best-effort metadata for last-run timing. Log cache write failures but let successful collector runs complete normally.


stephenbourke (approved):
LGTM

PR #486, opened 2026-06-15
fix(rhc-collector): execute collector binary instead of shell command

mjcr99:
## Description 

* Card ID: CCT-2535

When rhc-collector was triggered to start data collection, it passed the command argument to /bin/sh instead of executing the actual collector binary.

Changes:
- Swap argument order to COMMAND COLLECTOR-ID (was COLLECTOR-ID COMMAND)
- Validate command is "run" before proceeding
- Execute /usr/libexec/rhc/collectors/{COLLECTOR-ID} with "collect" subcommand
- Pass tmpDir as argument to the collector binary
- Update executeCollector signature to accept collectorId instead of command
- Simplify tests to focus on binary execution validation

Assisted-by: Claude Code

### DIFF
--- a/cmd/rhc-collector/main.go
+++ b/cmd/rhc-collector/main.go
@@ -23,10 +23,10 @@ const (
 
 func main() {
 	if len(os.Args) <= 2 {
-		slog.Error("usage: rhc-collector COLLECTOR-ID COMMAND")
+		slog.Error("usage: rhc-collector COMMAND COLLECTOR-ID")
 		os.Exit(exitcode.Usage)
 	}
-	collectorId, command := os.Args[1], os.Args[2]
+	command, collectorId := os.Args[1], os.Args[2]
 	slog.Info("starting rhc-collector", slog.String("id", collectorId))
 	if err := run(collectorId, command); err != nil {
 		slog.Error("rhc-collector exited with error", "error", err)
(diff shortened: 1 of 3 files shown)


cmd/rhc-collector/main.go
```diff
@@ -35,17 +34,22 @@ func main() {
 }
 
 func run(collectorId, command string) error {
-	config, err := getConfig(collectorId)
-	if err != nil {
-		return err
+	// Validate command
```
stephenbourke: It might be worth validating the collectorId too. It is done elsewhere and might help prevent things like path traversal in the name etc (`../`)?

e.g:

```
	collectorId, err := collector.ValidateID(collectorId)
	if err != nil {
		slog.Error("invalid collector ID", "error", err)
		return fmt.Errorf("invalid collector ID: %w", err)
	}
```

cmd/rhc-collector/main.go
```diff
@@ -112,43 +116,20 @@ func getConfig(collectorId string) (collector.Config, error) {
 	return config, nil
 }
 
-// executeCollector runs the specified collector command and stores output in tmpDir.
+// executeCollector runs the specified collector binary and stores output in tmpDir.
```
stephenbourke: Maybe mention the parameter used too?

" ... runs the specified collector binary with the collect argument and ..."

cmd/rhc-collector/main.go
mjcr99: Addressed, good point.

cmd/rhc-collector/main.go
mjcr99: Addressed.

stephenbourke (approved):
LGTM

pkoprda (approved):
ACK

PR #488, opened 2026-06-16
test: switch integration tests to use activation-key

Archana-PandeyM:
*Card Id : CCT-2577

Basic auth is not supported on satellite, causing these integration tests to fail in satellite environments. Replacing those cases with activation-key authentication so the tests are satellite-compatible and expand effective coverage on satellite runs.


stephenbourke (approved):
Looks good. 

Just one question. Can the `test_connect` test also be updated to use the activation-key instead of user/pass? I see that it skips Satellite.

PR #491, opened 2026-06-17
fix: Update completion flag for urfave/cli v3

Peto-RH:
Updated bash completion script to use --generate-shell-completion instead of deprecated --generate-bash-completion flag after migration to urfave/cli v3.


stephenbourke (approved):
Change looks good!

PR #496, opened 2026-06-24
feat(collector): add minimal collector

pkoprda:
* Card ID: CCT-2393

Add minimal data collector that generates HBI ingress message format with canonical facts only (not full insights archive).

---
Testing:
1. Install rpm
2. Register to subscription-manager
3. Run `/usr/libexec/rhc/rhc-collector run com.redhat.minimal`
4. Login to console.redhat.com and find your system - it should have the data obtained by the collector (instead of by executing insights-client)

Note: The 4th step doesn't entirely work yet and I don't know why that happens. The archive is being correctly uploaded to Ingress, but the facts are not visible in console.redhat.com.

### DIFF
--- a/cmd/minimal-collector/minimal-collector.go
+++ b/cmd/minimal-collector/minimal-collector.go
@@ -1,0 +1,581 @@
+package main
+
+import (
+	"cmp"
+	"crypto/x509"
+	"encoding/json"
+	"encoding/pem"
+	"fmt"
+	"log/slog"
+	"net"
+	"os"
+	"path/filepath"
+	"slices"
+	"strings"
+
+	"github.com/redhatinsights/rhc/internal/canonical_facts"
+	"github.com/redhatinsights/rhc/pkg/version"
+)
+
+// Satellite management constants. Both set to -1 to indicate "not managed by Satellite".
+// When puptoo processes branch_info, it checks: if remote_branch != -1, it sets
+// satellite_managed=true and satellite_id=remote_leaf.
+const (
+	remoteBranch = -1
+	remoteLeaf   = -1
+)
+
+// MetadataSpec represents metadata for a collected spec that is written to
+// the meta_data/ directory in insights archives. These metadata files are
+// required by insights-core's hydration process (serde.py) to load and parse
+// the archive data. The metadata format follows the structure created by
+// insights-client when it collects data using insights-core's spec framework.
+type MetadataSpec struct {
+	// Name is the fully qualified spec name in insights-core's namespace.
+	Name string `json:"name"`
+
+	// ExecTime is the execution time in seconds for collecting this spec.
+	// For minimal collector, use a placeholder value like 0.0001.
+	ExecTime float64 `json:"exec_time"`
+
+	// Errors is a list of error messages encountered during collection.
+	Errors []string `json:"errors"`
+
+	// Results contains the spec type and location information.
+	// Structure: {"type": "<provider_type>", "object": {"relative_path": "...", ...}}
+	// Provider types:
+	//   - insights.core.spec_factory.TextFileProvider (for file content)
+	//   - insights.core.spec_factory.CommandOutputProvider (for command output)
+	//   - insights.core.spec_factory.DatasourceProvider (for computed data)
+	// The "relative_path" field points to the actual data file in data/ directory.
+	Results any `json:"results"`
+
+	// SerTime is the serialization time in seconds for this spec's results.
+	// For minimal collector, use a placeholder value like 0.0001.
+	SerTime float64 `json:"ser_time"`
+}
+
+// SpecObject represents the "object" field within MetadataSpec.Results.
+// This structure describes how insights-core should locate and process the data file.
+type SpecObject struct {
+	// RelativePath is the path to the data file relative to the data/ directory.
+	RelativePath string `json:"relative_path"`
+
+	// SaveAs specifies an alternate path to save the file during collection.
+	// For minimal collector, always nil.
+	SaveAs any `json:"save_as"`
+
+	// RC is the command exit code (for CommandOutputProvider) or nil (for file-based providers).
+	// When a command is executed during collection, this stores the exit code.
+	// For minimal collector, always nil.
+	RC any `json:"rc"`
+
+	// Cmd is the original command string (for CommandOutputProvider only).
+	// For minimal collector, always nil.
+	Cmd any `json:"cmd"`
+
+	// Args are the command arguments (for CommandOutputProvider only).
+	// For minimal collector, always nil.
+	Args any `json:"args"`
+}
+
+// VersionInfo represents the version_info file content.
+type VersionInfo struct {
+	CoreVersion   string  `json:"core_version"`
+	ClientVersion *string `json:"client_version"`
+}
+
+// BranchInfo represents the branch_info file content.
+type BranchInfo struct {
+	RemoteBranch int `json:"remote_branch"`
+	RemoteLeaf   int `json:"remote_leaf"`
+}
+
+// createArchiveStructure creates an insights-core compatible archive structure.
+func createArchiveStructure(outputDir string) (string, string, error) {
+	dataDir := filepath.Join(outputDir, "data")
+	metaDataDir := filepath.Join(outputDir, "meta_data")
+
+	if err := os.MkdirAll(dataDir, 0700); err != nil {
+		return "", "", fmt.Errorf("failed to create data directory: %w", err)
+	}
+	if err := os.MkdirAll(metaDataDir, 0700); err != nil {
+		return "", "", fmt.Errorf("failed to create meta_data directory: %w", err)
+	}
+	// insights_archive.txt is an empty file used by insights-core to identify this as
+	// a SerializedArchiveContext. Its presence indicates this is an insights archive.
+	archiveMarker := filepath.Join(outputDir, "insights_archive.txt")
+	if err := os.WriteFile(archiveMarker, []byte{}, 0600); err != nil {
+		return "", "", fmt.Errorf("failed to write insights_archive.txt: %w", err)
+	}
+
+	return dataDir, metaDataDir, nil
+}
+
+// getOrgID retrieves the organization ID from the subscription-manager certificate.
+// The org_id is stored in the Organization field (O) of the certificate Subject.
+func getOrgID() (string, error) {
+	certPath := "/etc/pki/consumer/cert.pem"
+	data, err := os.ReadFile(certPath)
+	if err != nil {
+		slog.Error("failed to read consumer cert", "path", certPath, "error", err)
+		return "", fmt.Errorf("failed to read consumer cert: %w", err)
+	}
+
+	block, _ := pem.Decode(data)
+	if block == nil {
+		slog.Error("failed to decode PEM data from consumer cert", "path", certPath)
+		return "", fmt.Errorf("failed to decode PEM data from %s", certPath)
+	}
+
+	cert, err := x509.ParseCertificate(block.Bytes)
+	if err != nil {
+		slog.Error("failed to parse certificate from consumer cert", "path", certPath, "error", err)
+		return "", fmt.Errorf("failed to parse certificate: %w", err)
+	}
+
+	if len(cert.Subject.Organization) == 0 {
+		slog.Error("failed to parse organization from consumer cert", "path", certPath)
+		return "", fmt.Errorf("failed to parse organization from %s", certPath)
+	}
+
+	orgID := cert.Subject.Organization[0]
+	slog.Debug("found org ID from certificate", "org_id", orgID)
+	return orgID, nil
+}
+
+// writeVersionInfo creates version_info file and metadata.
+//
+// The version_info file contains collector version information that Puptoo extracts
+// and includes in the system profile sent to HBI:
+//   - core_version → system_profile.insights_egg_version
+//   - client_version → system_profile.insights_client_version
+//
+// These fields help track which collector version generated the archive, useful for
+// debugging data collection issues and understanding feature availability across hosts.
+// For minimal collector, core_version is set to rhc's version and client_version is null
+// since there's no separate insights-client component.
+func writeVersionInfo(dataDir, metaDataDir string) error {
+	versionInfo := VersionInfo{
+		CoreVersion:   version.Version,
+		ClientVersion: nil,
+	}
+
+	versionData, err := json.Marshal(versionInfo)
+	if err != nil {
+		return fmt.Errorf("failed to marshal version_info: %w", err)
+	}
+
+	versionPath := filepath.Join(dataDir, "version_info")
+	if err = os.WriteFile(versionPath, versionData, 0600); err != nil {
+		return fmt.Errorf("failed to write version_info: %w", err)
+	}
+
+	// Write metadata
+	metadata := MetadataSpec{
+		Name:     "insights.specs.Specs.version_info",
+		ExecTime: 0.0001,
+		Errors:   []string{},
+		Results: map[string]any{
+			"type": "insights.core.spec_factory.DatasourceProvider",
+			"object": SpecObject{
+				RelativePath: "version_info",
+				SaveAs:       nil,
+				RC:           nil,
+				Cmd:          nil,
+				Args:         nil,
+			},
+		},
+		SerTime: 0.0001,
+	}
+
+	metaData, err := json.Marshal(metadata)
+	if err != nil {
+		return fmt.Errorf("failed to marshal version_info metadata: %w", err)
+	}
+
+	metaPath := filepath.Join(metaDataDir, "insights.specs.Specs.version_info.json")
+	if err := os.WriteFile(metaPath, metaData, 0600); err != nil {
+		return fmt.Errorf("failed to write version_info metadata: %w", err)
+	}
+
+	return nil
+}
+
+// writeBranchInfo creates branch_info file and metadata. Puptoo uses this to
+// determine Satellite management: if remote_branch != -1, sets satellite_managed=true
+// and satellite_id=remote_leaf. For minimal collector, both values are -1.
+func writeBranchInfo(dataDir, metaDataDir string) error {
+	branchInfo := BranchInfo{
+		RemoteBranch: remoteBranch,
+		RemoteLeaf:   remoteLeaf,
+	}
+
+	branchData, err := json.Marshal(branchInfo)
+	if err != nil {
+		return fmt.Errorf("failed to marshal branch_info: %w", err)
+	}
+
+	branchPath := filepath.Join(dataDir, "branch_info")
+	if err := os.WriteFile(branchPath, branchData, 0600); err != nil {
+		return fmt.Errorf("failed to write branch_info: %w", err)
+	}
+
+	// Write metadata
+	metadata := MetadataSpec{
+		Name:     "insights.specs.Specs.branch_info",
+		ExecTime: 0.0001,
+		Errors:   []string{},
+		Results: map[string]any{
+			"type": "insights.core.spec_factory.DatasourceProvider",
+			"object": SpecObject{
+				RelativePath: "branch_info",
+				SaveAs:       nil,
+				RC:           nil,
+				Cmd:          nil,
+				Args:         nil,
+			},
+		},
+		SerTime: 0.0001,
+	}
+
+	metaData, err := json.Marshal(metadata)
+	if err != nil {
+		return fmt.Errorf("failed to marshal branch_info metadata: %w", err)
+	}
+
+	metaPath := filepath.Join(metaDataDir, "insights.specs.Specs.branch_info.json")
+	if err := os.WriteFile(metaPath, metaData, 0600); err != nil {
+		return fmt.Errorf("failed to write branch_info metadata: %w", err)
+	}
+
+	return nil
+}
+
+// writeCanonicalFacts collects and writes canonical facts to the archive.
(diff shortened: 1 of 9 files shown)


m-horky:
That is an internal Kafka format for Inventory. We need to send the message in a format Ingress and Puptoo accept.

m-horky (changes requested):
Let's look at the code once the workflow is functional.

data/collectors/com.redhat.minimal.toml
```diff
@@ -0,0 +1,9 @@
+[meta]
+name = "Minimal Host Inventory Collector"
+feature = "analytics"
+type = "ingress"
+
+[ingress]
+user = "root"
+group = "root"
+content_type = "application/vnd.redhat.advisor.collection"
```
pkoprda: What should be this content_type? `application/vnd.redhat.advisor.minimal` also works. Validation in insights-ingress-go says, it has to start with prefix `application/vnd.redhat.`: 
https://github.com/RedHatInsights/insights-ingress-go/blob/d230bb2b0bccb12f9860c52e43544dc74b1a9cc9/internal/upload/validation.go#L10

m-horky: I'm in favor of setting different category field :+1:

rhc.spec
```diff
@@ -77,6 +77,7 @@ export GO_LDFLAGS="-X github.com/redhatinsights/rhc/pkg/version.Version=%{versio
 %gobuild -o %{gobuilddir}/bin/rhc           %{goipath}/cmd/rhc
 %gobuild -o %{gobuilddir}/bin/rhc-server    %{goipath}/cmd/rhc-server
 %gobuild -o %{gobuilddir}/bin/rhc-collector %{goipath}/cmd/rhc-collector
+%gobuild -o %{gobuilddir}/bin/minimal-collector %{goipath}/cmd/minimal-collector
```
m-horky: What would be the pros/cons of building it as com.redhat.minimal here already? Non-blocking.

cmd/minimal-collector/minimal_collector.go
```diff
@@ -0,0 +1,485 @@
...
+	"os"
+	"path/filepath"
+	"sort"
+	"strings"
+
+	"github.com/redhatinsights/rhc/internal/canonical_facts"
+)
+
+const (
+	coreVersion  = "3.99.0"
```
m-horky: How about we report the version of rhc here? We're not pretending we are insights-core.

cmd/minimal-collector/minimal_collector.go
```diff
@@ -0,0 +1,485 @@
...
+	remoteBranch = -1
+	remoteLeaf   = -1
+)
+
+// MetadataSpec represents metadata for a collected spec.
```
m-horky: Could you expand this docstring a bit, describing how it is used, which code/component cares about it? I'd also see the individual fields documented.

cmd/minimal-collector/minimal-collector.go
```diff
@@ -0,0 +1,485 @@
...
+	if err := os.MkdirAll(dataDir, 0700); err != nil {
+		return "", "", fmt.Errorf("failed to create data directory: %w", err)
+	}
+	if err := os.MkdirAll(metaDataDir, 0700); err != nil {
+		return "", "", fmt.Errorf("failed to create meta_data directory: %w", err)
+	}
+	archiveMarker := filepath.Join(outputDir, "insights_archive.txt")
+	if err := os.WriteFile(archiveMarker, []byte{}, 0644); err != nil {
+		return "", "", fmt.Errorf("failed to write insights_archive.txt: %w", err)
+	}
```
m-horky, line 110: What does the file otherwise contain? Could we add an inline comment about it?

cmd/minimal-collector/minimal-collector.go
```diff
@@ -0,0 +1,485 @@
...
+
+	return dataDir, metaDataDir, nil
+}
+
+// getOrgID retrieves the organization ID from the subscription-manager certificate.
+// The org_id is stored in the Organization field (O) of the certificate Subject.
+func getOrgID() (string, error) {
```
m-horky, line 117: Comment: Long-term, there should be an `internal/` package that parses certificates. This isn't really a frontend concern. However, since this is _supposed to be temporary_, and full functionality is contained within `cmd/`, I won't let it block the review, it might actually be the better way to handle it.

cmd/minimal-collector/minimal-collector.go
```diff
@@ -0,0 +1,485 @@
...
+		slog.Error("failed to parse organization from consumer cert", "path", certPath)
+		return "", fmt.Errorf("failed to parse organization from %s", certPath)
+	}
+
+	orgID := cert.Subject.Organization[0]
+	slog.Debug("found org ID from certificate", "org_id", orgID)
+	return orgID, nil
+}
+
+// writeVersionInfo creates version_info file and metadata.
```
m-horky, line 147: Could you document how Puptoo uses this artifact? I am aware this is effectively an internal Advisor API, but I'd like to spend some time documenting the reasons for this method's (and the JSON file's) existence.

cmd/minimal-collector/minimal_collector.go
```diff
@@ -0,0 +1,485 @@
...
+
+	metaPath := filepath.Join(metaDataDir, "insights.specs.Specs.version_info.json")
+	if err := os.WriteFile(metaPath, metaData, 0644); err != nil {
+		return fmt.Errorf("failed to write version_info metadata: %w", err)
+	}
+
+	return nil
+}
+
+// writeBranchInfo creates branch_info file and metadata.
```
m-horky: ditto

cmd/minimal-collector/minimal_collector.go
```diff
@@ -0,0 +1,485 @@
...
+	metadata := MetadataSpec{
+		Name:     "insights.specs.Specs.machine_id",
+		ExecTime: 0.0001,
+		Errors:   []string{},
+		Results: map[string]interface{}{
+			"type": "insights.core.spec_factory.TextFileProvider",
+			"object": map[string]interface{}{
+				"save_as":       false,
+				"relative_path": "etc/insights-client/machine-id",
+				"rc":            nil,
```
m-horky: Could we define this as a struct as well? I'm interested why `save_as` and `rc` are the values they are.

cmd/minimal-collector/minimal_collector.go
```diff
@@ -0,0 +1,485 @@
...
+		ExecTime: 0.0001,
+		Errors:   []string{},
+		Results: map[string]interface{}{
+			"type": "insights.core.spec_factory.CommandOutputProvider",
+			"object": map[string]interface{}{
+				"rc":            nil,
+				"cmd":           nil,
+				"args":          nil,
+				"save_as":       false,
+				"relative_path": "insights_commands/subscription-manager_identity",
```
m-horky: Ditto, I'd like to see this as a struct that describes the APIs it needs to fulfill.

m-horky (changes requested):
I'm happy with the patch, it looks reasonable and maintainable. I have some non-critical requests inline. How do we plan to test this? Will we have unit/integration tests?

cmd/minimal-collector/minimal_collector.go
```diff
@@ -0,0 +1,485 @@
...
+// writeSubscriptionManagerID writes subscription-manager identity output.
+func writeSubscriptionManagerID(commandsDir, metaDataDir, subMgrID string) error {
+	orgID, err := getOrgID()
+	if err != nil {
+		slog.Warn("failed to get org_id for subscription-manager output", "error", err)
+		orgID = "unknown"
+	}
+
+	output := fmt.Sprintf("system identity: %s\nname: %s\norg name: %s\norg ID: %s\n",
+		subMgrID, "system", orgID, orgID)
```
m-horky: Are we hardcoding 'system' as the hostname, instead of passing in the actual system hostname?

cmd/minimal-collector/minimal-collector.go
```diff
@@ -0,0 +1,485 @@
...
+		SerTime: 0.0001,
+	}
+
+	return writeMetadata(metaDataDir, "insights.specs.Specs.ip_addresses.json", metadata)
+}
+
+// writeMACAddresses writes MAC addresses metadata (no data file needed).
+func writeMACAddresses(metaDataDir string) error {
+	// Get network interfaces to match MAC addresses with interface names
+	ifaces, err := net.Interfaces()
```
m-horky, line 423: Could we use the `internal/canonical_facts` package to get a list of MACs?

cmd/minimal-collector/minimal-collector.go
```diff
@@ -0,0 +1,485 @@
...
+		ClientVersion: nil,
+	}
+
+	versionData, err := json.Marshal(versionInfo)
+	if err != nil {
+		return fmt.Errorf("failed to marshal version_info: %w", err)
+	}
+
+	versionPath := filepath.Join(dataDir, "version_info")
+	if err = os.WriteFile(versionPath, versionData, 0644); err != nil {
```
m-horky: Should all file permissions have permissions of `0600` instead of `0644`? The parent directory has `0700`.

cmd/minimal-collector/minimal_collector.go
```diff
@@ -0,0 +1,485 @@
...
+
+	return writeMetadata(metaDataDir, "insights.specs.Specs.machine_id.json", metadata)
+}
+
+// writeSubscriptionManagerID writes subscription-manager identity output.
+func writeSubscriptionManagerID(commandsDir, metaDataDir, subMgrID string) error {
+	orgID, err := getOrgID()
+	if err != nil {
+		slog.Warn("failed to get org_id for subscription-manager output", "error", err)
+		orgID = "unknown"
```
m-horky: I think we should error out in case we aren't capable of reading the consumer certificate.

cmd/minimal-collector/minimal_collector.go
```diff
@@ -0,0 +1,485 @@
...
+				"relative_path": "insights_commands/hostname_-I",
+			},
+		},
+		SerTime: 0.0001,
+	}
+
+	return writeMetadata(metaDataDir, "insights.specs.Specs.ip_addresses.json", metadata)
+}
+
+// writeMACAddresses writes MAC addresses metadata (no data file needed).
```
m-horky: > no data file needed

Really? The function surely does not create `data/sys/class/net/eth0/address`, but the results contain `{"type": "insights.core.spec_factory.TextFileProvider", "object": {"relative_path": "sys/class/net/eth0/address"}}`.

m-horky (changes requested):

rhc.spec
pkoprda: No cons, I will change it to `com.redhat.minimal`

cmd/minimal-collector/minimal-collector.go
pkoprda: The problem is that canonical_facts only returns MAC address values (the strings), but to write the archive files we need the interface names (like `eth0`,  `wlp9s0f0`, etc.) to create paths like `sys/class/net/eth0/address`. So this package is not really useful here.

cmd/minimal-collector/minimal-collector.go
pkoprda: I wasn't sure how the permissions should be set. So I defined the permissions for directories as `0700`, but probably I forgot to change the permissions for files.

I will change the permissions for files to `0600`, but let me know, if the permissions for directories should be `0755` for directories and `0644` for files.

cmd/minimal-collector/minimal-collector.go
m-horky: Aah, I see.

cmd/minimal-collector/minimal_collector.go
pkoprda: Fixed:
```
if err != nil {
	slog.Warn("failed to get hostname for subscription-manager output", "error", err)
	hostname = "unknown"
}

output := fmt.Sprintf("system identity: %s\nname: %s\norg name: %s\norg ID: %s\n", subMgrID, hostname, orgID, orgID)
```

pkoprda:
> How do we plan to test this? Will we have unit/integration tests?

I've added some unit tests, but I would rather not create integration tests for this in this PR. It already has around 1000 lines of code change.

cmd/minimal-collector/minimal_collector.go
pkoprda: You are right, I've removed the text in parentheses.

cmd/minimal-collector/minimal-collector.go
```diff
@@ -0,0 +1,549 @@
...
+	})
+
+	// Build results array with one entry per interface
+	results := make([]map[string]interface{}, 0, len(ifaces))
+	for _, iface := range ifaces {
+		results = append(results, map[string]interface{}{
+			"type": "insights.core.spec_factory.TextFileProvider",
+			"object": map[string]interface{}{
+				"save_as":       false,
+				"relative_path": fmt.Sprintf("sys/class/net/%s/address", iface.Name),
```
stephenbourke: I don't see a filepath.join being used like the other functions to create the directory structure. Will this path exist?

pkoprda: Probably not, I added this so the directory structure should exist:
```go
ifaceDir := filepath.Join(dataDir, "sys", "class", "net", iface.Name)
if err := os.MkdirAll(ifaceDir, 0700); err != nil {
	slog.Error("failed to create interface directory", "path", ifaceDir, "error", err)
	return fmt.Errorf("failed to create directory %s: %w", ifaceDir, err)
}

macPath := filepath.Join(ifaceDir, "address")
macAddr := iface.HardwareAddr.String()
if err := os.WriteFile(macPath, []byte(macAddr+"\n"), 0600); err != nil {
	slog.Error("failed to write MAC address to file", "path", macPath, "error", err)
	return fmt.Errorf("failed to write MAC address to %s: %w", macPath, err)
}
```

cmd/minimal-collector/minimal_collector.go
pkoprda: I change it to SpecObject, but then I revert it (see the last change), because it didn't work correctly. Not sure what is the reason why it didn't work correctly :thinking:

pkoprda: Nevermind, I found the reason. The problem was when I explicitly set `RC: nil`, the `omitempty` tag on the `SpecObjects` fields causes it to be omitted from the JSON output because a nil interface{} is considered as "empty".

So now it should work correctly with SpecObjects

cmd/minimal-collector/minimal-collector.go
```diff
@@ -0,0 +1,575 @@
...
+
+	if facts.InsightsID != "" {
+		if err = writeInsightsID(etcDir, metaDataDir, facts.InsightsID); err != nil {
+			return err
+		}
+	}
+
+	if err = writeSubscriptionManagerID(commandsDir, metaDataDir, facts.SubscriptionManagerID); err != nil {
+		return err
+	}
```
m-horky: Is there a reason this is the single `write*` call that fails whole operation if it returns an error? The current code means you can't debug the collection if the host isn't registered, right?

cmd/minimal-collector/minimal-collector.go
```diff
@@ -0,0 +1,575 @@
...
+
+	if len(facts.IPAddresses) > 0 {
+		if err = writeIPAddresses(commandsDir, metaDataDir, facts.IPAddresses); err != nil {
+			return err
+		}
+	}
+
+	if len(facts.MACAddresses) > 0 {
+		if err = writeMACAddresses(dataDir, metaDataDir); err != nil {
```
m-horky: Is there a reason `writeMACAddresses` collects the data anew and isn't using `facts.MACAddresses`?

cmd/minimal-collector/minimal-collector.go
m-horky: I'd go with 0600 / 0700, probably.

cmd/minimal-collector/minimal-collector.go
```diff
@@ -0,0 +1,575 @@
...
+	}
+
+	return nil
+}
+
+// writeCanonicalFacts collects and writes canonical facts to the archive.
+func writeCanonicalFacts(dataDir, metaDataDir string) error {
+	facts, err := canonical_facts.GetCanonicalFacts()
+	if err != nil {
+		slog.Error("failed to get canonical facts", "path", dataDir, "error", err)
```
m-horky: Isn't the `dataDir` irrelevant to the error being logged?

cmd/minimal-collector/minimal-collector_test.go
```diff
@@ -0,0 +1,396 @@
...
+}
+
+func TestWriteMetadata(t *testing.T) {
+	tmpDir := t.TempDir()
+
+	metadata := MetadataSpec{
+		Name:     "test.spec",
+		ExecTime: 0.0001,
+		Errors:   []string{},
+		Results: map[string]interface{}{
```
m-horky: Can we use `any` instead of `interface{}`?

cmd/minimal-collector/minimal-collector.go
```diff
@@ -0,0 +1,575 @@
...
+
+// SpecObject represents the "object" field within MetadataSpec.Results.
+// This structure describes how insights-core should locate and process the data file.
+type SpecObject struct {
+	// RelativePath is the path to the data file relative to the data/ directory.
+	RelativePath string `json:"relative_path"`
+
+	// SaveAs specifies an alternate path to save the file during collection.
+	// For minimal collector archives (already collected), this is always nil or false.
+	SaveAs interface{} `json:"save_as"`
```
m-horky: > this is always nil or false

Could we make it a bool and set `false`? I'm probably not understanding the meaning of this field.

cmd/minimal-collector/minimal-collector.go
```diff
@@ -0,0 +1,575 @@
...
+	ifaces, err := net.Interfaces()
+	if err != nil {
+		slog.Error("failed to get network interfaces", "error", err)
+		return fmt.Errorf("failed to get network interfaces: %w", err)
+	}
+
+	// Sort interfaces by name for consistent ordering
+	sort.Slice(ifaces, func(i, j int) bool {
+		return ifaces[i].Name < ifaces[j].Name
+	})
```
m-horky: `slices.SortFunc` + `cmp.Compare`?

cmd/minimal-collector/minimal-collector.go
```diff
@@ -0,0 +1,575 @@
...
+	}
+
+	return nil
+}
+
+// writeInsightsID writes the insights-client machine-id.
+func writeInsightsID(etcDir, metaDataDir, insightsID string) error {
+	path := filepath.Join(etcDir, "machine-id")
+	if err := os.WriteFile(path, []byte(insightsID), 0644); err != nil {
+		slog.Error("failed to write insights machine-id to file", "path", path, "err", err)
```
m-horky: Could you replace `"err"` with `"error"` for consistency?

cmd/minimal-collector/minimal-collector.go
```diff
@@ -0,0 +1,575 @@
...
+	"strings"
+
+	"github.com/redhatinsights/rhc/internal/canonical_facts"
+	"github.com/redhatinsights/rhc/pkg/version"
+)
+
+const (
+	remoteBranch = -1
+	remoteLeaf   = -1
+)
```
m-horky, line 26: Can we add a docstring to explain `-1` represents "Not managed by Satellite"?

m-horky (changes requested):
We're getting there :)

cmd/minimal-collector/minimal-collector.go
pkoprda: Yes, see my comment above: https://github.com/RedHatInsights/rhc/pull/496#discussion_r3481730924

cmd/minimal-collector/minimal-collector.go
pkoprda: That wouldn't make sense at all, since the expected value is path. I updated the comment, for minimal collector this value is always nil.

cmd/minimal-collector/minimal-collector.go
pkoprda: I added a condition before: `if facts.SubscriptionManagerID != ""`

cmd/minimal-collector/minimal-collector.go
m-horky: - Should we ignore the field gathered by canonical facts and spin our own then?
- Should we extended the canonical facts collector to gather it as well, and continue not writing it down when triggered with `canonical-facts` command?

pkoprda: > Should we ignore the field gathered by canonical facts and spin our own then?

This is already implemented, see https://github.com/RedHatInsights/rhc/pull/496#discussion_r3492125209. I would just remove the if condition here so it would be this:
```go
if err = writeMACAddresses(dataDir, metaDataDir); err != nil {
    return err
}
```
instead of this:
```go
if len(facts.MACAddresses) > 0 {
    if err = writeMACAddresses(dataDir, metaDataDir); err != nil {
        return err
    }
}
```
Is that okay?

m-horky: My concern was inconsistency -- some `write*` methods take an argument, some don't. It feels wrong from data flow management perspective. But the impact is fairly limited, I don't want to block the PR on it.

m-horky (dismissed):
ACK, all things considered. You may resolve the last thread we have.

stephenbourke (approved):
LGTM

PR #497, opened 2026-06-24
chore(subman): Refactor system registration

m-horky:
* Card ID: CCT-2401

Following previous commit starting the internal/subman refactor, this patch adds system registration functionality. It moves away from direct D-Bus calls into a structure more readable and concern-separated.

Since original rhsm.go and connect_cmd.go files mix and match layers, the code still isn't plugged into the actual CLI.

errors.go
- Defines several sentinel values the API consumers can match on.
- Makes D-Bus error handling private, isolating it to use within the package.
- newDbusError() is less nested than the original UnpackDBusError().

dbus.go
- bus() centralizes SystemBus acquisition and documents the singleton nature of the Go D-Bus library we use.
- withPrivateRegisterSocket() manages the details of connecting to the private UNIX socket, and allows callers to focus on their own business logic of calling methods over that socket.

identity.go
- Defines IsRegistered() and GetOrganizations() as a new clean API for external callers.
- Unexports GetConsumerUUID(), since none of the surrounding code read its value.
- Defines RegisterOptions{} to wrap common arguments both password-based and activation-key-based registration options accept.
- Defines RegisterWithPassword() and RegisterWithActivationKey() as public methods with cleaner implementations. They use aforementioned withPrivateRegisterSocket() and named closures instead of deeply nested code that gets tricky to follow.

content.go
- Nothing really notable.
- Switches to positive-branch convention and sets content="1" when the 'enabled' value is true.

connect_cmd.go, status_cmd.go
- Use some of the new subman.* API.
- Due to the layer mixing in connect_cmd, most of the original code has been kept intact to limit disruptions.

### DIFF
--- a/internal/subman/content.go
+++ b/internal/subman/content.go
@@ -5,61 +5,62 @@ import (
 	"log/slog"
 
 	"github.com/godbus/dbus/v5"
-
 	"github.com/redhatinsights/rhc/internal/localization"
 )
 
-// IsContentManagementEnabled returns true if content management is enabled for the system in rhsm.conf
+// IsContentManagementEnabled reports whether content management is enabled for
+// the system in rhsm.conf (rhsm.manage_repos).
 func IsContentManagementEnabled() (bool, error) {
-	conn, err := dbus.SystemBus()
+	slog.Debug("Checking content management status")
+	conn, err := bus()
 	if err != nil {
-		return false, fmt.Errorf("cannot connect to system D-Bus: %w", err)
+		return false, err
 	}
-	// godbus implements SystemBus as a singleton, do not call conn.Close()
 
 	locale := localization.GetLocale()
 	config := conn.Object("com.redhat.RHSM1", "/com/redhat/RHSM1/Config")
 
-	var contentEnabled string
+	var value string
 	err = config.Call(
 		"com.redhat.RHSM1.Config.Get",
 		dbus.Flags(0),
 		"rhsm.manage_repos",
-		locale).Store(&contentEnabled)
+		locale,
+	).Store(&value)
 	if err != nil {
-		slog.Debug("Failed to get rhsm.manage_repos config", "error", err)
-		return false, UnpackDBusError(err)
+		return false, fmt.Errorf("reading content management setting: %w", newDbusError(err))
 	}
 
-	return contentEnabled == "1", nil
+	return value == "1", nil
 }
 
-// SetContentManagement tries to enable or disable content management for the system in rhsm.conf
+// SetContentManagement enables or disables content management for the system
+// in rhsm.conf (rhsm.manage_repos).
 func SetContentManagement(enabled bool) error {
-	conn, err := dbus.SystemBus()
+	slog.Debug("Setting content management", "enabled", enabled)
+	conn, err := bus()
 	if err != nil {
-		return fmt.Errorf("cannot connect to system D-Bus: %w", err)
+		return err
 	}
-	// godbus implements SystemBus as a singleton, do not call conn.Close()
 
 	locale := localization.GetLocale()
 	config := conn.Object("com.redhat.RHSM1", "/com/redhat/RHSM1/Config")
-	enabledStr := "1"
-	if !enabled {
-		enabledStr = "0"
+
+	value := "0"
+	if enabled {
+		value = "1"
 	}
 
 	err = config.Call(
 		"com.redhat.RHSM1.Config.Set",
 		dbus.Flags(0),
 		"rhsm.manage_repos",
-		enabledStr,
+		value,
 		locale,
 	).Err
 	if err != nil {
-		slog.Debug("Failed to set rhsm.manage_repos config", "error", err)
-		return UnpackDBusError(err)
+		slog.Debug("Could not set rhsm.manage_repos", "error", err)
+		return fmt.Errorf("setting content management: %w", newDbusError(err))
 	}
-
 	return nil
 }
--- a/internal/subman/errors.go
+++ b/internal/subman/errors.go
@@ -3,29 +3,41 @@ package subman
 import (
 	"encoding/json"
 	"errors"
-	"fmt"
 	"log/slog"
 
 	"github.com/godbus/dbus/v5"
 )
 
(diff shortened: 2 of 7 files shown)


internal/subman/content.go
```diff
@@ -5,61 +5,61 @@ import (
...
-// SetContentManagement tries to enable or disable content management for the system in rhsm.conf
+// SetContentManagement enables or disables content management for the system
+// in rhsm.conf (rhsm.manage_repos).
 func SetContentManagement(enabled bool) error {
-	conn, err := dbus.SystemBus()
+	slog.Debug("Setting content management", "enabled", enabled)
+	conn, err := bus()
 	if err != nil {
-		return fmt.Errorf("cannot connect to system D-Bus: %w", err)
+		return err
```
pkoprda, line 43: Why are you removing this error wrapping?

internal/subman/content.go
```diff
@@ -5,61 +5,61 @@ import (
...
 		dbus.Flags(0),
 		"rhsm.manage_repos",
-		enabledStr,
+		value,
 		locale,
 	).Err
 	if err != nil {
-		slog.Debug("Failed to set rhsm.manage_repos config", "error", err)
-		return UnpackDBusError(err)
+		return fmt.Errorf("setting content management: %w", newDbusError(err))
```
pkoprda, line 63: I would like to keep here some debug message

internal/subman/errors.go
```diff
@@ -3,29 +3,41 @@ package subman
...
-	return fmt.Sprintf("%v: %v", dbusError.Severity, dbusError.Message)
+func (e dbusError) Error() string {
+	return e.Message
 }
 
-// UnpackDBusError tries to unpack a JSON document (part of the error) into the structure DBusError.
-// When it is not possible to parse an error into structure, then a corresponding or original error
-// is returned. When it is possible to parse error into structure, then DBusError is returned
-func UnpackDBusError(err error) error {
+// newDbusError translates a raw D-Bus error into a richer dbusError when
```
pkoprda: I don't like the word "richer" here
```suggestion
// newDbusError translates a raw D-Bus error into a structured dbusError when
```

internal/subman/errors.go
```diff
@@ -35,10 +47,10 @@ func UnpackDBusError(err error) error {
...
 
-	var dbusError DBusError
-	if jsonErr := json.Unmarshal([]byte(e.Error()), &dbusError); jsonErr != nil {
+	var d dbusError
+	if jsonErr := json.Unmarshal([]byte(e.Error()), &d); jsonErr != nil {
 		slog.Debug("Failed to unmarshal D-Bus error body", "error", jsonErr)
 		return err
 	}
-	return dbusError
+	return d
```
pkoprda, line 55: The original variable was more descriptive :(

internal/subman/identity.go
```diff
@@ -51,34 +73,178 @@ func IsRegistered() (bool, error) {
...
+	slog.Debug("Registering system with username and password")
+	conn, err := bus()
+	if err != nil {
+		return err
+	}
+
+	uuid, err := getConsumerUUID(conn)
+	if err != nil {
+		return err
+	}
```
pkoprda, line 135: Shouldn't we use some error wrapping here?

internal/subman/identity.go
```diff
@@ -51,34 +73,178 @@ func IsRegistered() (bool, error) {
...
+
+	conn, err := bus()
+	if err != nil {
+		return err
+	}
+
+	uuid, err := getConsumerUUID(conn)
+	if err != nil {
+		return err
+	}
```
pkoprda, line 189: Ditto

internal/subman/identity.go
```diff
@@ -51,34 +73,178 @@ func IsRegistered() (bool, error) {
 	return false, nil
 }
 
-// Unregister removes the system registration from RHSM.
+// GetOrganizations returns the list of organization names available for the
+// given username and password.
+func GetOrganizations(username, password string) ([]string, error) {
```
pkoprda, line 78: This method doesn't seem to be used anywhere and also wasn't even mentioned in the Jira ticket

pkoprda (changes requested):
Thanks for the PR, I left some comments below

internal/subman/content.go
m-horky: The `bus()` method already provides the message and the error in case it goes wrong, all its users can return without adding more information or logging it.

https://github.com/RedHatInsights/rhc/blob/aaa2da39aa727ffc3176ed8f3f68a8b93d971d7d/internal/subman/errors.go#L12
https://github.com/RedHatInsights/rhc/blob/aaa2da39aa727ffc3176ed8f3f68a8b93d971d7d/internal/subman/dbus.go#L16-L20

internal/subman/errors.go
m-horky: With the object being named `dbusError`, do you have any specific preference? It's used three lines in a six line block, does it matter?

internal/subman/identity.go
m-horky: Let's use IsRegistered() instead. I've added error wrapping there.

internal/subman/identity.go
m-horky: It is needed due to a slightly different design pattern. The `internal/rhsm` interactively asks the user which organization they'd like to register to. We shouldn't be doing that, we should keep that logic in `cmd/`.

Instead, the two registration methods may raise `ErrOrganizationRequired`. The presentation layer should then call `GetOrganizations` to obtain their list, let the user choose one, and call `Register*` method again with the organization ID set.

internal/subman/content.go
pkoprda: Okay, let's keep it as it is now

internal/subman/errors.go
pkoprda: `dbusErr`? But it doesn't really matter, so I will keep decision up to you :)

pkoprda (changes requested):
Looks good, just please resolve the conflicts

pkoprda (approved):
LGTM

m-horky:
Thanks. I'll let the tests run before merging.

PR #498, opened 2026-06-24
chore(cmd): Clean up status

m-horky:
* Card ID: CCT-2404

- Status command code, as a presentation layer, should not be making direct D-Bus calls to obtain systemd service states.
- Low-level calls to spinner.Spinner in insightsStatus() should be using our abstraction internal/ui.Spinner() that also handles machine-readable output.
- rhc should not be displaying raw systemd messages, as they cannot be translated. Instead, it should present its own higher-level message that provides the user with information whether the state is fine or not. The integration tests for this have been updated.

### DIFF
--- a/cmd/rhc/status_cmd.go
+++ b/cmd/rhc/status_cmd.go
@@ -6,17 +6,11 @@ import (
 	"fmt"
 	"log/slog"
 	"os"
-	"reflect"
-	"time"
 
-	"github.com/godbus/dbus/v5"
 	"github.com/urfave/cli/v3"
 
-	"github.com/briandowns/spinner"
-	systemd "github.com/coreos/go-systemd/v22/dbus"
-
 	"github.com/redhatinsights/rhc/internal/datacollection"
-	"github.com/redhatinsights/rhc/internal/localization"
+	"github.com/redhatinsights/rhc/internal/remotemanagement"
 	"github.com/redhatinsights/rhc/internal/subman"
 	"github.com/redhatinsights/rhc/internal/ui"
 	"github.com/redhatinsights/rhc/pkg/exitcode"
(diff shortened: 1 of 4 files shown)


cmd/rhc/status_cmd.go
```diff
@@ -49,40 +43,26 @@ func rhsmStatus(systemStatus *SystemStatus) error {
...
-		return subman.UnpackDBusError(err)
+		return fmt.Errorf("unable to check content management: %w", err)
 	}
 
-	uuid, err := subman.GetConsumerUUID()
+	registered, err := subman.IsRegistered()
 	if err != nil {
 		systemStatus.returnCode += 1
 		systemStatus.ContentError = err.Error()
 		return fmt.Errorf("unable to get consumer UUID: %s", err)
```
stephenbourke: Is this error message still accurate? Possibly a more general error, for example "Unable to get registration status".

m-horky: Thanks for catching that, you are right, fixed. I think we'll need to revisit the errors in the future anyway, it's all kind of self-grown. At worse when we introduce localization we'll need to think about the presentation.

cmd/rhc/status_cmd.go
```diff
@@ -134,63 +113,35 @@ func insightStatus(systemStatus *SystemStatus) error {
...
+	if activeState == "active" {
 		systemStatus.YggdrasilRunning = true
 		infoMsg := "The yggdrasil service is active"
 		slog.Info(infoMsg)
 		ui.Printf("%s[%v] Remote Management ... %v\n", ui.Indent.Medium, ui.Icons.Ok, infoMsg)
 	} else {
 		systemStatus.returnCode += 1
-		loadState := properties["LoadState"]
-		if loadState == "loaded" {
+		if loadError == "" {
```
stephenbourke: This might be an unlikely event. It seems in the case where the unit is missing,  loadError can be empty. But the message below ("service is inactive but loaded") does not reflect this state.

cmd/rhc/status_cmd.go
stephenbourke: Makes sense. My other related comment below fits into the same category.

cmd/rhc/status_cmd.go
m-horky: Good point, that state might be misrepresented by the UI. I'll address that.

stephenbourke (approved):
LGTM

PR #499, opened 2026-06-25
feat: Add consumer varlink interface

cnsnyder:
* Card ID: CCT-2372
* This commit adds the com.redhat.rhsm.testing.consumer varlink api. The new api has three methods:
  - GetUUID -> Returns a string of the UUID from the local consumer certificate.
  - GetOrganization -> Returns a string of the organization from the local consumer certificate.
  - GetEnvironments -> Returns a (potentially nil or empty list) of Environment objects as returned from Candlepin using the installed consumer certificate.
Assisted-by: Cursor <noreply@cursor.com>

This PR needs: https://github.com/jirihnidek/rhsm2/pull/92
Instructions for testing prior to merge of the above are below.

These can be tested by:
0) Install subscription-manager, podman and remove any prior rhc:
* ```sudo dnf install subscription-manager podman```
* ```sudo dnf remove rhc```

1)Start a local candlepin:  ```sudo podman run --rm -it -d --name=candlepin --hostname=candlepin.local -p 8443:8443 -p 8080:8080 ghcr.io/ptoscano/candlepin-unofficial:latest```

2) Add an environment ```curl -k --user admin:admin --request POST \                                                                                  
  --header 'Content-Type: application/json' \
  --data '{"id":"env-id-01","name":"env-name-01","description":"Testing environment #1"}' \
  https://candlepin.local:8443/candlepin/owners/donaldduck/environments```

3) Add another environment ```curl -k --user admin:admin --request POST \                                                                                  
  --header 'Content-Type: application/json' \
  --data '{"id":"env-id-02","name":"env-name-02","description":"Testing environment #2"}' \
  https://candlepin.local:8443/candlepin/owners/donaldduck/environments```

4) Update rhsm.conf to point to the local CP (edit the beginning of /etc/rhsm/rhsm.conf to be as follows):
``` # Unified Entitlement Platform Configuration
[server]
# Server hostname:
hostname = candlepin.local

# Server prefix:
prefix = /candlepin

# Server port:
port = 8443
```
5) Register using subman: ```sudo subscription-manager register --username doc --password password --org donaldduck --environments env-name-01, env-name-02```

6) In a the same parent directory as your rhc repo, clone https://github.com/cnsnyder/rhsm2 and checkout branch 'csnyder/cct_2372'.

7) Add the following line to 'go.mod': ```replace github.com/jirihnidek/rhsm2 => ../rhsm2``` (this allows the local build of rhc to use the version of rhsm2 required for this PR to work).

8) Back in the local rhc checkout build off this branch: ```make clean && make rpm```

9) Install the rpm: ```sudo rpm -i --force /home/$USER/rpmbuild/RPMS/x86_64/rhc-0.3.9-1.fc44.x86_64.rpm``` (actual version of the rpm might be different, so check the output of make rpm to be sure)

10) Ensure the rhc-server.socket is running (this is started after install & reboot or isolating default.target but quicker to just start directly): ```sudo systemctl start rhc-server.socket```

Now you can test each of the methods as below:

* ``` sudo varlinkctl call /run/rhc/com.redhat.rhc com.redhat.rhsm.testing.consumer.GetUUID '{}' ```
* ``` sudo varlinkctl call /run/rhc/com.redhat.rhc com.redhat.rhsm.testing.consumer.GetOrganization '{}' ```
* ``` sudo varlinkctl call /run/rhc/com.redhat.rhc com.redhat.rhsm.testing.consumer.GetEnvironments '{}' ```


cmd/rhc-server/rhsm.go
jirihnidek, line 1: @cnsnyder Hi, it will be necessary to update go.mod file of `rhc`, because current `rhsm2` does not contain this new `GetConsumer()` functions. You can do it using: `go get github.com/jirihnidek/rhsm2`

cmd/rhc-server/rhc-server.go
```diff
@@ -115,3 +116,46 @@ func (b *RHSMBackend) IsRegistered(in *rhsmapi.IsRegisteredIn) (*rhsmapi.IsRegis
 	}
 	return &rhsmapi.IsRegisteredOut{Registered: registered}, nil
 }
+
+// RHSMBackend implements the rhsmapi.Backend interface.
+type ConsumerBackend struct{}
+
+// NewRHSMBackend creates a new RHSMBackend instance.
```
jirihnidek: This doc string is not correct. It should start with `// NewConsumerBackend ...`

cmd/rhc-server/rhc-server.go
```diff
@@ -115,3 +116,46 @@ func (b *RHSMBackend) IsRegistered(in *rhsmapi.IsRegisteredIn) (*rhsmapi.IsRegis
 	}
 	return &rhsmapi.IsRegisteredOut{Registered: registered}, nil
 }
+
+// RHSMBackend implements the rhsmapi.Backend interface.
```
jirihnidek: This doc string is not correct. It should start with `// ConsumerBackend ...`

cmd/rhc-server/rhsm.go
```diff
@@ -54,3 +56,66 @@ func IsSystemRegistered() (bool, error) {
...
+
+// GetConsumerOrg retrieves the organization key from the installed consumer certificate.
+func GetConsumerOrg() (*string, error) {
+	appName := AppName
+	rhsmClient, err := rhsm2.GetRHSMClient(&appName, nil)
+	if err != nil {
+		return nil, &ClientError{Message: err.Error()}
+	}
+
+	owner, err := rhsmClient.GetOwner()
```
jirihnidek: This is not correct. We need to return an object representing organization. There wasn't method for that in rhsm2. I opened PR for that: https://github.com/jirihnidek/rhsm2/pull/93

varlink/consumerapi/com.redhat.rhsm.testing.consumer.varlink
```diff
@@ -0,0 +1,21 @@
...
+)
+
+# Return organization that the registered system belongs to
+method GetOrganization() -> (
+    org: object
+)
+
+# Return environments assigned to the registered consumer
+method GetEnvironments() -> (
+    environments: ?[]object
```
jirihnidek: I believe that it should be only `environments: []object`. When there will be no environment, then we will just return empty list.

cmd/rhc-server/rhsm.go
```diff
@@ -54,3 +56,66 @@ func IsSystemRegistered() (bool, error) {
...
+	environmentsJSON, err := json.Marshal(consumer.Environments)
+	if err != nil {
+		return nil, err
+	}
+
+	var environments []json.RawMessage
+	err = json.Unmarshal(environmentsJSON, &environments)
+	if err != nil {
+		return nil, err
+	}
```
jirihnidek: This does not seem as optimal solution. You marshal some part of structure to string and then you unmarshal this string to `[]json.RawMessage`.

cmd/rhc-server/rhsm.go
```diff
@@ -54,3 +56,66 @@ func IsSystemRegistered() (bool, error) {
...
+	owner, err := rhsmClient.GetOwner()
+	if err != nil {
+		return nil, err
+	}
+
+	return owner, nil
+}
+
+// GetConsumerEnvironments retrieves the environments from the registered consumer.
+func GetConsumerEnvironments() (*[]json.RawMessage, error) {
```
jirihnidek: I believe that this function should return list of structures representing environments. It should be something like this:

```go
func GetConsumerEnvironments() ([]rhsm2.Environment, error) {
    // rest of method
}
```

To be able to implement it in elegant way I created this PR: https://github.com/jirihnidek/rhsm2/pull/94

Then it will be able to simplify this function to something like this:

```go
// GetConsumerEnvironments retrieves the environments from the registered consumer.
func GetConsumerEnvironments() ([]rhsm2.Environment, error) {
	appName := AppName
	rhsmClient, err := rhsm2.GetRHSMClient(&appName, nil)
	if err != nil {
		return nil, &ClientError{Message: err.Error()}
	}

	consumer, err := rhsmClient.GetConsumer(nil)
	if err != nil {
		return nil, err
	}

	return consumer.Environments, nil
}
```

cmd/rhc-server/rhc-server.go
```diff
@@ -115,3 +116,46 @@ func (b *RHSMBackend) IsRegistered(in *rhsmapi.IsRegisteredIn) (*rhsmapi.IsRegis
...
+	org, err := GetConsumerOrg()
+	if err != nil {
+		slog.Debug("Failed to get consumer organization", "error", err)
+		return nil, &consumerapi.SystemNotRegisteredError{}
+	}
+	orgData := map[string]string{"key": *org}
+	orgJSON, err := json.Marshal(orgData)
+	if err != nil {
+		return nil, fmt.Errorf("failed to marshal organization data: %w", err)
+	}
```
jirihnidek: This will have to be simplified when related PR for rhsm2 will be merged.

integration-tests/test_consumer_api.py
jirihnidek, line 1: I hope that we will be able to use `behave` soon. If it was used, then all you would need to do is to modify this file: https://github.com/jirihnidek/rhsm-server/blob/main/features/com.redhat.rhsm.feature

jirihnidek (changes requested):
Thanks for the PR. I have several comments and I created several PRs for rhsm2 to be able to simplify this PR:
* https://github.com/jirihnidek/rhsm2/pull/93
* https://github.com/jirihnidek/rhsm2/pull/94

cmd/rhc-server/rhsm.go
```diff
@@ -54,3 +56,66 @@ func IsSystemRegistered() (bool, error) {
...
+	uuid, err := rhsmClient.GetConsumerUUID()
+	if err != nil {
+		return nil, err
+	}
+
+	return uuid, nil
+}
+
+// GetConsumerOrg retrieves the organization key from the installed consumer certificate.
+func GetConsumerOrg() (*string, error) {
```
m-horky: I'd like to call this `Organization` (in full) instead of `Org` for consistency with the rest of the codebase.

integration-tests/test_consumer_api.py
```diff
@@ -0,0 +1,242 @@
...
+        )
+
+    yield
+
+    if not was_active:
+        subprocess.run(
+            ["systemctl", "disable", "--now", socket_name],
+            check=False,
+            capture_output=True,
+        )
```
m-horky: This now duplicates the code in `test_collector_api`. Any reason we aren't moving them to a separate location? It may be done after this PR if we needed to, but if before, it should be a separate commit & PR.

integration-tests/test_consumer_api.py
```diff
@@ -0,0 +1,242 @@
...
+            3. UUID is a non-empty string in standard UUID format
+        """
+        response = run_varlinkctl("com.redhat.rhsm.testing.consumer.GetUUID")
+
+        assert "uuid" in response
+        assert isinstance(response["uuid"], str)
+        assert len(response["uuid"]) > 0
+        # UUID format: 8-4-4-4-12 hex characters
+        parts = response["uuid"].split("-")
+        assert len(parts) == 5, f"UUID format invalid: {response['uuid']}"
```
m-horky: Could we use `uuid.UUID()` so we don't implement our own parsing?

varlink/consumerapi/com.redhat.rhc.testing.rhsm.consumer.varlink
```diff
@@ -0,0 +1,21 @@
...
+error SystemNotRegistered()
+
+# Return UUID from installed consumer certificate
+method GetUUID() -> (
+    uuid: string
+)
+
+# Return organization that the registered system belongs to
+method GetOrganization() -> (
+    org: object
```
m-horky: I'd like to be explicit about the returned value, instead of it being an arbitrary map. At least for the production version of this API, I'm fine with returning `object` as long it doesn't happen outside of the `.testing.` Varlink namespace.

varlink/consumerapi/com.redhat.rhsm.testing.consumer.varlink
```diff
@@ -0,0 +1,21 @@
+# Consumer interface for the RHSM API
+# Any method in this interface can be used only when the system is registered
+interface com.redhat.rhsm.testing.consumer
```
m-horky: Why not `com.redhat.rhc.testing.rhsm.consumer` or something like that? The socket is named `com.redhat.rhc`, I'm not sure whether it should be serving other namespaces.

cmd/rhc-server/rhsm.go
m-horky: I'd like to follow Jiri's guidance on always returning `[]Environment` object (not nullable). Returning a pointer to a slice isn't usually done in Go.

integration-tests/test_consumer_api.py
```diff
@@ -0,0 +1,242 @@
...
+        """
+        response = run_varlinkctl(
+            "com.redhat.rhsm.testing.consumer.GetEnvironments"
+        )
+
+        assert "environments" in response or response.get("environments") is None
+        envs = response.get("environments")
+        assert envs is None or isinstance(envs, list)
+
+    def test_get_uuid_consistency(self):
```
m-horky: What value does this test provide? The UUID is read from a disk, is there a situation where it might not, and this test would catch it?

cmd/rhc-server/rhsm.go
m-horky, line 1: Is there any reason to use `rhsm2` dependency instead of the already established code in `internal/subman/`? This may cause weird behavior if the library differs from how subscription-manager handles it.

cmd/rhc-server/rhsm.go
```diff
@@ -54,3 +56,66 @@ func IsSystemRegistered() (bool, error) {
 
 	return true, nil
 }
+
+// GetConsumerUUID retrieves the consumer UUID from the installed consumer certificate.
+func GetConsumerUUID() (*string, error) {
+	appName := AppName
+	rhsmClient, err := rhsm2.GetRHSMClient(&appName, nil)
```
m-horky: Is the a reason the library accepts a reference rather than a value?

m-horky (changes requested):
Hi, thanks for the patch, I've left notes inline.

There are no logs performed for happy paths, is that intentional?

cmd/rhc-server/rhsm.go
cnsnyder: That is fair! I was under the impression that moving to the rhsm2 dependency was already decided on. Happy to discuss and / or reimplement as needed/desired!

As I just fell down a rabbit hole on golang's type system and it is late here, I will sleep on this and bring thoughts in a few hours! Cheers!

cmd/rhc-server/rhsm.go
```diff
@@ -54,3 +54,55 @@ func IsSystemRegistered() (bool, error) {
 
 	return true, nil
 }
+
+// GetConsumerUUID retrieves the consumer UUID from the installed consumer certificate.
+func GetConsumerUUID() (*string, error) {
```
m-horky, line 51: Should we return `(string, error)` instead? Without a pointer, the callers wouldn't have to check for nil dereference.

cmd/rhc-server/rhc-server.go
m-horky, line 1: It seems all errors (network, TLS, library issues) result in a `SystemNotRegisteredError`. Can we differentiate between the failure modes? I worry this could confuse users.

cmd/rhc-server/main.go
```diff
@@ -59,6 +60,9 @@ func run() error {
 	collectorapi.Handler{Backend: NewCollectorBackend()}.Register(registry)
 	rhsmapi.Handler{Backend: NewRHSMBackend()}.Register(registry)
 
+	consumerHandler := consumerapi.Handler{Backend: NewConsumerBackend()}
+	consumerHandler.Register(registry)
```
m-horky, line 64: Nit: Is there any reason this needs to be split as two statements, but the two handler calls above are a single line?

varlink/consumerapi/com.redhat.rhc.testing.rhsm.consumer.varlink
```diff
@@ -0,0 +1,23 @@
+# Consumer interface for the RHSM API
+# Any method in this interface can be used only when the system is registered
+interface com.redhat.rhc.testing.rhsm.consumer
```
m-horky, line 8: I know I made a comment earlier, but now we are adding a second `.testing.` interface with a different naming convention:

```
com.redhat.rhsm.testing
com.redhat.rhc.testing.rhsm.consumer
```

I don't want to block this PR on this issue, but I'd like to resolve it at some point.

Nothing actionable here probably, feel free to close this thread.

varlink/consumerapi/com.redhat.rhc.testing.rhsm.consumer.varlink
```diff
@@ -0,0 +1,23 @@
...
+
+# Error returned when a method is called on an unregistered system
+error SystemNotRegistered()
+
+# Return UUID from installed consumer certificate
+method GetUUID() -> (
+    uuid: string
+)
```
m-horky, line 57: Should these methods implement [a `Metadata` field](https://github.com/RedHatInsights/rhc/blob/249a32bb6c353b5df38617eb1cca8c3d0fe90f3a/varlink/rhsmapi/com.redhat.rhsm.testing.varlink#L4-L11)?

cmd/rhc-server/rhc-server.go
```diff
@@ -115,3 +116,67 @@ func (b *RHSMBackend) IsRegistered(in *rhsmapi.IsRegisteredIn) (*rhsmapi.IsRegis
...
+func (b *ConsumerBackend) GetEnvironments(in *consumerapi.GetEnvironmentsIn) (*consumerapi.GetEnvironmentsOut, error) {
+	environments, err := GetConsumerEnvironments()
+	if err != nil {
+		slog.Debug("Failed to get consumer environments", "error", err)
+		return nil, &consumerapi.SystemNotRegisteredError{}
+	}
+	environmentsJSON, err := goObjToVarlinkObj(environments)
+	if err != nil {
+		return nil, err
+	}
```
m-horky, line 165: This makes the server return untyped error instead of a proper registered object. Should we be logging the error here as well?

cmd/rhc-server/rhsm.go
```diff
@@ -54,3 +54,55 @@ func IsSystemRegistered() (bool, error) {
...
+	appName := AppName
+	rhsmClient, err := rhsm2.GetRHSMClient(&appName, nil)
+	if err != nil {
+		return nil, &ClientError{Message: err.Error()}
```
m-horky: Calling `.Error()` discards the original error, hiding the wrapped chain, breaking `errors.As`/`errors.Is`. Can we use `fmt.Errorf("message: %w", err)` instead?

m-horky (changes requested):
Architectural notes: Lots of the new code (GetConsumerUUID, GetConsumerOrganization, GetConsumerEnvironment) from `cmd/rhc-server/rhsm.go`) implements business logic, which should live in `pkg/` or `internal/`. They also duplicate `internal/subman/identity.go` which is already at the correct place.

varlink/consumerapi/com.redhat.rhc.testing.rhsm.consumer.varlink
jirihnidek: We agreed in the design document that testing interface will return generic object and later, when we will be sure what we want to use from these objects, then we will define structures for `Organization`, `Environment`, etc.

varlink/consumerapi/com.redhat.rhsm.testing.consumer.varlink
jirihnidek: We agreed with this name in the design document.

varlink/consumerapi/com.redhat.rhc.testing.rhsm.consumer.varlink
jirihnidek: As I said before, we agreed with this interface name `com.redhat.rhsm.testing` in the design document.

varlink/consumerapi/com.redhat.rhc.testing.rhsm.consumer.varlink
jirihnidek: No, because this method does not communicate with candlepin server at all. It only reads consumer certificate, when it exists.

cmd/rhc-server/rhsm.go
jirihnidek: Yeah, I agree. I would return unmodified error.

cmd/rhc-server/rhsm.go
jirihnidek: Correct me if I am wrong, but `internal/subman/` is intended for D-Bus communication with `rhsm.service`. Only `rhc` CLI should use D-Bus API. This `rhsm.go` is like translation layer between `rhsm2` and Varlink API. I would not mix these two things together.

I think that the code for Varlink API provided by `rhc-server` would need some refactoring sooner or later as I did it here: https://github.com/jirihnidek/rhsm-server/blob/main/cmd/rhsm-server/varlink-com_redhat_rhsm.go

Each interface should have it's own file implementing the interface.

I think that the e.g. `collector.go`, `rhsm.go`, etc.  will not be needed in the future, because the main business logic is in `./internal` or `rhsm2` and each `varlink-com_redhat_*.go` file would implement only calling related methods from `./internal` or `rhsm2`.

Thus, I would keep the code as it is now and it would show how to refactor code later.

cmd/rhc-server/rhsm.go
jirihnidek: It was caused by C coder in me. I prefer to use reference, because it does not duplicate string in the memory, but we can change it in the future. I do not have strong opinion on this in Go.

cmd/rhc-server/rhsm.go
m-horky: > Thus, I would keep the code as it is now and it would show how to refactor code later.

I am worried about introducing parallel functionality between sub-man and rhsm2 that might not match each other well enough. I'd like to replace the D-Bus calls with native implementation over time, but subscription-manager integration isn't supposed to be dropped as a dependency in this moment.

m-horky:
Thanks for the PR. I'm closing it after our internal discussion. There will be a time in the future when we will implement it one way or another, but it is not required at the moment.

PR #500, opened 2026-06-26
chore(internal): Refactor RHSM D-Bus interaction

m-horky:
* Card ID: CCT-2405

This patch replaces package-level functions in internal/subman with a Service interface and its RHSMClient implementation backed by D-Bus.

- Service & RHSMClient are defined in internal/subman/. They represent the API contract rhc has with subscription-manager.
- 'connect' command uses the new internal/subman API. Instead of letting the internal logic handle cases like "we need to ask the user for the name of organization the system should be registered to", the presentation layer does it now, and tries to call client.RegisterwithPassword again with the organization field set.

Assisted-by: Claude Opus <noreply@anthropic.com>


internal/subman/dbus.go
m-horky, line 1: I am not 100% happy with the variable names here, I'd love to brainstorm different options for `dbusConn`, `dialFunc`, `productionDial`.

internal/subman/service.go
```diff
@@ -0,0 +1,56 @@
...
+
+	// GetOrganizations returns the organization keys available for the credentials.
+	GetOrganizations(username, password string) ([]string, error)
+}
+
+// RHSMClient implements [Service] using D-Bus calls to subscription-manager.
+type RHSMClient struct {
+	conn        dbusConn
+	dialPrivate dialFunc
+}
```
m-horky, line 43: I'd like to think of better name for `dialPrivate` as well.

internal/subman/service_test.go
m-horky, line 1: I'd like to brainstorm whether unit testing is worth it here. It's a lot of code and a more complicated production implementation to be able to unit test the methods that generally wrap the D-Bus calls anyway. The maintenance burden might not be worth it.

internal/subman/dbus.go
stephenbourke: I think "dialProduction" describes an environment more than what the function does. Based off the log output of this function, something like "dialPrivateDBusSocket"?

pkoprda: * `dbusConn` - seems fine
* `dialFunc` - what about `dbusConnector`? The `-er` suffix is idiomatic Go for function types
* `productionDial` - maybe `dbusSocketDial`?

internal/subman/service_test.go
pkoprda: I don't think it's worth it, I am afraid we would burn a lot of time maintaining such complicated tests.

m-horky: Dropping.

internal/subman/dbus.go
m-horky: This got solved by itself by dropping unit tests (see the other thread).

cmd/rhc/configure_features_cmd.go
```diff
@@ -64,7 +64,11 @@ func beforeFeaturesStatusAction(ctx context.Context, cmd *cli.Command) (context.
 // featuresStatusAction displays the current status or preferences of all features.
 func featuresStatusAction(ctx context.Context, cmd *cli.Command) error {
 	logCommandStart(cmd)
-	isRegistered, err := subman.IsRegistered()
+	rhsmClient, err := subman.NewRHSMClient()
+	if err != nil {
+		return err
+	}
+	isRegistered, err := rhsmClient.IsRegistered()
```
jirihnidek, line 71: Interesting... There will be two implementations of functions replying question: "Is the system registered?" in this git repository. The first function will call D-Bus method provided by `rhsm.service`, and the second function will implement direct communication with candlepin server for `rhc-server` providing Varlink API. There is a risk that somebody will do some mistake in the future and `rhc` will start to call wrong method and it will call candlepin server directly.

cmd/rhc/configure_features_cmd.go
```diff
@@ -64,7 +64,11 @@ func beforeFeaturesStatusAction(ctx context.Context, cmd *cli.Command) (context.
 // featuresStatusAction displays the current status or preferences of all features.
 func featuresStatusAction(ctx context.Context, cmd *cli.Command) error {
 	logCommandStart(cmd)
-	isRegistered, err := subman.IsRegistered()
+	rhsmClient, err := subman.NewRHSMClient()
+	if err != nil {
+		return err
```
mjcr99: Should we handle the error the same way as is done in the `featuresEnableAction` and `featuresDisableAction` functions?

pkg/feature/content.go
```diff
@@ -24,13 +24,25 @@ func (c Content) RequiredBy() []string {
 }
 
 func (c Content) Enable() error {
-	return subman.SetContentManagement(true)
+	client, err := subman.NewRHSMClient()
```
mjcr99, line 27: Does it make sense to wrap this in a function?
```go
	client, err := subman.NewRHSMClient()
	if err != nil {
		return err
	}
```

mjcr99 (dismissed):
Thanks @m-horky! I have a couple of suggestions.

pkg/feature/content.go
m-horky: We could make the `Content` object own the Client, but considering this code will likely be dropped via CCT-2411 I'm not sure there is any value in it.

cmd/rhc/configure_features_cmd.go
m-horky: Good point, done.

cmd/rhc/configure_features_cmd.go
m-horky: That second function isn't implemented anywhere yet. This PR only restructures existing code, that worry is real, but out of scope for this review.

pkoprda (approved):
LGTM, just please rebase before merging

PR #501, opened 2026-06-29
chore: Drop Release CI jobs

m-horky:
With the removal of Meson, they have no value.


pkoprda (approved):
Make sense, approved 👍

PR #504, opened 2026-06-29
feat(ci): add intelligent test selection shadow mode

Archana-PandeyM:
Implement an initial Intelligent Test Selection (ITS) flow in CI to evaluate selective test execution without affecting existing CI.

- add deterministic ITS selector script and test map
- run non-blocking its-compare in go.yml and publish selection summary/artifact

Running the Script Locally

To test the Intelligent Test Selection (ITS) script locally, run:

python3 scripts/its_select_tests.py \
  --base 167c83f9eb7b7b7909a711572f66c2d98eb3c963^ \
  --head 167c83f9eb7b7b7909a711572f66c2d98eb3c963

Replace the --base and --head values with the Git commits you want to compare.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -35,3 +35,108 @@ jobs:
 
       - name: Go vet
         run: go vet ./...
+
+  its-compare:
+    name: "ITS compare (non-blocking)"
+    runs-on: ubuntu-latest
+    needs: build
+    if: always()
+    continue-on-error: true
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v7
+        with:
+          fetch-depth: 0
+
+      - name: Set up Python
+        uses: actions/setup-python@v6
+        with:
+          python-version: "3.12"
+
+      - name: Set up Go
+        uses: actions/setup-go@v6
+        with:
+          go-version: "stable"
+
+      - name: Install selector dependencies
+        run: python -m pip install --upgrade pip pyyaml
+
+      - name: Resolve base and head refs
+        id: refs
+        run: |
+          if [[ "${{ github.event_name }}" == "pull_request" ]]; then
+            BASE_REF="${{ github.event.pull_request.base.sha }}"
+            HEAD_REF="${{ github.event.pull_request.head.sha }}"
+          else
+            BASE_REF="$(git merge-base HEAD origin/main)"
+            HEAD_REF="$(git rev-parse HEAD)"
+          fi
+          echo "base=${BASE_REF}" >> "${GITHUB_OUTPUT}"
+          echo "head=${HEAD_REF}" >> "${GITHUB_OUTPUT}"
+
+      - name: Run ITS selector
+        id: selector
+        run: |
+          python scripts/its_select_tests.py \
+            --base "${{ steps.refs.outputs.base }}" \
+            --head "${{ steps.refs.outputs.head }}" \
+            --map-file "scripts/its_test_map.yml" \
+            --output-json "selector-output.json" \
+            --github-output "${GITHUB_OUTPUT}"
+
+      - name: Run selected Go tests
+        id: selected-go
+        continue-on-error: true
+        run: |
+          GO_CMD="$(python - <<'PY'
+          import json
+          with open("selector-output.json", "r", encoding="utf-8") as file:
+              data = json.load(file)
+          print(data["recommended_commands"]["go"])
+          PY
+          )"
+          echo "Running: ${GO_CMD}"
+          eval "${GO_CMD}"
+
+      - name: Upload ITS compare artifact
+        uses: actions/upload-artifact@v5
+        if: always()
+        with:
+          name: its-selector-output
+          path: selector-output.json
+
+      - name: Publish ITS comparison summary
+        if: always()
+        env:
+          FULL_GO_RESULT: ${{ needs.build.result }}
+          SELECTED_GO_RESULT: ${{ steps.selected-go.outcome }}
+        run: |
+          python - <<'PY'
+          import json
+          import os
+
+          with open("selector-output.json", "r", encoding="utf-8") as file:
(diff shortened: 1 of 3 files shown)


scripts/its_select_tests.py
pkoprda, line 1: I am guessing this tests selection will be used only when PR is opened. Why do we need to create a new directory `scripts/` in root directory? Can it be created in `.github/` directory?

scripts/its_select_tests.py
```diff
@@ -0,0 +1,211 @@
...
+
+
+def normalize_go_package(path: str) -> str | None:
+    if not path.endswith(".go"):
+        return None
+    parts = Path(path).parts
+    if not parts:
+        return None
+    if parts[0] not in ("cmd", "internal", "pkg"):
+        return None
```
pkoprda: Do we need to hardcode this specific directory structure? Can't be this part of `test_map.yml` file?

scripts/its_test_map.yaml
pkoprda, line 1: Can you change `.yaml`  to `.yml`? I think every YAML file in this repo already uses `.yml`, so we should stay consistent with that.

pkoprda (changes requested):

scripts/its_select_tests.py
Archana-PandeyM: This directory already exists and I tried avoiding any new directories.

scripts/its_test_map.yaml
```diff
@@ -0,0 +1,116 @@
...
+      - "^pkg/operations/"
+    go_packages:
+      - "./cmd/rhc"
+      - "./pkg/feature"
+      - "./pkg/operations"
+    pytest_files:
+      - "integration-tests/test_configure_feature.py"
+      - "integration-tests/test_connect.py"
+
+  - id: collector-surface
```
stephenbourke, line 89: Hi Archana,

This may have been discussed.

I'm wondering about maintenance of directory/file path map for test selection. In a lot of cases, tests to run can align closer with import chains over file paths. For example, a change under varlink should pull in collector/server tests due to cmd/rhc-server and cmd/rhc being dependent on it. So varlink would needs to be included in collector-surface. Keeping pattern lists aligned with every dependency, on large refactors for example, could be tricky.

I don't know, maybe  a dependency graph type of selection process could also work.

scripts/its_select_tests.py
Archana-PandeyM: Good suggestion. The current implementation assumes the repository layout (cmd/, internal/, pkg/). Moving these repository-specific package roots into test_map.yml would make the selector more configurable

scripts/its_test_map.yaml
Archana-PandeyM: That's a good point. The current implementation intentionally uses a static path-to-test mapping because it's deterministic, easy to understand, and straightforward to validate as part of the shadow-mode PoC. However, I agree that keeping the mapping aligned with transitive dependencies could become harder as the codebase evolves or undergoes larger refactoring.

A dependency-aware approach (for example, using the Go import graph to identify impacted packages and then selecting tests) could reduce the maintenance burden and improve selection accuracy. I think that's worth exploring as a future enhancement once we've validated the basic ITS approach. I will mention it in proposal doc.

.github/workflows/go.yml
```diff
@@ -35,3 +35,108 @@ jobs:
...
+
+  its-compare:
+    name: "ITS compare (non-blocking)"
+    runs-on: ubuntu-latest
+    needs: build
+    if: always()
+    continue-on-error: true
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v6
```
pkoprda: This should be `v7`, see #494

scripts/its_select_tests.py
```diff
@@ -0,0 +1,230 @@
...
+
+
+def git_changed_files(base: str, head: str) -> list[str]:
+    diff_ref = f"{base}...{head}"
+    result = subprocess.run(
+        ["git", "diff", "--name-only", diff_ref],
+        check=True,
+        capture_output=True,
+        text=True,
+    )
```
pkoprda, line 42: Shouldn't we throw some exception, if the diff_ref is incorrect (non-existent commit hashes etc.)?

scripts/its_select_tests.py
```diff
@@ -0,0 +1,230 @@
...
+        "changed_files_count": len(selection["changed_files"]),
+        "docs_only": "true" if selection["docs_only_change"] else "false",
+    }
+
+    with open(path, "a", encoding="utf-8") as file:
+        for key, value in outputs.items():
+            file.write(f"{key}={value}\n")
+
+
+def main() -> int:
```
pkoprda: I understand that this script will not be shipped to customers. However, I believe that we should keep the CI scripts also maintainable. This main() function is not really easy to maintain, I believe it can be split into several functions, so it doesn't violate Single Responsibility Principle.

pkoprda (changes requested):

scripts/its_select_tests.py
Archana-PandeyM: check=True already causes subprocess.run() to raise a CalledProcessError if git diff fails (for example, due to invalid or non-existent commit hashes), so the script already fails fast in that scenario.
If your worry is about user friendly message then  I can catch the exception and surface a more user-friendly error message. 

Since this is currently a shadow-phase workflow intended to validate the ITS approach  I prefer to keep the implementation simple and revisit more user-friendly error handling in next iteration may be.

scripts/its_select_tests.py
Archana-PandeyM: Thanks, agreed! Even though this is CI-only code, maintainability still matters. I have refactored the main().

pkoprda (approved):
ACK, failing unit tests on Fedora rawhide were fixed in #519

PR #506, opened 2026-06-29
chore: Explicit permissions of rhc-server.socket

m-horky:
When rhc-server.socket was introduced in 7ead513b, the permissions of com.redhat.rhc socket have not been discussed, and we went with 0660 semi-automatically. However, that opens a space to make assumptions about the API, which is generally not a good idea.

This patch explicity sets both User= and Group= values to 'root', as well as the socket mode to 0600 (dropping group permissions). This change matches the previous behavior, but makes the API explicit.

In the future, we might introduce a system user ('rhc' or otherwise) that will act as a group owner of the .socket, and any additional files rhc owns, allowing other trusted non-root processes running on the system to execute Varlink methods. Once we are there, we will alter the content of this file again to reflect it.

### DIFF
--- a/data/systemd/rhc-server.service
+++ b/data/systemd/rhc-server.service
@@ -6,6 +6,8 @@ After=rhc-server.socket
 
 [Service]
 Type=simple
+User=root
+Group=root
 ExecStart=/usr/libexec/rhc/rhc-server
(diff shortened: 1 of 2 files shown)


jirihnidek (approved):
LGTM

PR #513, opened 2026-06-30
[release-0.2] ci: stop testing on CentOS Stream 8

pkoprda:
Since CentOS Stream 8 has reached end of life, it does not make sense to run the tests on this distribution.


jirihnidek (approved):
LGTM

PR #514, opened 2026-06-30
chore: Update rpm dependencies

m-horky:
- insights-client has never been available in Fedora.
- yggdrasil-worker-package-manager will not be shipped in RHEL 11, we are dropping it from Fedora as well.

### DIFF
--- a/rhc.spec
+++ b/rhc.spec
@@ -44,14 +44,12 @@ BuildRequires:  /usr/bin/dbus-launch
 %endif
 
 Requires: subscription-manager
-%if 0%{?fedora}
-# insights-client does not exist on Fedora
-Recommends: insights-client
-%else
-Requires: insights-client
-%endif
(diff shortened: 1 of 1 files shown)


pkoprda (approved):
ACK

yselkowitz:
@m-horky please note that this does *NOT* drop the dependency from RHEL 11 (or ELN), the conditional needs to be `%if 0%{?rhel} && 0%{?rhel} < 11` to accomplish that.

m-horky:
@yselkowitz Hi, yes I know. The `main` branch only tracks CentOS Stream/RHEL 10 in addition to Fedora. Once we branch upstream in ~a year, the .spec file will drop the mention of those dependencies completely.

yselkowitz:
@m-horky that's not how things work anymore.  ELN represents the ongoing development branch of what will be RHEL 11, so you need to take RHEL 11 into account now.

PR #517, opened 2026-07-02
chore(rhc.spec): Handle Fedora ELN

m-horky:
Fedora ELN defines %{?rhel} macro as the next major RHEL version, currently 11. To successfully build rhc on ELN, the RHEL version needs to be constrained further.


pkoprda (approved):
ACK

PR #518, opened 2026-07-09
feat(operations): add EnableFeature and DisableFeature operations

pkoprda:
* Card ID: CCT-2406

Creates `EnableFeature()` and `DisableFeature()` methods in `pkg/operations/` with explicit dependency cascade using hardcoded switch statements rather than string-based graph traversal. `RemoteManagement` explicitly calls `EnableFeature` for `Content` and `Analytics`. `Results` use `Status` string field ("enabled", "already-enabled", "failed", "dependency-failed") with nested result trees for full cascade visibility.

### DIFF
--- a/pkg/operations/feature_disable.go
+++ b/pkg/operations/feature_disable.go
@@ -1,0 +1,122 @@
+package operations
+
+import (
+	"fmt"
+
+	"github.com/redhatinsights/rhc/internal/datacollection"
+	"github.com/redhatinsights/rhc/internal/remotemanagement"
+	"github.com/redhatinsights/rhc/internal/subman"
+)
+
+// DisableFeatureResult represents the result of disabling a feature.
+// It includes the target feature, status of the operation, nested results
+// for dependents, and any error that occurred.
+type DisableFeatureResult struct {
+	// Feature is the target feature that was requested to be disabled.
+	Feature Feature
+	// Status indicates the outcome: "disabled", "already-disabled", "failed", "dependent-failed"
+	Status string
+	// DependentsDisabled contains the nested results for dependent features that were
+	// disabled during the cascade, in the order they were disabled.
+	DependentsDisabled []DisableFeatureResult
+	// Err contains any error that occurred during the disable operation.
+	// If non-nil, the operation failed and Feature may not be disabled.
+	Err error
+}
+
+// Status constants for DisableFeatureResult
+const (
+	DisableStatusDisabled        = "disabled"
+	DisableStatusAlreadyDisabled = "already-disabled"
+	DisableStatusFailed          = "failed"
+	DisableStatusDependentFailed = "dependent-failed"
+)
+
+// DisableFeature disables a feature and its dependents.
+// It follows the dependent cascade: if other features depend on this feature,
+// those dependents are disabled first, recursively.
+//
+// The dependency graph (explicit, compile-time verified):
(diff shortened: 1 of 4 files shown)


pkoprda:
Failing unit tests on Fedora rawhide are fixed in #519.

pkg/operations/feature_disable.go
```diff
@@ -0,0 +1,136 @@
...
+		return result
+	}
+	if !status.Enabled {
+		result.Status = DisableStatusAlreadyDisabled
+		return result
+	}
+
+	// Disable dependents first (explicit dependency graph reversed)
+	switch opts.Feature {
+	case Analytics:
```
mjcr99: `Analytics` and `Content` cases have the exact same behavior. Could we consider using something like:
```golang
case Analytics, Content:
``` 
To avoid duplicity?

mjcr99 (changes requested):
Hi @pkoprda, thanks!

I'm afraid this PR does not fully align with the ADR we should be following. Based on ADR-11:

```
each feature operation is an explicit function in `pkg/operations/`: `EnableAnalytics`, `DisableAnalytics`, `EnableContent`, `DisableContent`, `EnableRemoteManagement`, `DisableRemoteManagement`` 
```

Also, based on ADR-11:
```golang 
type Feature int
const (
    Analytics Feature = iota
    Content
    RemoteManagement
)
```
Should be located at `pkg/feature/`.

I think there may be additional misalignments, please review the mentioned ADR.behavior

pkoprda:
Hi @mjcr99, thanks for the review. 
> I'm afraid this PR does not fully align with the ADR we should be following. Based on ADR-11:
> 
> ```
> each feature operation is an explicit function in `pkg/operations/`: `EnableAnalytics`, `DisableAnalytics`, `EnableContent`, `DisableContent`, `EnableRemoteManagement`, `DisableRemoteManagement`` 
> ```

This is up to us how we want to define this. Because the Jira card CCT-2406 is saying that it's up to developer if it should be the wrapper around other methods.

> 
> Also, based on ADR-11:
> 
> ```go
> type Feature int
> const (
>     Analytics Feature = iota
>     Content
>     RemoteManagement
> )
> ```
> 
> Should be located at `pkg/feature/`.

I think this is out of scope for CCT-2406.
<br>
I would like to ask @m-horky for review to clarify these concerns.

m-horky:
@pkoprda @mjcr99 It is fair the two artifacts don't line up, my bad. I will leave this up to your preference; unified `EnableFeature` that has the feature passed in can be as elegant as the separate methods.

> `type Feature int` [...] should be located at pkg/feature/.

It might be fine to keep it [as-is in `pkg/operations/`](https://github.com/RedHatInsights/rhc/blob/eef5e6422b6915858ebd7c2de278818d327f8cb5/pkg/operations/feature.go#L6) for now. When there's a need to move it somewhere else later, we can do that. If nothing is pressuring us to move the definition now, let's not concern ourselves by that.

stephenbourke (dismissed):
LGTM

pkg/operations/feature_disable.go
pkoprda: Resolved, please check

stephenbourke (approved):

mjcr99 (approved):
LGTM, thanks @pkoprda!

PR #519, opened 2026-07-09
test(collector): fix failing tests on Fedora rawhide

pkoprda:
JSON unmarshalling error message is now more flexible to handle Go versions differences. Older Go versions reference the parent struct type (timerDto) while newer versions reference the nested struct type. Tests now check for the common substring. Also changed TestEnsureOutputDir() to ignore errors when removing the default output directory during cleanup, as it may already be removed.

Assisted-By: Claude Sonnet <noreply@anthropic.com>

### DIFF
--- a/internal/collector/collector_test.go
+++ b/internal/collector/collector_test.go
@@ -492,9 +492,8 @@ func TestCreateArchive(t *testing.T) {
 func TestEnsureOutputDir(t *testing.T) {
 	t.Run("empty string defaults to defaultOutputDir", func(t *testing.T) {
 		defer func() {
-			if os.RemoveAll(defaultOutputDir) != nil {
-				t.Fatalf("ensureOutputDir() failed to remove default output directory")
-			}
+			// Clean up the default output directory, ignore errors if it's already removed or in use
+			_ = os.RemoveAll(defaultOutputDir)
 		}()
 
 		got, err := ensureOutputDir("")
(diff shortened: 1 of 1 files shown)


stephenbourke:
> JSON unmarshalling error message is now more flexible to handle Go versions differences. Older Go versions reference the parent struct type (timerDto) while newer versions reference the nested struct type. Tests now check for the common substring. Also changed TestEnsureOutputDir() to ignore errors when removing the default output directory during cleanup, as it may already be removed.
> 
> Assisted-By: Claude Sonnet [noreply@anthropic.com](mailto:noreply@anthropic.com)

internal/collector/collector_test.go
```diff
@@ -730,7 +729,7 @@ func TestNewTimer(t *testing.T) {
 					"exit_code": 0
 				}
 			}`,
-			wantError: "json: cannot unmarshal string into Go struct field startedEventDto.last_started.timestamp of type int64",
+			wantError: "last_started.timestamp of type int64",
```
stephenbourke, line 732: This quite surprising. I guess this issue covers the some of the problem.
https://github.com/golang/go/issues/74713

Not anything to hold back this PR as it fixes the tests. But i wonder if we could assert on the err struct (Field, Type, value etc) given that Go apparent does not support stable error messages for unmarshal.

```
err := json.Unmarshal(badJSON, &timer)
	if err != nil {
		var unmarshalErr *json.UnmarshalTypeError
		if errors.As(err, &unmarshalErr) {
            assert.Equal(t, "TimerDto", unmarshalErr.Struct, "Should catch the correct root struct")
```

stephenbourke (approved):
LGTM

internal/collector/collector_test.go
pkoprda: Good point, but I am in favor to keep it simple as possible, so for now we will keep the current structure.

PR #526, opened 2026-07-14
feat(rhc-collector): add bash completion

Peto-RH:
Collector IDs are now suggested as positional arguments for info, enable, and disable subcommands. IDs are read from installed collector configs in /usr/lib/rhc/collectors/.

* Card ID: CCT-1846


cmd/rhc/util.go
```diff
@@ -124,6 +134,9 @@ func validateCollectorCommand(cmd *cli.Command, requiresCollectorID, requiresFor
 		}
 	}
 	if requiresCollectorID && cmd.Args().Len() == 0 {
+		if isShellCompletion() {
+			return nil
+		}
```
Peto-RH, line 139: This was necessary because the CLI library validates the command even when I press `<Tab>`.

For example:
`rhc collect enable <Tab>`
calls:
`rhc collect --generate-shell-completion`
and validates the command. The command is invalid because the collector ID is missing.

data/completion/rhc.bash
```diff
@@ -12,6 +12,7 @@ _cli_bash_autocomplete() {
     else
       opts=$( "${COMP_WORDS[@]:0:$COMP_CWORD}" --generate-shell-completion )
     fi
+    opts=$(echo "${opts}" | cut -d: -f1)
```
Peto-RH, line 15: This was necessary because the completion previously behaved like this:

`rhc collector <Tab>`
```
a                collector        commands         for              info:Display
available        collectors       disable:Disable  help             information
collection       command          enable:Enable    help:Shows       list
list:List        of               one              or               timers
timers:List      timer-based
```

After adding this line:

`opts=$(echo "${opts}" | cut -d: -f1)`

the completion output became:

`disable  enable  help  info  list  timers`

The command processes each input line separately and keeps only the part before the `:` character.

cmd/rhc/collector_cmd.go
```diff
@@ -44,6 +44,24 @@ func newCollectorClient() (*collectorapi.Client, func(), error) {
...
+// as a positional argument.
+func collectorIDShellComplete(_ context.Context, cmd *cli.Command) {
+	PrintFlagNames(cmd.VisibleFlags(), cmd.Root().Writer)
+
+	if cmd.Args().Len() > 0 {
+		return
+	}
+
+	collectors, err := collector.GetCollectors()
+	if err != nil {
```
pkoprda, line 57: Isn't calling here `GetCollectors()` too much heavy-lifting? `GetCollectors()` calls `loadConfigFromFile()` for every collector - during tab completion, this parses every TOML file in `/usr/lib/rhc/collectors/`. Maybe for completion purposes, just listing filenames and striping `.toml` would be faster and sufficient.

pkoprda (changes requested):
Please include Jira card in the commit description and in the PR description.
<br>

When I tried to use `<Tab>` after `list` or `timers` sub-command, it didn't work as expected:
```
# rhc collector list <Tab>
```
produced:
```
# rhc collector list help
```
but I would expect this:
```
# rhc collector list <Tab>
-f    --format    -h    --help
```

But this seems to be broken also for other commands (e.g. `rhc configure features`), so it's out of scope for this PR.

pkoprda (approved):
LGTM

PR #527, opened 2026-07-15
feat: Varlink API for com.redhat.rhsm.testing.content.release

jirihnidek:
* Card ID: CCT-2376
* Implemented Varlink API for release. Implemented the Varlink API as it was designed in design document. The interface com.redhat.rhsm.testing.content.release has 4 new methods for release management:
  * `Download()` ... download release from candlepin server
  * `GetCurrentRelease()` ... returns current release, when it is set
  * `GetAvailableReleases()` ... returns the list of available releases
  * `SetRelease()` ... set the release in the release file and sends the release to candlepin server


cmd/rhc-server/rhsm.go
```diff
@@ -54,3 +52,44 @@ func IsSystemRegistered() (bool, error) {
 
 	return true, nil
 }
+
+// DownloadRelease downloads the current release information from the server. The release can be set
+// on the candlepin server for the given system. The client has to have chance to get this information
+// from the server.
+func DownloadRelease(ipcSender *string, locale *string, correlationID *string) (*string, error) {
+	rhsmClient, err := rhsm2.GetRHSMClient(nil, nil)
```
jlocash, line 102: Could these functions be receivers on some parent type that holds a single connection? It seems odd to reinitialize it in each of these functions.

varlink/releaseapi/com.redhat.rhsm.testing.content.release.varlink
jlocash, line 1: The interface file you've defined here is somewhat different from the interface in the design document

First, there are several error types in the design document that I don't see here:
- `error ContentNotManaged()`
- `error NoReleasesAvailable()`
- `error UnknownRelease()`

Second, the method `UnsetRelease` was in the design document, but not here.

What is the reason for the changes?

jlocash (commented):
Thanks for your PR! I left a couple of comments

jirihnidek:
The default product certificate (`123.pem`) for Fedora 44 that could be used for testing on Fedora 44:

```
-----BEGIN CERTIFICATE-----
MIIE2DCCAsCgAwIBAgIBAzANBgkqhkiG9w0BAQsFADBEMRwwGgYDVQQDDBNDYW5k
bGVwaW4gU2VydmVyIENBMRIwEAYDVQQLDAlDYW5kbGVwaW4xEDAOBgNVBAoMB1Jl
ZCBIYXQwHhcNMjYwNzEwMTM1NTM4WhcNNDYwNzEwMTM1NTM4WjBEMUIwQAYDVQQD
DDlSZWQgSGF0IFByb2R1Y3QgSUQgWzAwMDAwMDAwLTAwMDAtMDAwMC0wMDAwLTAw
MDAwMDAwMDAwMF0wggEiMA0GCSqGSIb3DQEBAQUAA4IBDwAwggEKAoIBAQDEGnRb
vV9IRN757wAdAnWdc1xbFeiTF5uL3RGZnc5WecPF5gPV0yufTSvJ8euJPdKTAC3L
M2ISqwzlvVjWhxpvOQKogxfCMI00C+f1woryVLmionHgFzil3l4wVKho9Z3Ih0M6
n5MySRPrqMOl87loMspWcbn7M/hETvD+4pmsF8a6XeRzV4edxLhz1FNGZimUkU7t
ojH7/GeoWlw5gftlc+5e3XXxAZz/Yw7SO1TRd6sl9gyw01O2wa0YQAYwXgQZdzyZ
/lim7HK99XaABDCuCIeTT7cO+f7GHtycndl3EFUUADVLBRg2kz4Hg77jBE5kSq7p
7e6Myu9pzHfQae2jAgMBAAGjgdQwgdEwCQYDVR0TBAIwADAoBgsrBgEEAZIICQF7
AQQZDBdSZWQgSGF0IEV4YW1wbGUgUHJvZHVjdDAUBgsrBgEEAZIICQF7AgQFDAMx
LjAwFwYLKwYBBAGSCAkBewMECAwGeDg2XzY0MCsGCysGAQQBkggJAXsEBBwMGmZl
ZG9yYS00NCxmZWRvcmEtNDQteDY0Xzg2MB0GA1UdDgQWBBTLPim0sSenVtOAfbBy
QKohLkFrYjAfBgNVHSMEGDAWgBRRR+62lDQ6EOcMTY6y+4CAPH68kzANBgkqhkiG
9w0BAQsFAAOCAgEAdEgMGWf7FsLxDNy7aDz7Unw91kQAnT3R5HTyB/MESHga9k73
1wSvm5+EL+iVHXhc6eQ5VosriYs1bd9Z+GqiwTfBAnIQZt45cwgW82Jh3lVv7Egz
20x1UAQlLLYX0K0Mr2VbR11qH2/+MUYhzNwuiGWbn0s+U9QoCc3nTPD8xin3X5r3
JUYhdpf9+H5os9jaYa624jjp1dXNfp/Krxx8H5bJdFD+gXbz6ySN3y8XefbWc1bV
DKzahFZCmT6jUO1kB7JvwQAaDaS1hfduLbM/uWgohIEOWuHKMmyZ+EDuYq3mZ15U
pd6erBrD5m8O3deTfmIi+Yp9vUKavWpoPQWaJxUE1DmBV0FYJOzA7A3QMOlcUKzm
afHw+CKvNs6vx0nJYohTDUtRGB4XxOt2J4AspckJt7JClfNR7UjGosMveW/Y2pmp
0vzqlqI2EF8Czn5AvcMDetSoo+WLZm9I/LdjJSVbs5MFo0a9Ev6P3LdLskF9ITyz
EvHG8jpR8mXmH3YJtttqSixlWcp0JyUYDcpaEc3KRjh4yRJsjviZzwCKsmwZKRnp
XwivNUy5+6IyjC1M1bJ8ys+Gg+uYfxBZwagpldrs+TL0No6XbQgc5ZuQ0bXWXpgK
UaYVppS04SanZnREOSrS7zGZ+GiCJ3afBTU4Z/olE03sp2GQP9ag2jlChTQ=
-----END CERTIFICATE-----
```

It contains necessary tags. You can display this certificate with `rct cc /etc/pki/product-default/123.pem` file.

varlink/releaseapi/com.redhat.rhsm.testing.content.release.varlink
jirihnidek: Yeah, there are differences, because I realized that majority of errors are not needed. I will tell why: 

* `ContentNotManaged` ... it is not needed, because users should be able to set release despite the content is not managed by `subscription-manager`. Current version of `subscription-manager` also allows to set release despite the management of content is disabled.
* `NoReleasesAvailable` ... no need to return special error from `GetAvailableReleases`. It is enough to return empty list.
* `UnknownRelease` ... this is IMHO also not needed, because users should be able to set any release string to `/etc/dnf/vars/releasever` file. The candlepin also accept any string. Thus, we should not raise some special error from Varlink API. On the other side, the `dnf5` plugin should not allow to set any release. It should allow to set only releases that are returned by `GetAvailableReleases`. This is valid business logic, but it does not belong to rhc-server providing Varlink API. Calling `SetRelease()` should not have unexpected side-effect. It should not do HTTP call to CDN.

It is possible to unset release with `SetRelease()`, when `Release` is set to empty string. However, I can add `UnsetRelease()`. It should be easy.

cmd/rhc-server/rhsm.go
jirihnidek: The `rhsm2.GetRHSMClient()` returns singleton. Thus, when you run this function first time, then it creates  the singleton representing client and it also creates all connections. Note: the `rhsm2.GetRHSMClient()` is called, when `rhc-server` is started.

varlink/releaseapi/com.redhat.rhsm.testing.content.release.varlink
jirihnidek: BTW: We could consider implementation of some caching mechanism of available releases in the future, but we should analyze risks of doing that.

jirihnidek: @jlocash Added `UnsetRelease()` method.

jlocash: Shouldn't `SetRelease()` be limited to whatever `GetAvailableReleases()` returns? What would be the use of setting a value not in that list?

jirihnidek: There are several reasons, why I think that we should not limit `SetRelease()` in this way:
1. As I already said `SetRelease()` should not have unexpected side-effects. It should set release in the `/etc/dnf/vars/releasever` file and then it should send the information to the server asynchronously in go routine. If a client application using Varlink API wants to be sure that it given release is also in the list of available releases, then the client app should call `GetAvailableReleases()` first, then compare the list of releases with given release and then the app can print warning or it can not allow to set release. The Varlink API should not enforce such functionality.
2. It is completely fine to send empty string which is not in any listing file.
3. The logic of setting releases is not perfect in general (subscription-manager or rhsm2) The `GetCdnReleases()` returns union of all CDN releases as you can see here: https://github.com/jirihnidek/rhsm2/blob/main/release.go#L496 ... Thus, it is possible to set release that is not supported by all products (e.g. some products are EUS and some are not). In the worst case setting release could lead to 404 errors of yum/dnf and we cannot do anything about it. It is described e.g. here: https://access.redhat.com/solutions/238533

jirihnidek: There is actually one use case, where calling `GetAvailableReleases()` "inside" `SetRelease()` would not make any sense and it would unnecessarily slow down things and it is `rhsm` libdnf5 plugin. It will call `Download()` Varlink method and then it will set the release using `SetRelease()`, when the release would be different from current release in `/etc/dnf/vers/release` (obtained by `GetCurrentRelease()`). Thus, calling `GetAvailableReleases()` inside this process would not make any sense and it would only slow down the libdnf5 plugin and all dnf transactions.

Generic rule is: We have to be very careful with all REST API calls, because it can cause unnecessary slow down of client applications.

cmd/rhc-server/rhc-server.go
```diff
@@ -192,3 +194,113 @@ func (b *ContentOverrideBackend) Upload(in *overrideapi.UploadIn) (*overrideapi.
...
+		return nil, err
+	}
+
+	return &releaseapi.DownloadOut{Release: *release}, nil
+}
+
+// GetAvailableReleases retrieves a list of available releases and sorts them alphabetically.
+func (c ComRedhatRhsmContentReleaseBackend) GetAvailableReleases(in *releaseapi.GetAvailableReleasesIn) (*releaseapi.GetAvailableReleasesOut, error) {
+	registered, err := IsSystemRegistered()
+	if err != nil || !registered {
```
jlocash, line 240: The original `err` here is being swallowed. Perhaps we should wrap it or add a debug log

cmd/rhc-server/rhc-server.go
```diff
@@ -192,3 +194,113 @@ func (b *ContentOverrideBackend) Upload(in *overrideapi.UploadIn) (*overrideapi.
...
+		releases, err = GetAvailableReleases(nil, nil, nil)
+	}
+
+	if err != nil {
+		return nil, err
+	}
+
+	// Convert map of structure (workaround for absence of set in Go) to list of strings
+	// and sort it alphabetically.
+	var releaseList []string
```
jlocash: In cases where there are no releases, I believe this will serialize to `null` rather than `[]` in the JSON. Is that expected?

varlink/releaseapi/com.redhat.rhsm.testing.content.release.varlink
```diff
@@ -0,0 +1,50 @@
...
+
+# Get release from the candlepin server
+# When system is not registered, then return error SystemNotRegistered
+method Download(
+    metadata: ?Metadata
+) -> (
+    release: string
+)
+
+# Return current release, when it is set. Otherwise, return nul
```
jlocash: typo 

```suggestion
# Return current release, when it is set. Otherwise, return null
```

cmd/rhc-server/rhc-server.go
```diff
@@ -192,3 +194,113 @@ func (b *ContentOverrideBackend) Upload(in *overrideapi.UploadIn) (*overrideapi.
...
+
+// ComRedhatRhsmContentReleaseBackend implements the interface for the com.redhat.rhsm.content.release.varlink
+type ComRedhatRhsmContentReleaseBackend struct{}
+
+// NewComRedhatRhsmContentReleaseRHSMBackend creates a new ComRedhatRhsmContentReleaseBackend instance.
+func NewComRedhatRhsmContentReleaseRHSMBackend() *ComRedhatRhsmContentReleaseBackend {
+	return &ComRedhatRhsmContentReleaseBackend{}
+}
+
+// Download implements the interface for the com.redhat.rhsm.content.release.varlink. It
```
jlocash: This comment cuts abruptly after "It"

Also, the interface name in your comment does not include the `.testing.` bit

cmd/rhc-server/rhc-server.go
```diff
@@ -192,3 +194,113 @@ func (b *ContentOverrideBackend) Upload(in *overrideapi.UploadIn) (*overrideapi.
...
+	var releaseList []string
+	for release := range releases {
+		releaseList = append(releaseList, release)
+	}
+	slices.Sort(releaseList)
+
+	return &releaseapi.GetAvailableReleasesOut{Releases: releaseList}, nil
+}
+
+func (c ComRedhatRhsmContentReleaseBackend) GetCurrentRelease(in *releaseapi.GetCurrentReleaseIn) (*releaseapi.GetCurrentReleaseOut, error) {
```
jlocash, line 283: Please add a doc comment

cmd/rhc-server/rhc-server.go
```diff
@@ -192,3 +194,113 @@ func (b *ContentOverrideBackend) Upload(in *overrideapi.UploadIn) (*overrideapi.
...
+
+	// When release is empty string, return nil pointer
+	if release == "" {
+		return &releaseapi.GetCurrentReleaseOut{Release: nil}, nil
+	}
+
+	return &releaseapi.GetCurrentReleaseOut{Release: &release}, nil
+}
+
+func (c ComRedhatRhsmContentReleaseBackend) SetRelease(in *releaseapi.SetReleaseIn) (*releaseapi.SetReleaseOut, error) {
```
jlocash, line 304: Please add a doc comment

cmd/rhc-server/rhc-server.go
```diff
@@ -192,3 +194,113 @@ func (b *ContentOverrideBackend) Upload(in *overrideapi.UploadIn) (*overrideapi.
...
+		err = SetRelease(in.Release, nil, nil, nil)
+	}
+	if err != nil {
+		return nil, &ServerError{Message: err.Error()}
+	}
+
+	return &releaseapi.SetReleaseOut{Success: true}, nil
+}
+
+func (c ComRedhatRhsmContentReleaseBackend) UnsetRelease(in *releaseapi.UnsetReleaseIn) (*releaseapi.UnsetReleaseOut, error) {
```
jlocash, line 320: Please add a doc comment

jlocash (commented):
I've left some additional comments

varlink/releaseapi/com.redhat.rhsm.testing.content.release.varlink
jlocash: I'm fine leaving it open, but please add a comment to the interface definition spelling out this behavior

cmd/rhc-server/rhc-server.go
jirihnidek: Fixed and fixed in other cases, when `IsRegistered()` is called.

cmd/rhc-server/rhc-server.go
jirihnidek: You are right. It is strange that empty slice is serialized to `null`. Fixed.

varlink/releaseapi/com.redhat.rhsm.testing.content.release.varlink
jirihnidek: Fixed

cmd/rhc-server/rhc-server.go
jirihnidek: Fixed

cmd/rhc-server/rhc-server.go
jirihnidek: Fixed

cmd/rhc-server/rhc-server.go
jirihnidek: Fixed

cmd/rhc-server/rhc-server.go
jirihnidek: Fixed

jirihnidek:
BTW: I added a new feature to our integration tests. When some scenario of integration test fail, then log of `rhc-server` is captured since beginning of scenario and it is printed as part of captured scenario output as you can see on the following example:

```
----
CAPTURED STDOUT: scenario
Step 'varlink method is called' failed!
context stderr: Failed to issue com.redhat.rhsm.testing.content.release.GetAvailableReleases() call: Connection reset by peer
rhc-server log lines since scenario start:
{"level":"debug","time":"2026-08-07T06:00:00Z","message":"cdnURL: http://candlepin-repo:8080, host: candlepin-repo:8080, port 8080"}
{"level":"debug","time":"2026-08-07T06:00:00Z","message":"unable to get release tags: no installed product certificate found"}
2026/08/07 06:00:00 varlink: serving connection: handling call: no installed product certificate found
---- CAPTURED_SCENARIO_OUTPUT_END ----
```

It helped me to discover the bug, because no product cert is installed by default on Fedora, but my testing machine had one testing product certificate installed in `/etc/pki/product-default`.

jlocash (approved):
Thanks for the updates! I spent some time testing the new interface in a c10s VM, and it looks good to me 👍

PR #528, opened 2026-07-16
test(ci): Added integration tests for RHSM Varlink API

jirihnidek:
* Added integration tests for Varlink interface com.redhat.rhsm.testing
* The [behave](https://behave.readthedocs.io/en/latest/) testing framework is used
* Tests are written in Gherkin language, but steps are written in Python programming language
* The com.redhat.rhsm.testing contains only two simple methods and integration tests are written for many combinations of use cases
* Added GitHub action for triggering integration tests
* We use the same approach for testing dnf5 plugins in https://github.com/candlepin/rhsm-dnf5-plugins

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -1,0 +1,136 @@
+name: Varlink Integration Tests
+
+on:
+  workflow_dispatch:
+  pull_request:
+
+permissions:
+  contents: read
+
+jobs:
+  build-rpms:
+    uses: ./.github/workflows/rpmbuild.yml
+
+  integration-tests:
+    name: "Integration Tests (${{ matrix.name }})"
+    needs: build-rpms
+    runs-on: ubuntu-latest
+
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - name: "Fedora latest"
+            image: "registry.fedoraproject.org/fedora:latest"
+            artifact: "rpms-fedora-latest"
+          - name: "Fedora Rawhide"
+            image: "registry.fedoraproject.org/fedora:rawhide"
+            artifact: "rpms-fedora-rawhide"
+
+    container:
+      image: ${{ matrix.image }}
+
+    services:
+      postgres:
+        image: ghcr.io/candlepin/candlepin-db:latest
+      candlepin:
+        image: ghcr.io/candlepin/candlepin-app:latest
+        options: --hostname candlepin.local
+        ports:
+          - 8443:8443
+      candlepin-repo:
+        image: ghcr.io/candlepin/candlepin-repo:latest
+        ports:
+          - 8080:8080
+
+    steps:
+      - name: "Checkout repository"
+        uses: actions/checkout@v7
+
+      - name: "Download RPM artifacts"
+        uses: actions/download-artifact@v8
+        with:
+          name: ${{ matrix.artifact }}
+          path: ./rpms
+
+      - name: "Install test dependencies"
+        run: |
+          dnf --setopt install_weak_deps=False install -y \
+            subscription-manager rhc python3-behave openssl curl \
+            python3-jsonschema dbus-daemon
+
+      - name: "Install built RPMs"
+        run: |
+          dnf install -y ./rpms/*.rpm
+
+      - name: "Wait for candlepin to be ready"
+        run: |
+          for i in $(seq 1 60); do
+            if curl -sfk https://candlepin:8443/candlepin/status > /dev/null 2>&1; then
+              echo "Candlepin is ready"
+              curl -sk https://candlepin:8443/candlepin/status
+              exit 0
+            fi
+            echo "Waiting for candlepin... attempt $i/60"
+            sleep 5
+          done
+          echo "ERROR: Candlepin failed to start within timeout"
+          exit 1
+
+      - name: "Configure subscription-manager for candlepin"
+        run: |
+          subscription-manager config \
+            --server.hostname candlepin \
+            --server.port 8443 \
+            --server.prefix /candlepin \
+            --server.insecure 1 \
+            --rhsm.baseurl http://candlepin-repo:8080
+
+      - name: "Trust candlepin TLS certificate"
+        run: |
+          openssl s_client -connect candlepin:8443 -showcerts </dev/null 2>/dev/null \
+            | openssl x509 -outform PEM > /etc/rhsm/ca/candlepin-ca.pem
+          ln -sf /etc/rhsm/ca/candlepin-ca.pem \
+            /etc/pki/ca-trust/source/anchors/candlepin-ca.pem
+          update-ca-trust
+
+      - name: "Import candlepin RPM GPG key"
+        run: |
+          curl -sf http://candlepin-repo:8080/RPM-GPG-KEY-candlepin \
+            > /etc/pki/rpm-gpg/RPM-GPG-KEY-candlepin
+
+      - name: "Start D-Bus system bus"
+        run: |
+          mkdir -p /run/dbus
+          dbus-daemon --system --fork
+          sleep 1
+
+      - name: "Start RHSM D-Bus service"
+        run: |
+          /usr/libexec/rhsm-service &
+          sleep 2
+          busctl --system status com.redhat.RHSM1
+
+      - name: "Verify candlepin connectivity"
+        run: |
+          subscription-manager version
+
+      - name: "Start RHC Varlink service"
+        run: |
+          /usr/libexec/rhc/rhc-server &
+          sleep 2
+          varlinkctl info /run/rhc/com.redhat.rhc
+
+      - name: "Run integration tests"
+        run: |
+          mkdir -p results
+          behave features/ --no-capture --verbose \
+            --junit --junit-directory results/
+
+      - name: "Upload test results"
+        if: always()
+        uses: actions/upload-artifact@v7
+        with:
+          name: test-results-${{ matrix.artifact }}
+          path: results/
+          retention-days: 30
--- a/features/com.redhat.rhsm.feature
+++ b/features/com.redhat.rhsm.feature
@@ -1,0 +1,78 @@
+Feature: The Varlink interface com.redhat.rhsm
+  The basic Varlink interface com.redhat.rhsm provides basic
+  methods that allows to check if the system is registered and
+  if the system can reach candlepin server
+
+
+  Scenario: IsRegistered() method returns false on unregistered system
+    Given system is not registered
+    When varlink method is called
+      | method       | interface               | arguments |
+      | IsRegistered | com.redhat.rhsm.testing | '{}'      |
+    Then varlink method returns
+      """
+      {"registered":false}
+      """
+
+  Scenario: Ping() method gets status of candlepin server on unregistered system
+    Given system is not registered
+    When varlink method is called
+      | method | interface               | arguments |
+      | Ping   | com.redhat.rhsm.testing | '{}'      |
+    Then method call was successful
+    And method returned JSON compliant with 'com.redhat.rhsm.testing.Ping.json' schema
+
+
+  Scenario: IsRegistered() method returns true on registered system
+    Given system is registered against candlepin server
+    When varlink method is called
+      | method       | interface               | arguments |
+      | IsRegistered | com.redhat.rhsm.testing | '{}'      |
+    Then varlink method returns
+      """
+      {"registered":true}
+      """
+
+  Scenario: Ping() method gets status of candlepin server on registered system
+    Given system is registered against candlepin server
+    When varlink method is called
+      | method | interface               | arguments |
+      | Ping   | com.redhat.rhsm.testing | '{}'      |
+    Then method call was successful
+    And method returned JSON compliant with 'com.redhat.rhsm.testing.Ping.json' schema
+
+
+  Scenario: Ping() method gets status of candlepin server on registered system with metadata
+    Given system is registered against candlepin server
+    When varlink method is called
+      | method | interface               | arguments          |
+      | Ping   | com.redhat.rhsm.testing | '{"metadata": {}}' |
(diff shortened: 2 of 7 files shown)


features/steps/common.py
```diff
@@ -0,0 +1,60 @@
...
+    :param can_fail: whether the command can fail without raising an error
+    :param expected_exit_code: expected exit code of the command
+    :param run_args: additional arguments to pass to subprocess.Popen
+    :return: None
+    """
+    if getattr(context, "faketime", None) is not None:
+        cmd = "NO_FAKE_STAT=1 " + context.faketime + cmd
+
+    if getattr(context, "fake_kernel_release", None) is not None:
+        cmd = context.fake_kernel_release + cmd
```
m-horky: Could you document what this code is doing and why it needs to be here?

features/steps/test_varlink_method.py
```diff
@@ -0,0 +1,97 @@
+from behave import given, when, then, step
+
+import json
+import jsonschema
+
+from common import run_in_context
+from constants import VARLINK_SOCKET
+
+@given("system is registered against candlepin server")
```
m-horky, line 11: This statement hardcodes self-hosted Candlepin into the expectations. Instead, I think we should treat Candlepin deployments as something you configure before the test suite runs, so we could use this for production, stage, Satellite and self-hosted without changing the test itself.

features/steps/test_varlink_method.py
```diff
@@ -0,0 +1,97 @@
...
+def step_impl(context):
```
m-horky: Could we type hint the context parameters?

features/steps/test_varlink_method.py
```diff
@@ -0,0 +1,97 @@
...
+    :param context: behave context
+    :return: None
+    """
+    cmd = "rhc status --format json"
+    run_in_context(context, cmd, can_fail=True)
+    result = json.loads(context.cmd_stdout)
+    if not result["rhsm_connected"]:
+        # TODO: Read credentials from config file
+        cmd = "rhc connect --username admin --password admin --organization donaldduck " \
+              "--enable-feature content --disable-feature analytics --disable-feature remote-management"
```
m-horky: Could you use
```py
cmd = (
    "rhc connect ..."
    "--enable-feature content ..."
)
```
instead? This would get reformatted with black.

features/steps/test_varlink_method.py
```diff
@@ -0,0 +1,97 @@
...
+        varlink_method = row["method"]
+        varlink_args = row["arguments"]
+        counter += 1
+    assert counter == 1, f"Expected exactly one row in table for 'varlink method called' ({counter} provided)"
+
+    cmd = f"varlinkctl call --no-pager --json=short {VARLINK_SOCKET} {varlink_interface}.{varlink_method} {varlink_args}"
+    run_in_context(context, cmd, can_fail=False)
+
+
+@then("method returns")
```
m-horky: Could we be more explicit and have it be `varlink method returns`?

features/steps/test_varlink_method.py
```diff
@@ -0,0 +1,97 @@
...
+        cmd = "rhc disconnect"
+        run_in_context(context, cmd, can_fail=False)
+
+@when("varlink method is called")
+def step_impl(context):
+    """
+    Call a varlink method on a specified interface.
+    :param context: behave context
+    :return: None
+    """
```
m-horky, line 67: Could you please include an example table this statement expects? It's not clear, since `context` isn't typed and `context.table` may contain arbitrary values.

features/steps/test_varlink_method.py
```diff
@@ -0,0 +1,97 @@
...
+
+    :param context:
+    :param json_schema_doc:
+    :return:
+    """
+    with open(f"./features/json-schemas/{json_schema_doc}") as f:
+        schema = json.load(f)
+    result = json.loads(context.cmd_stdout)
+    try:
+        jsonschema.validate(instance=result, schema=schema)
```
m-horky, line 107: How to we plan to keep the JSON schema documents up to date with the changes to Varlink interfaces? Will it be generated, or do we expect this to be reactive and change the doc when tests start failing?

features/com.redhat.rhsm.feature
```diff
@@ -0,0 +1,78 @@
+Feature: The Varlink interface com.redhat.rhsm
+  The basic Varlink interface com.redhat.rhsm should provide basic
+  methods that allows to check if the system is registered and
+  if the system can reach candlepin server
+
+
+  Scenario: IsRegistered() method should return false on unregistered system
```
m-horky: This statement isn't compliant with the Gherkin reference: we shouldn't be using `should` in there, we need to be descriptive:

```
Scenario: IsRegistered() returns false when the system is not registered
```

This applies to all other Scenarios and the Feature description.

features/environment.py
```diff
@@ -0,0 +1,64 @@
...
+    :return: None
+    """
+    if step.status == "failed":
+        print(f"Step '{step.name}' failed!")
+        if hasattr(context, "cmd_stdout") and context.cmd_stdout:
+            print(f"context stdout: {context.cmd_stdout}")
+        if hasattr(context, "cmd_stderr") and context.cmd_stderr:
+            print(f"context stderr: {context.cmd_stderr}")
+        # TODO: Print content rhsm-server.service since the
+        #       the scenario was started
```
m-horky: Could you please better explain what needs to be done here? I'm not sure what you mean.

m-horky (dismissed):
I like the idea of having declarative tests. I left some notes inline.

features/steps/common.py
jirihnidek: Good question. The code was copy-pasted from: https://github.com/candlepin/rhsm-dnf5-plugins/blob/main/features/steps/common.py It seems that code was AI generated by @DuckBoss . It seems that it does not have any value and it could be deleted.

features/steps/test_varlink_method.py
jirihnidek: Yeah, the candlepin is hard-coded now. The purpose is to run integration tests only against candlepin deployed in the unofficial container. The RHSM Varlink API for dnf5 plugins cannot be used with RHEL. It could be used only with Fedora against locally deployed candlepin. Using this against production nor stage nor satellite does not have any value now. Thus, I think that we could improve it later, when it will be really needed or when we will have a capacity for that.

jirihnidek: BTW: we have a new version of unofficial candlepin container that uses original candlepin container. I will update this PR to use this new split containers. More details here:
* https://github.com/candlepin/candlepin-container-unofficial/pull/12
* https://github.com/candlepin/candlepin-container-unofficial/pull/14
* Another improvement waiting for review: https://github.com/candlepin/candlepin-container-unofficial/pull/16

features/steps/test_varlink_method.py
jirihnidek: Probably yes, but official documentation does not use type hints: https://behave.readthedocs.io/en/latest/tutorial/ and dnf team also doe not use type hints as you can see here: https://github.com/rpm-software-management/ci-dnf-stack/tree/main/dnf-behave-tests Not sure why.

jirihnidek: It should be `context: behave.runner.Context`: https://behave.readthedocs.io/en/latest/api/#behave.runner.Context and it seems that it works, but it is Python and it is common practice to add any attribute to this context like `cmd_exitcode`

features/steps/test_varlink_method.py
jirihnidek: Note: I decided to test returned JSON document according JSON scheme, because there could be infinity number of variants of returned JSON document.

The JSON scheme was generated from example of JSON document provided by candlepin swagger. This JSON scheme should be stable. We can consider auto-generate it somehow in the future from Open API for testing interface, but I am not sure if it is worth doing that.

The production API should return type defined in the Varlink interface file. We can consider auto-generate the JSON scheme from the interface file in the future.

features/steps/test_varlink_method.py
jirihnidek: Loading credentials from `settings.json` was actually easy to add.

features/com.redhat.rhsm.feature
jirihnidek: Fixed

features/steps/test_varlink_method.py
jirihnidek: The type hint (context: behave.runner.Context) was added, then table is part of the class Context by default. https://behave.readthedocs.io/en/latest/api/#behave.runner.Context.table . I will add here example.

features/steps/test_varlink_method.py
jirihnidek: Fixed

features/steps/test_varlink_method.py
jirihnidek: Fixed

features/steps/test_varlink_method.py
jirihnidek: Fixed

features/steps/common.py
jirihnidek: Fixed

features/steps/test_varlink_method.py
jirihnidek: In theory it could be possible to load other settings from `settings.json` and configure `rhsm.conf` file from `settings.json` file and we can implement that later, but is it really so important?

features/steps/test_varlink_method.py
jirihnidek: Fixed

features/environment.py
```diff
@@ -0,0 +1,64 @@
+"""
+This module contains environment setup and teardown functions for the test suite.
+"""
+
+import os
+import shutil
+import tempfile
```
Archana-PandeyM: Unused imports  os, shuffled, tempfile are imported but never referenced. Any specific reason?

features/environment.py
```diff
@@ -0,0 +1,64 @@
...
+ENTITLEMENT_BACKUP_DIR_PREFIX = "entitlement-backup-"
+RELEASEVER_FILE = "/etc/dnf/vars/releasever"
+RHSM_HOST_CONFIG_DIR = "/etc/rhsm-host"
+ENTITLEMENT_HOST_CERT_DIR = "/etc/pki/entitlement-host"
+
+
+def before_scenario(context, scenario) -> None:
+    """
+    This function is executed before each scenario in the test suite.
+    Make sure that rhsm-server.socket is running
```
Archana-PandeyM: The docstring  says “Make sure that rhsm-server.socket is running”, but there is no implementation, will it be implemented in another PR?

features/environment.py
```diff
@@ -0,0 +1,64 @@
+"""
+This module contains environment setup and teardown functions for the test suite.
+"""
+
+import os
+import shutil
+import tempfile
+
+ENTITLEMENT_CERT_DIR = "/etc/pki/entitlement/"
```
Archana-PandeyM, line 5: defined but not used.

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,136 @@
+name: Integration Tests
```
Archana-PandeyM: Can we rename it to something like Varlink Integration Tests

features/json-schemas/com.redhat.rhsm.testing.Ping.json
```diff
@@ -0,0 +1,74 @@
...
+        "deviceAuthUrl",
+        "deviceAuthClientId",
+        "deviceAuthScope"
+      ]
+    }
+  },
+  "required": [
+    "status"
+  ]
+}
```
Archana-PandeyM: missing trailing newline

features/steps/test_varlink_method.py
```diff
@@ -0,0 +1,118 @@
...
+    :param context: behave context
+    :return: None
+    """
+    json_object = context.text
+    result = json.loads(context.cmd_stdout)
+    assert result == json.loads(json_object), f"Expected {json_object}, got {result}"
+
+@step("method returned JSON compliant with '{json_schema_doc}' schema")
+def step_impl(context: behave.runner.Context, json_schema_doc):
+    """
```
Archana-PandeyM, line 96: this docstring has no summary - only empty :param / :return placeholders.

Archana-PandeyM (changes requested):

features/environment.py
jirihnidek: Fixed

features/environment.py
jirihnidek: Caused by copy-paste from original repository. Removed.

features/environment.py
jirihnidek: It is correct that it is not used now, but it is Python module constant and it is OK that it is not used. It will be useful in the future.

features/json-schemas/com.redhat.rhsm.testing.Ping.json
jirihnidek: Fixed

features/steps/test_varlink_method.py
jirihnidek: Fixed

.github/workflows/integration-tests.yml
jirihnidek: It is doable, but I would like to know why do you want to rename it? There is no other GitHub Action with integration tests and it could be possible to write integration tests for other parts of rhc using `behave`. It could be useful not only for testing of Varlink API.

Archana-PandeyM: I suggest to rename it to match what’s in the PR. ‘Integration Tests’ implies the main suite; that still lives in pytest/TF, and behave for general rhc isn’t an agreed direction yet.

jirihnidek: OK, fixed.

Archana-PandeyM (approved):
LGTM!

PR #533, opened 2026-07-21
refactor(configure-features): use `pkg/operations` in `configure_features_cmd.go` business logic

mjcr99:
* Card ID: CCT-2408

## Description
- Replaces feature.Get/All/MustGet with operations.ParseFeature/AllFeatures
- Replaces subman.NewRHSMClient().IsRegistered() with operations.IsRegistered()
- Replaces manual dependency cascade with operations.EnableFeature/DisableFeature
- Replaces f.IsEnabled() with operations.FeatureStatus()
- Adds AllFeatures(), Description(), Requires(), RequiredBy() to operations.Feature
- Adds operations.IsRegistered() in new registration.go

Removes direct pkg/feature and internal/subman imports from configure_features_cmd.go per ADR-011 layering rules.

Assited-by: Claude Code

### DIFF
--- a/cmd/rhc/configure_features_cmd.go
+++ b/cmd/rhc/configure_features_cmd.go
@@ -7,11 +7,10 @@ import (
 
 	"github.com/urfave/cli/v3"
 
-	"github.com/redhatinsights/rhc/internal/subman"
 	"github.com/redhatinsights/rhc/internal/ui"
 	"github.com/redhatinsights/rhc/pkg/exitcode"
-	"github.com/redhatinsights/rhc/pkg/feature"
 	"github.com/redhatinsights/rhc/pkg/feature/prefcache"
+	"github.com/redhatinsights/rhc/pkg/operations"
 )
 
 type ConfigureFeatureStatus struct {
@@ -64,11 +63,7 @@ func beforeFeaturesStatusAction(ctx context.Context, cmd *cli.Command) (context.
 // featuresStatusAction displays the current status or preferences of all features.
 func featuresStatusAction(ctx context.Context, cmd *cli.Command) error {
 	logCommandStart(cmd)
-	rhsmClient, err := subman.NewRHSMClient()
-	if err != nil {
-		return cli.Exit(fmt.Sprintf("failed to check registration status: %v", err), exitcode.Software)
-	}
-	isRegistered, err := rhsmClient.IsRegistered()
+	isRegistered, err := operations.IsRegistered()
 	if err != nil {
 		return cli.Exit(fmt.Sprintf("failed to check registration status: %v", err), exitcode.Software)
 	}
@@ -117,8 +112,9 @@ func featuresStatusActionNotRegistered(_ context.Context, cmd *cli.Command) erro
 	var status = ConfigureFeaturesStatus{Connected: false}
 	headers := []string{"FEATURE", "PREFERENCE", "DESCRIPTION"}
 	rows := [][]string{}
-	for _, f := range feature.All() {
-		enabled, err := cache.Get(f.ID())
+	for _, f := range operations.AllFeatures() {
+		featureID := f.String()
+		enabled, err := cache.Get(featureID)
 		if err != nil {
 			return cli.Exit(fmt.Sprintf("failed to get feature preference: %v", err), exitcode.Software)
 		}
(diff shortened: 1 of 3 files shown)


cmd/rhc/configure_features_cmd.go
```diff
@@ -260,37 +254,30 @@ func featuresEnableActionNotRegistered(_ context.Context, _ *cli.Command, target
...
-				continue
-			}
-			if err = required.Enable(); err != nil {
-				return cli.Exit(fmt.Sprintf("failed to enable required feature '%s': %v", requiredName, err), exitcode.Software)
-			}
-			fmt.Printf("Feature '%s' enabled (required by '%s').\n", requiredName, targetName)
+		target, _ := operations.ParseFeature(targetName)
+		result := operations.EnableFeature(operations.FeatureOperationOptions{Feature: target})
+
+		if result.Err != nil {
```
stephenbourke: It might be useful to move this exit to after the DependenciesEnabled loop so a user could see partial success in the logs e.g. if one feature was successfully enabled but another was unsuccessfully enabled.

cmd/rhc/configure_features_cmd.go
```diff
@@ -260,37 +254,30 @@ func featuresEnableActionNotRegistered(_ context.Context, _ *cli.Command, target
...
-			if requiredEnabled {
-				slog.Debug("feature is already enabled", "feature", requiredName)
-				continue
-			}
-			if err = required.Enable(); err != nil {
-				return cli.Exit(fmt.Sprintf("failed to enable required feature '%s': %v", requiredName, err), exitcode.Software)
-			}
-			fmt.Printf("Feature '%s' enabled (required by '%s').\n", requiredName, targetName)
+		target, _ := operations.ParseFeature(targetName)
+		result := operations.EnableFeature(operations.FeatureOperationOptions{Feature: target})
```
stephenbourke, line 264: This is not strictly to do with your change. It probably was missed in the original review. But the code path is being enabled, so it might be worth discussing or addressing at some point.

From the `EnableFeature` function being called here.

https://github.com/RedHatInsights/rhc/blob/main/pkg/operations/feature_enable.go#L64
```
	if status.Enabled {
		result.Status = EnableStatusAlreadyEnabled
		return result
	}
```

Do you think we should handle an inconsistent state where the target is shown as enabled but it's dependencies are not? If a user is in this situation, re-running enable returns early ("already enabled"), but never enables (fixes) the dependency chain. Remediating this state is then quite tricky it seems.

From what I can tell an inconsistent state could be reached on disconnect where one of the steps fails (yggdrasil deactivation) but others succeed (unregister). On disconnect everything is allow run and there no fail early mechanism. Or perhaps an inconsistent state could be reached by something being updated outside of rhc, but perhaps we cant really cover these scenarios.

cmd/rhc/configure_features_cmd.go
mjcr99: Addressed, thanks for the suggestion!

cmd/rhc/configure_features_cmd.go
mjcr99: I think it's a good catch @stephenbourke, but fixing it here might be a little disaligned from the main purpose of this PR. I'm gonna address it in a different PR, so it stays easier to review, and we do not mix fixes, but definitely well spotted in my opinion.

stephenbourke: Absolutely. Make sense!

stephenbourke (dismissed):
Change LGTM!

cmd/rhc/configure_features_cmd.go
```diff
@@ -220,9 +213,10 @@ func featuresEnableActionNotRegistered(_ context.Context, _ *cli.Command, target
 	}
 
 	for _, targetName := range targetNames {
-		target := feature.MustGet(targetName)
+		target, _ := operations.ParseFeature(targetName)
```
pkoprda: Shouldn't we at least log some error if `ParseFeature()` returns an error?

cmd/rhc/configure_features_cmd.go
```diff
@@ -348,9 +331,10 @@ func featuresDisableActionNotRegistered(_ context.Context, _ *cli.Command, targe
...
```
pkoprda: Ditto

cmd/rhc/configure_features_cmd.go
stephenbourke: I assume it's ignored because `beforeFeaturesEnableAction` loops though each feature and catches an error from `ParseFeature`. I suppose if `featuresEnableAction` was used outside of that before hook it could be useful to catch the error here.

mjcr99: Addressed.

cmd/rhc/configure_features_cmd.go
mjcr99: Addressed.

pkoprda (approved):
ACK

stephenbourke (approved):

PR #535, opened 2026-07-21
feat(collector): run as configured user/group

Peto-RH:
* Card ID: CCT-2128

Execute the collector binary as the user and group defined in the collector TOML configuration instead of inheriting the rhc-collector process privileges.

### DIFF
--- a/internal/collector/collector.go
+++ b/internal/collector/collector.go
@@ -7,14 +7,23 @@ import (
 	"log/slog"
 	"os"
 	"os/exec"
+	"os/user"
 	"path/filepath"
 	"regexp"
+	"strconv"
 	"strings"
+	"syscall"
 	"time"
 
 	"github.com/BurntSushi/toml"
 	"github.com/redhatinsights/rhc/internal/systemd"
 )
+
+// UserLookupFunc is a function that looks up a user by username.
+type UserLookupFunc func(string) (*user.User, error)
+
+// GroupLookupFunc is a function that looks up a group by name.
+type GroupLookupFunc func(string) (*user.Group, error)
 
 // ConfigDir is the default directory path where collector configuration files are stored.
 const ConfigDir = "/usr/lib/rhc/collectors/"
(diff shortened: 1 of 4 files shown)


internal/collector/collector.go
```diff
@@ -188,6 +197,35 @@ func ValidateID(id string) (string, error) {
 	return filepath.Base(id), nil
 }
 
+// ResolveUserGroupAttr resolves a username and group name into a syscall.SysProcAttr
+// suitable for setting on exec.Cmd to run a process as the specified user and group.
+func ResolveUserGroupAttr(username, groupname string, lookupUser UserLookupFunc, lookupGroup GroupLookupFunc) (*syscall.SysProcAttr, error) {
```
Peto-RH, line 220: The `lookupUser UserLookupFunc` and `lookupGroup GroupLookupFunc` parameters are passed to `ResolveUserGroupAttr` to make the function easier to unit test.

stephenbourke: Should a `TestResolveUserGroupAttr` unit test be added to `collector_test.go` ?

Peto-RH: Yes, however, it is out of scope for this card. Should I add the unit tests later, or should I contact the reporter and ask whether it would be a good idea to include them in this card?

mjcr99: Hi @Peto-RH, thanks for the PR! 
The way you have implemented it is to make it easier to test, so I think it would make sense to add the UT in this PR.

Other than that, the change looks fine to me.

stephenbourke: Hi @Peto-RH ,

Just my own thoughts on this on an often debated topic. I would take it for granted that unit tests are included on a PR where it makes sense. Unit tests primarily are a benefit to the code reviewer/reader, providing evidence of code correctness and a way to interpret the code. I wouldn't see them as meaning much to a ticket reporter. I think that is where higher level poof is required, integration, E2E. 

I think the change you have in this PR is good either way.

Peto-RH:
I have added unit tests and 2 log messages. Feel free to merge this if you don't have any more comments/suggestions. 
@stephenbourke @mjcr99

stephenbourke (approved):
LGTM

PR #538, opened 2026-07-30
Remove pkg/feature/ (IFeature) dead code and migrate remaining consumers

mjcr99:
Card ID: CCT-2411

## Description
This PR removes the dead code from IFeature and ensures functions from `pkg/operations` and `pkg/feature` are used properly instead of the removed code.


cmd/rhc/connect_cmd.go
pkoprda: Should be this comment updated or removed since IFeature no longer exists?

pkoprda (commented):
Looks good, I have just one clarifying question below.

cmd/rhc/connect_cmd.go
mjcr99: Well spotted @pkoprda! I've also addressed the FIXME since it perfectly fits with the intent of this PR. Now it looks even cleaner, thanks!

pkoprda (dismissed):
LGTM

cmd/rhc/connect_cmd.go
```diff
@@ -602,13 +568,16 @@ func connectAction(ctx context.Context, cmd *cli.Command) error {
...
+			operations.FeatureOperationOptions{Feature: operations.Content}).Enabled
+		connectResult.Features.Analytics.Enabled = operations.FeatureStatus(
+			operations.FeatureOperationOptions{Feature: operations.Analytics}).Enabled
+		connectResult.Features.RemoteManagement.Enabled = operations.FeatureStatus(
+			operations.FeatureOperationOptions{Feature: operations.RemoteManagement}).Enabled
 		fmt.Println(connectResult.Error())
 	}
 
-	err = cmd.Root().Metadata[connectCacheKey].(*prefcache.PreferenceCache).Delete()
+	err = cmd.Root().Metadata[connectCacheKey].(*feature.PreferenceCache).Delete()
```
stephenbourke: You already have the have the cache pulled out from the metadata earlier in this function.

```
cache := cmd.Root().Metadata[connectCacheKey].(*feature.PreferenceCache)
```

So you can probably do this instead:
```
err = cache.Delete()
```

pkg/feature/prefcache.go
```diff
@@ -97,28 +99,18 @@ func LoadCache(filePath string) (*PreferenceCache, error) {
...
-func (c *PreferenceCache) Set(featureName string, enabled bool) error {
-	currentValue, exists := c.prefs[featureName]
-	if !exists {
-		return fmt.Errorf("no such feature: %q", featureName)
-	}
-	if currentValue != enabled {
-		c.prefs[featureName] = enabled
+// Set updates the preference value for a feature and marks the cache as dirty.
+func (c *PreferenceCache) Set(f operations.Feature, enabled bool) {
+	key := f.String()
```
stephenbourke, line 109: Might be worth checking the key exists here?

```
current, exists := c.prefs[key]
if !exists {
	return
}
if current != enabled {
	c.prefs[key] = enabled
	c.dirty = true
}
```

stephenbourke (dismissed):

pkg/feature/prefcache.go
mjcr99: I don't really think so, since the argument `f operations.Feature` is always previously filled using `operations.ParseFeature` which already checks the feature is valid.

cmd/rhc/connect_cmd.go
mjcr99: Good point, addressed!

stephenbourke (approved):

PR #540, opened 2026-08-04
test: add integration tests for rhc collector CLI

Archana-PandeyM:
Covers rhc list, info, and timers against the packaged com.redhat.minimal collector.

Resolves : CCT-2712

### DIFF
--- a/integration-tests/test_collector_cli.py
+++ b/integration-tests/test_collector_cli.py
@@ -6,15 +6,20 @@
 :upstream: Yes
 """
 
-import pytest
 import json
-import subprocess
 import os
+import subprocess
+import time
+
+import pytest
 
 from conftest import (
     RHC_COLLECTOR,
     TIMER_CACHE_DIR,
     MINIMAL_COLLECTOR_ID,
+    MINIMAL_COLLECTOR_NAME,
+    MINIMAL_COLLECTOR_CONFIG_PATH,
+    MINIMAL_SERVICE_UNIT,
     MINIMAL_TIMER_UNIT,
 )
 from utils.systemctl import is_unit_enabled
@@ -68,7 +73,7 @@ def test_rhc_collector_writes_timer_cache(collector_config):
 
 
 @pytest.mark.tier2
-def test_collector_enable_disable_via_cli(minimal_collector_timer_disabled):
+def test_collector_enable_disable_via_cli(rhc, minimal_collector_timer_disabled):
     """
     :id: d0e1f2a3-b4c5-6789-0abc-def012345678
     :title: Verify enable/disable via CLI updates systemd timer state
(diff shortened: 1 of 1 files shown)


stephenbourke:
Thanks @Archana-PandeyM ,

Just to note @mjcr99 has a refactor PR on this same area https://github.com/RedHatInsights/rhc/pull/542
Hopefully there is not too much code conflict.

integration-tests/test_collector_cli.py
```diff
@@ -0,0 +1,442 @@
...
+    """
+    result = rhc.run("collector", "list", "--format", "yaml", check=False)
+
+    assert result.returncode != 0
+    output = result.stdout + result.stderr
+    assert "unsupported format" in output
+
+
+@pytest.mark.tier2
+def test_collector_cli_info(rhc, minimal_collector, minimal_timer_cache):
```
stephenbourke: I don't think `minimal_timer_cache` fixture is used here?

integration-tests/test_collector_cli.py
```diff
@@ -0,0 +1,442 @@
...
+
+MINIMAL_COLLECTOR_ID = "com.redhat.minimal"
+MINIMAL_COLLECTOR_NAME = "Minimal Host Inventory Collector"
+MINIMAL_COLLECTOR_CONFIG = "/usr/lib/rhc/collectors/com.redhat.minimal.toml"
+MINIMAL_COLLECTOR_TIMER = "rhc-collector-com.redhat.minimal.timer"
+MINIMAL_COLLECTOR_SERVICE = "rhc-collector-com.redhat.minimal.service"
+
+
+@pytest.fixture(scope="module")
+def rhc_server_socket():
```
stephenbourke: Is this intentional to have a duplicate function of what is in `test_collector_api.py`? It will be moved in https://github.com/RedHatInsights/rhc/pull/542.

integration-tests/test_collector_cli.py
```diff
@@ -0,0 +1,442 @@
...
+
+
+@pytest.mark.tier2
+@pytest.mark.skip(
+    reason=(
+        "Known issue: JSON output does not match human-readable output, "
+        "tracked by RHEL-217910."
+    )
+)
+def test_collector_cli_list_format_json(rhc, minimal_collector):
```
stephenbourke: Maybe it's useful have a check for expected output on invalid formats?
e.g.
```
rhc collector info --format yaml
unsupported format: yaml (supported formats: "json")`
```

integration-tests/test_collector_cli.py
```diff
@@ -0,0 +1,442 @@
...
+    :expectedresults:
+        1. Timer cache is created
+        2. Command succeeds with exit code 0
+        3. Output contains name, feature, config path, service, timer, and a
+           relative last-run time
+    """
+    result = rhc.run("collector", "info", minimal_collector["id"], check=False)
+
+    assert result.returncode == 0
+    assert f"Name:      {minimal_collector['name']}" in result.stdout
```
stephenbourke: What do you think about a looser match rather than exact spacing? It might make the tests a little less likely to break if the format drifts. 

`"Name:" in result.stdout and minimal_collector['name'] in result.stdout`

Or maybe a regex allowing for any number of spaces?

`r"Name:\s+Minimal Host…"`

integration-tests/test_collector_cli.py
Archana-PandeyM: I agree, lets first merge #542 and then I will refactor this PR again.

integration-tests/test_collector_cli.py
Archana-PandeyM: I looked at this and think one invalid-format test is enough here. --format validation for list / info / timers all go through the same checkFormatFlag helper (validateCollectorCommand with requiresFormat=true).

stephenbourke: Agreed, one test is enough here.

integration-tests/test_collector_cli.py
```diff
@@ -154,3 +146,293 @@ def test_collector_enable_missing_timer(collector_config):
...
+    """
+    result = rhc.run("collector", "list", "--format", "yaml", check=False)
+
+    assert result.returncode != 0
+    output = result.stdout + result.stderr
+    assert "unsupported format" in output
+
+
+@pytest.mark.tier2
+def test_collector_cli_info(rhc, minimal_collector_timer_cache):
```
stephenbourke, line 243: I don't think the `minimal_collector_timer_cache` is used here?

Archana-PandeyM:
Tests are failing because auto-services is inaccessible at the moment due to migration/upgrade of hypervisors.

stephenbourke (approved):
LGTM

PR #542, opened 2026-08-04
refactor(integration-tests): unify systemctl helpers and use shipped collector in tests

mjcr99:
* Card ID: CCT-2713
    
Replace is_socket_enabled and is_timer_enabled with a single
is_unit_enabled helper. Migrate collector API tests to use the shipped
com.redhat.minimal collector instead of generating configs, binaries,
and systemd units from scratch. Add integration tests for CLI
enable/disable, missing timer error, and next_run cross-check against
systemctl.
    
It also splits test_collector_api.py into three focused modules:
- test_collector_api.py: varlink api calls are made
- test_collector_cli.py: rhc collector CLI tests

### DIFF
--- a/integration-tests/conftest.py
+++ b/integration-tests/conftest.py
@@ -2,8 +2,237 @@ import pytest
 import subprocess
 import logging
 import os
+import json
+import time
+import textwrap
+
+from utils.systemctl import is_service_active
 
 logger = logging.getLogger(__name__)
+
+RHC_COLLECTOR = "/usr/libexec/rhc/rhc-collector"
+TIMER_CACHE_DIR = "/var/cache/rhc/collectors"
+
+MINIMAL_COLLECTOR_ID = "com.redhat.minimal"
+MINIMAL_COLLECTOR_NAME = "Minimal Host Inventory Collector"
+MINIMAL_COLLECTOR_CONFIG_PATH = "/usr/lib/rhc/collectors/com.redhat.minimal.toml"
+MINIMAL_TIMER_UNIT = f"rhc-collector-{MINIMAL_COLLECTOR_ID}.timer"
+MINIMAL_SERVICE_UNIT = f"rhc-collector-{MINIMAL_COLLECTOR_ID}.service"
+
+
+@pytest.fixture(scope="module")
+def rhc_server_socket():
+    """
+    Fixture to ensure rhc-server.socket is enabled and running before collector tests.
+    This is required for varlinkctl to communicate with the rhc-server.
+    """
+    socket_name = "rhc-server.socket"
+
+    was_active = is_service_active(socket_name)
+
+    if not was_active:
+        subprocess.run(
+            ["systemctl", "enable", "--now", socket_name],
+            check=True,
+            capture_output=True,
+        )
+
+    yield
+
+    if not was_active:
+        subprocess.run(
+            ["systemctl", "disable", "--now", socket_name],
+            check=False,
+            capture_output=True,
+        )
+
+
+@pytest.fixture
+def collector_config():
+    """
+    Fixture to create a test collector configuration and binary
+    that has NO systemd timer/service units.
+    Used by tests that need a collector without systemd units
+    (e.g. testing the 'missing timer' error path).
+    """
+    collector_config_dir = "/usr/lib/rhc/collectors"
+    collector_bin_dir = "/usr/libexec/rhc/collectors"
+    collector_id = "test.integration.collector"
+    collector_config_path = os.path.join(collector_config_dir, f"{collector_id}.toml")
+    collector_bin_path = os.path.join(collector_bin_dir, collector_id)
+
+    config_content = textwrap.dedent("""
+        [meta]
+        name = "Test Integration Collector"
+        feature = "analytics"
+        type = "ingress"
+
+        [ingress]
+        user = "root"
+        group = "root"
+        content_type = "application/vnd.redhat.test.collection"
+    """).strip()
+
+    collector_script = textwrap.dedent("""
+        #!/bin/bash
+        if [ "$1" = "collect" ]; then
+            echo "test data" > "test-output.txt"
+            exit 0
+        else
+            echo "Usage: $0 collect"
+            exit 1
+        fi
+    """).strip()
+
+    os.makedirs(collector_config_dir, exist_ok=True)
+    with open(collector_config_path, "w") as f:
+        f.write(config_content)
+
+    os.makedirs(collector_bin_dir, exist_ok=True)
+    with open(collector_bin_path, "w") as f:
+        f.write(collector_script)
+    os.chmod(collector_bin_path, 0o755)
+
+    yield {
+        "id": collector_id,
+        "name": "Test Integration Collector",
+        "config_path": collector_config_path,
+        "bin_path": collector_bin_path,
+    }
+
+    if os.path.exists(collector_config_path):
+        os.remove(collector_config_path)
+    if os.path.exists(collector_bin_path):
+        os.remove(collector_bin_path)
+
+
+@pytest.fixture
+def collector_minimal():
+    """
+    Fixture to create a minimal collector with only a config file.
+    No binary, no cache, no systemd units.
+    """
+    collector_dir = "/usr/lib/rhc/collectors"
+    os.makedirs(collector_dir, exist_ok=True)
+
+    collector_id = "test.collector1"
+    collector_name = "Test Minimal Collector"
+
+    config_path = os.path.join(collector_dir, f"{collector_id}.toml")
+    config_content = textwrap.dedent("""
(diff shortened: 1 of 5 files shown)


integration-tests/test_collector_api.py
```diff
@@ -108,48 +113,150 @@ def collector_config():
...
     """
-    Fixture to create timer cache for the test collector.
+    Fixture to create a minimal collector with only a config file.
+    No binary, no cache, no systemd units.
     """
-    timer_dir = "/var/cache/rhc/collectors"
-    collector_id = collector_config["id"]
-    timer_cache_path = os.path.join(timer_dir, f"{collector_id}.json")
+    collector_dir = "/usr/lib/rhc/collectors"
+    os.makedirs(collector_dir, exist_ok=True)
```
pkoprda: I am not sure, if this will work in image mode. Should we care about that or not? Or do you think this should be part of different work, tracked in CCT-2785?

integration-tests/test_collector_api.py
```diff
@@ -374,257 +464,84 @@ def test_collector_info_malformed_id():
...
         assert result.returncode != 0, f"Expected failure for ID: {malformed_id}"
-        # Error output should mention InvalidParameter
         assert (
             "InvalidParameter" in result.stderr
         ), f"Expected InvalidParameter for ID: {malformed_id}, got: {result.stderr}"
 
 
 @pytest.mark.tier2
-def test_collector_list_includes_test_collector(collector_config):
+def test_collector_list_includes_shipped_collector():
```
pkoprda: Wouldn't it make more sense to name this test `test_collector_list_includes_minimal_collector()`?

integration-tests/test_collector_api.py
```diff
@@ -804,3 +702,142 @@ def test_collector_info_multiple_calls_consistency(multiple_test_collectors):
...
+    :title: Verify enable fails with actionable message when timer unit is missing
+    :description:
+        Test that ``rhc collector enable`` returns a non-zero exit code and
+        an actionable error message when the systemd timer unit does not
+        exist on disk.  Uses a test collector that has a config
+        but no systemd units.
+    :tags: Tier 2
+    :steps:
+        1. Create a collector config (no systemd timer/service units)
+        2. Run ``rhc collector enable <id>``
```
pkoprda: Let's make this docstring consistent with other docstrings. What do you think?
```suggestion
        2. Run ``rhc collector enable ID``
```

pkoprda (changes requested):
Looks good, thanks @mjcr99!

I left some comments below. Also please fix the typo in the commit: `Assited-by` -> `Assisted-by`.

integration-tests/test_collector_api.py
```diff
@@ -14,11 +14,17 @@
 import textwrap
```
stephenbourke: I'm wondering if the name of this module (`test_collector_api.py`) is a little misleading as there are more tests than just API tests. Is it worth renaming to something that encapsulates all tests here?

integration-tests/test_collector_api.py
```diff
@@ -374,257 +464,84 @@ def test_collector_info_malformed_id():
...
         assert (
             "InvalidParameter" in result.stderr
         ), f"Expected InvalidParameter for ID: {malformed_id}, got: {result.stderr}"
 
 
 @pytest.mark.tier2
-def test_collector_list_includes_test_collector(collector_config):
+def test_collector_list_includes_shipped_collector():
     """
     :id: e5f6a7b8-c9d0-1234-ef01-23456789abcd
```
stephenbourke: A preexisting issue. Looks like this test shares the same Id as the `test_collector_info_malformed_id` test.

integration-tests/utils/systemctl.py
```diff
@@ -72,3 +74,26 @@ def disable_and_stop_socket(socket_name: str) -> None:
...
+def get_timer_next_trigger(timer_unit: str):
+    """
+    Return the next trigger time for *timer_unit* as a Unix timestamp (int),
+    or None if the timer is not scheduled.
+
+    :param timer_unit: Name of the systemd timer unit
+    :return: Unix timestamp (seconds) or None
+    """
+    result = subprocess.run(
+        ["systemctl", "list-timers", timer_unit, "--output=json"],
```
stephenbourke: Should this use the `--all` flag? I can see that the implementation in `/internal` uses it, 
https://github.com/RedHatInsights/rhc/blob/main/internal/systemd/systemd_timer.go#L67.

integration-tests/test_collector_api.py
```diff
@@ -108,48 +113,150 @@ def collector_config():
...
+        ["systemctl", "disable", "--now", MINIMAL_TIMER_UNIT],
+        check=False,
+        capture_output=True,
+    )
+    subprocess.run(["systemctl", "daemon-reload"], check=True)
+
+    yield
+
+    subprocess.run(
+        ["systemctl", "disable", "--now", MINIMAL_TIMER_UNIT],
```
stephenbourke: Should this yield not be running "systemctl enable" to restore the original state?

stephenbourke (commented):
Commit message has a typo.
"Assited-by" -> "Assisted-by"

integration-tests/test_collector_api.py
mjcr99: I don't know the best way to address it; the problem is that if we only ship a collector, the `test_collector_list_with_multiple_collectors` and `test_collector_info_without_cache_or_timer`tests can not be tested without mocking a new one.

integration-tests/test_collector_api.py
mjcr99: I have split the test file to better organize it.

integration-tests/utils/systemctl.py
mjcr99: Added, you are right!

integration-tests/conftest.py
```diff
@@ -2,9 +2,238 @@
...
 import os
+import json
+import time
+import textwrap
+
+from utils.systemctl import is_service_active
 
 logger = logging.getLogger(__name__)
 
+RHC_COLLECTOR = "/usr/libexec/rhc/rhc-collector"
```
stephenbourke, line 13: I think these constants should be placed in a separate constants module. Usually conftest is not directly imported by other modules, but is a plugin hook for pytest to automatically register test fixtures.

pkoprda: insights-client is using a separate file [constants.py](https://github.com/RedHatInsights/insights-client/blob/main/integration-tests/constants.py), maybe we can have that file with rhc specific constants in this repository

mjcr99: I think that's a good idea to implement, but also expand it to all the integration tests in rhc. I will do it in a different PR.

stephenbourke (dismissed):
Everything looks good to me!

integration-tests/test_collector_api.py
pkoprda: Okay, then it's fine for now.

pkoprda (dismissed):
LGTM, thanks!

Archana-PandeyM (changes requested):
I noticed this PR also includes changes related to automation for Jira https://redhat.atlassian.net/browse/CCT-2750 which appears to be outside the scope of the current Jira. 

Since that work is already planned in a separate PR/task, could we keep this PR focused on the current scope and move those changes to the appropriate PR.

integration-tests/conftest.py
```diff
@@ -2,9 +2,238 @@
...
+        ["systemctl", "disable", "--now", MINIMAL_TIMER_UNIT],
+        check=False,
+        capture_output=True,
+    )
+    subprocess.run(["systemctl", "daemon-reload"], check=True)
+
+    yield
+
+    subprocess.run(
+        ["systemctl", "enable", "--now", MINIMAL_TIMER_UNIT],
```
Archana-PandeyM, line 163: docstring vs actual teardown mismatch
Docstring says restore to disabled; teardown enables:

Archana-PandeyM (changes requested):

integration-tests/test_collector_cli.py
```diff
@@ -0,0 +1,156 @@
...
+    MINIMAL_COLLECTOR_ID,
+    MINIMAL_TIMER_UNIT,
+)
+from utils.systemctl import is_unit_enabled
+
+pytestmark = pytest.mark.usefixtures("rhc_server_socket")
+
+
+@pytest.mark.tier2
+def test_rhc_collector_writes_timer_cache(collector_config, rhc):
```
Archana-PandeyM: fixture rhc is unused

Archana-PandeyM:
Rest all looks good, Please update PR description before merging.

Archana-PandeyM (approved):
LGTM! Thanks Manuel!

PR #543, opened 2026-08-04
test: Added integration tests for content override Varlink API

DuckBoss:
This patch adds Behave integration tests for the
com.redhat.rhsm.testing.content.override interface, covering Download and Upload methods.

- Add feature file with 8 scenarios for content override API
- Add JSON schema for Download response validation
- Add step definitions for DNF5 override file management
- Add reusable generic steps for varlink error and result assertions
- Attempted a fix for the integration tests CI where locally built rhc binary was overriden
- Updated rhsm2 dependency to resolve invalid cached connections
- Updated UploadContentOverrides method to not attempt to send candlepin null overrides

Resolves: CCT-2809

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -56,7 +56,7 @@ jobs:
       - name: "Install test dependencies"
         run: |
           dnf --setopt install_weak_deps=False install -y \
-            subscription-manager rhc python3-behave openssl curl \
+            subscription-manager python3-behave openssl curl \
             python3-jsonschema dbus-daemon
 
       - name: "Install built RPMs"
--- a/cmd/rhc-server/rhsm.go
+++ b/cmd/rhc-server/rhsm.go
@@ -87,6 +87,10 @@ func UploadContentOverrides(ipcSender *string, locale *string, correlationID *st
 		return &ServerError{Message: fmt.Sprintf("failed to read local DNF5 repo overrides: %s", err)}
 	}
 
+	if len(overrides) == 0 {
+		return nil
+	}
+
 	clientInfo := rhsm2.RequestMetadata{IPCSender: ipcSender, Locale: locale, CorrelationId: correlationID}
 	if err := rhsmClient.SendContentOverrides(overrides, &clientInfo); err != nil {
 		return &ServerError{Message: err.Error()}
--- a/features/fixtures.py
+++ b/features/fixtures.py
@@ -1,0 +1,44 @@
+"""
+Behave fixtures for the rhc integration test suite.
+
+Each fixture is a generator function decorated with @fixture.
+Setup runs before yield, teardown runs after the scenario ends.
+Fixtures are activated via @fixture.<name> tags in feature files
+and looked up in fixture_registry by the before_tag hook in
+environment.py.
+"""
+
+import os
+import shutil
+
+from behave import fixture
+
+from steps.constants import OVERRIDE_DIR, OVERRIDE_FILE
+
+
+@fixture
+def dnf5_repo_override_file(context):
+    """
+    Backup and restore the DNF5 repo override file around a scenario.
+    Activated via the @fixture.dnf5_override tag in feature files.
+    :param context: behave context
+    :return: path of the override file
+    """
+    os.makedirs(OVERRIDE_DIR, exist_ok=True)
+
+    backup_file = None
+    if os.path.exists(OVERRIDE_FILE):
+        backup_file = OVERRIDE_FILE + ".behave-backup"
+        shutil.copy2(OVERRIDE_FILE, backup_file)
+
+    yield OVERRIDE_FILE
+
+    if backup_file and os.path.exists(backup_file):
+        shutil.move(backup_file, OVERRIDE_FILE)
+    elif os.path.exists(OVERRIDE_FILE):
+        os.remove(OVERRIDE_FILE)
+
+
+fixture_registry = {
+    "fixture.dnf5_override": dnf5_repo_override_file,
+}
--- a/features/fixtures.py
+++ b/features/fixtures.py
@@ -1,0 +1,50 @@
+"""
+Behave fixtures for the rhc integration test suite.
+
+Each fixture is a generator function decorated with @fixture.
+Setup runs before yield, teardown runs after the scenario ends.
+Fixtures are activated via @fixture.<name> tags in feature files
+and looked up in fixture_registry by the before_tag hook in
+environment.py.
+"""
+
+import os
+import shutil
+
+from behave import fixture
+
+
+@fixture
+def no_redhat_dnf5_override_installed(context):
+    """
+    Ensure that no Red Hat DNF5 repo override file is present during
+    a scenario. If the override file already exists, it is moved to a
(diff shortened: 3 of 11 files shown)


DuckBoss:
Note: some integration test failures are expected right now due to how rhsm2 handles client and connection caching. This needs a fix in rhsm2.

.github/workflows/integration-tests.yml
```diff
@@ -56,7 +56,7 @@ jobs:
       - name: "Install test dependencies"
         run: |
           dnf --setopt install_weak_deps=False install -y \
-            subscription-manager rhc python3-behave openssl curl \
```
jirihnidek, line 59: I had to do the same change in my branch. It should be in separate commit, because it does not add integration tests.

features/steps/test_content_override.py
```diff
@@ -0,0 +1,78 @@
...
+    )
+    labels = {o["content_label"] for o in result["content_overrides"]}
+    assert label in labels, (
+        f"Expected label '{label}' not found in content overrides. "
+        f"Available labels: {labels}"
+    )
+
+
+def _cleanup_override_file(context):
+    """Restore or remove the override file after a scenario."""
```
jirihnidek: Shouldn't we add such cleanup to `environment.py` to methods: `before_scenario()`  `after_scenario()`?

features/steps/test_varlink_method.py
```diff
@@ -117,3 +117,66 @@ def step_impl(context: behave.runner.Context):
     :return: None
     """
     assert context.cmd_exitcode == 0, f"Method call failed with exit code {context.cmd_exitcode}"
+
+
+@when("varlink method is called expecting failure")
+def step_impl(context: behave.runner.Context):
```
jirihnidek, line 123: I have a similar step in my PR #527

features/steps/test_varlink_method.py
```diff
@@ -117,3 +117,66 @@ def step_impl(context: behave.runner.Context):
...
+
+
+@then("varlink method returns error '{error_name}'")
+def step_impl(context: behave.runner.Context, error_name: str):
+    """
+    Verify that varlinkctl reported a specific varlink error.
+    varlinkctl outputs error info on stderr in the format:
+      Error: <interface>.<ErrorType>
+    or as JSON on stdout depending on version.
+    """
```
jirihnidek, line 153: I have a similar thing in my PR: #527

cmd/rhc-server/rhsm.go
```diff
@@ -87,6 +87,10 @@ func UploadContentOverrides(ipcSender *string, locale *string, correlationID *st
 		return &ServerError{Message: fmt.Sprintf("failed to read local DNF5 repo overrides: %s", err)}
 	}
 
+	if len(overrides) == 0 {
+		return nil
+	}
```
jirihnidek, line 92: This change belongs to separate commit, because it is fixing issue found during writing integration tests.

jirihnidek (changes requested):
Overall it looks good. I can see that you introduced two methods that I already implemented in my PR. I have few comments and suggestions.

features/steps/test_varlink_method.py
DuckBoss: I've changed mine to match your implementation here.

features/steps/test_varlink_method.py
DuckBoss: I've changed mine to match your implementation here.

features/steps/test_content_override.py
DuckBoss: I'm not sure if this really needs to be moved there. Those scenario hooks are called before/after every test, and it seems a bit overkill for this situation since this isn't a common file for other tests. 

Considering the above, if you still want me to make this change then I can.

DuckBoss:
Thanks for the review, I've addressed your comments and also added some missing docstrings.

features/steps/test_content_override.py
jirihnidek: You are right. Calling it before each scenario would be overkill. It would be better to use fixtures described here: https://behave.readthedocs.io/en/latest/fixtures/#docid-fixtures

features/fixtures.py
```diff
@@ -0,0 +1,44 @@
...
+    backup_file = None
+    if os.path.exists(OVERRIDE_FILE):
+        backup_file = OVERRIDE_FILE + ".behave-backup"
+        shutil.copy2(OVERRIDE_FILE, backup_file)
+
+    yield OVERRIDE_FILE
+
+    if backup_file and os.path.exists(backup_file):
+        shutil.move(backup_file, OVERRIDE_FILE)
+    elif os.path.exists(OVERRIDE_FILE):
```
jirihnidek: It seems that this fixture also tries to delete `/etc/dnf/repos.override.d/98-redhat.repo` file, when it was created during scenario. It is probably correct behavior, because the fixture tries to restore original state. I would describe this functionality in the doc string of this fixture.

features/fixtures.py
```diff
@@ -0,0 +1,44 @@
...
+        os.remove(OVERRIDE_FILE)
+
+
+fixture_registry = {
+    "fixture.dnf5_override": dnf5_repo_override_file,
```
jirihnidek: The fixture name should be self-explanatory. The `fixure.dnf5_override` doesn't say much. What about: `fixture.no_redhat_dnf5_override_installed`

features/fixtures.py
```diff
@@ -0,0 +1,44 @@
...
+def dnf5_repo_override_file(context):
+    """
+    Backup and restore the DNF5 repo override file around a scenario.
+    Activated via the @fixture.dnf5_override tag in feature files.
+    :param context: behave context
+    :return: path of the override file
+    """
+    os.makedirs(OVERRIDE_DIR, exist_ok=True)
+
+    backup_file = None
```
jirihnidek: Shouldn't we store the backup file in the context? Something like `context.redhat_repo_override_backup`? It seems like a common pattern in behave documentation.

features/fixtures.py
```diff
@@ -0,0 +1,44 @@
...
+    if os.path.exists(OVERRIDE_FILE):
+        backup_file = OVERRIDE_FILE + ".behave-backup"
+        shutil.copy2(OVERRIDE_FILE, backup_file)
```
jirihnidek: If this fixture was renamed `fixture.no_redhat_dnf5_override_installed`, then this would be changed to `shutil.move(OVERRIDE_FILE, backup_file)`.

features/steps/constants.py
```diff
@@ -1 +1,6 @@
+import os
+
 VARLINK_SOCKET = "/run/rhc/com.redhat.rhc"
+
+OVERRIDE_DIR = "/etc/dnf/repos.override.d"
+OVERRIDE_FILE = os.path.join(OVERRIDE_DIR, "98-redhat.repo")
```
jirihnidek: Other directories and file paths are defined in `environment.py`. I am not saying that it is the best place, but I would like to have all these constants on one place.

jirihnidek (changes requested):
Thanks for your updates. I have several suggestions.

jirihnidek (approved):
LGTM

PR #544, opened 2026-08-06
fix: read files directly in GetCanonicalFacts

Peto-RH:
* Card ID: CCT-2522

Replaced os.Stat check-then-read pattern with direct readFile calls, handling file-not-found errors inline. This eliminates race conditions flagged by security scanners.

### DIFF
--- a/internal/canonical_facts/canonical_facts.go
+++ b/internal/canonical_facts/canonical_facts.go
@@ -111,14 +111,11 @@ func CanonicalFactsFromMap(m map[string]interface{}) (*CanonicalFacts, error) {
 // data from the localhost.
 func GetCanonicalFacts() (*CanonicalFacts, error) {
 	var facts CanonicalFacts
-	var err error
 
-	if _, err := os.Stat("/etc/insights-client/machine-id"); !os.IsNotExist(err) {
-		insightsID, err := readFile("/etc/insights-client/machine-id")
-		if err != nil {
-			return nil, err
-		}
-		facts.InsightsID = insightsID
+	var err error
+	facts.InsightsID, err = readFile("/etc/insights-client/machine-id")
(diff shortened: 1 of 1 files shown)


internal/canonical_facts/canonical_facts.go
```diff
@@ -111,15 +111,12 @@ func CanonicalFactsFromMap(m map[string]interface{}) (*CanonicalFacts, error) {
...
-		insightsID, err := readFile("/etc/insights-client/machine-id")
-		if err != nil {
-			return nil, err
-		}
-		facts.InsightsID = insightsID
+	insightsID, err := readFile("/etc/insights-client/machine-id")
+	if err != nil && !os.IsNotExist(err) {
+		return nil, err
 	}
+	facts.InsightsID = insightsID
```
pkoprda: Wouldn't it make more sense to directly assign it to the variable `facts.InsightsID`? We don't have to create intermediate variable for this.

internal/canonical_facts/canonical_facts.go
```diff
@@ -130,13 +127,11 @@ func GetCanonicalFacts() (*CanonicalFacts, error) {
...
-		BIOSUUID, err := readFile("/sys/devices/virtual/dmi/id/product_uuid")
-		if err != nil {
-			return nil, err
-		}
-		facts.BIOSUUID = BIOSUUID
+	biosUUID, err := readFile("/sys/devices/virtual/dmi/id/product_uuid")
+	if err != nil && !os.IsNotExist(err) {
+		return nil, err
 	}
+	facts.BIOSUUID = biosUUID
```
pkoprda: Ditto

pkoprda (approved):
LGTM

PR #545, opened 2026-08-06
fix: Check for selinux store before execution of semanage commands

stephenbourke:
* Card Id: RHEL-233053

- Check for the presence of /var/lib/selinux/targeted before executing semange commands. This will remove warnings on package install on environments that do not have SELinux enabled. The fix covers builds where the package is installed using an image mounted inside a container (to be used on bare metal or VM) and SELinux is enabled after package installation.


pkoprda (approved):
LGTM, thanks

PR #546, opened 2026-08-06
build: support RPM builds without running tests

Peto-RH:
Allow `make rpm NOCHECK=1` to skip unit tests, giving developers a
faster way to build RPM packages when test execution isn’t required.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,6 +4,7 @@
 VERSION := $(shell rpmspec rhc.spec --query --srpm --queryformat '%{version}')
 LDFLAGS := -ldflags "-X github.com/redhatinsights/rhc/pkg/version.Version=$(VERSION)"
 GO_BUILD := go build $(LDFLAGS)
+RPMBUILD_CHECK_OPTION := $(if $(filter 1,$(NOCHECK)),--without check)
 
 # The 'build' target is not used during downstream packaging; it is present for upstream development purposes.
 .PHONY: build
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -4,6 +4,7 @@
 VERSION := $(shell rpmspec rhc.spec --query --srpm --queryformat '%{version}')
 LDFLAGS := -ldflags "-X github.com/redhatinsights/rhc/pkg/version.Version=$(VERSION)"
 GO_BUILD := go build $(LDFLAGS)
+RPMBUILD_CHECK_OPTION := $(if $(NOCHECK),--without check)
```
pkoprda: If I use ` make rpm NOCHECK=0`, it will also creates an RPM without running the tests. Can it be somehow fixed in here?

Peto-RH: I have changed it to accept only `NOCHECK=1`
In the documentation I also have that it accepts (only)
`make rpm NOCHECK=1`

Peto-RH: Of if you would like, I can create `make rpm-nocheck` if you prefer a value-free command

Peto-RH: It would look similar to this
```
.PHONY: rpm-nocheck
    RPMBUILD_CHECK_OPTION := --without check
    rpm
```

pkoprda: This looks fine

pkoprda (approved):
LGTM

PR #547, opened 2026-08-10
fix: ship systemd-tmpfiles configuration for runtime directories

pkoprda:
* Card ID: RHEL-235178

Add data/tmpfiles.d/rhc.conf declaring /run/rhc, /var/tmp/rhc and /var/log/rhc, and update the spec file to install it. This fixes socket activation on fresh boot, a TOCTOU race on /var/tmp/rhc permissions, and /var/log/rhc ownership on bootc systems.

Assisted-By: Claude Opus <noreply@anthropic.com>

### DIFF
--- a/rhc.spec
+++ b/rhc.spec
@@ -95,8 +95,6 @@ install -m 0755 -vp _build/bin/rhc-collector %{buildroot}%{_libexecdir}/%{name}/
 # Bash completion
 install -m 0755 -vd                     %{buildroot}%{bash_completions_dir}/
 install -m 0644 -vp data/completion/rhc.bash  %{buildroot}%{bash_completions_dir}/%{name}
-# Logs
-install -m 0755 -vd                     %{buildroot}%{_localstatedir}/log/%{name}/
 # Collector directories
 install -m 0755 -vd                     %{buildroot}%{_prefix}/lib/%{name}/collectors/
 install -m 0755 -vd                     %{buildroot}%{_libexecdir}/%{name}/collectors/
@@ -119,6 +117,9 @@ install -m 0755 -vd                     %{buildroot}%{_sysconfdir}/%{name}/
 # Minimal collector
 install -m 0755 -vp _build/bin/com.redhat.minimal %{buildroot}%{_libexecdir}/%{name}/collectors/com.redhat.minimal
 install -m 0644 -vp data/collectors/com.redhat.minimal.toml %{buildroot}%{_prefix}/lib/%{name}/collectors/
+# tmpfiles.d
+install -m 0755 -vd                     %{buildroot}%{_tmpfilesdir}
+install -m 0644 -vp data/tmpfiles.d/rhc.conf %{buildroot}%{_tmpfilesdir}/rhc.conf
 
 %if 0%{?with_rhcd_compat}
 # Yggdrasil used to be called rhcd, and was part of rhc. For historical reasons, rhc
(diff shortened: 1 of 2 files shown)


rhc.spec
mjcr99: Should we remove these lanes since we are now marking `/var/log/rhc` as `%ghost`? Now creating it in the buildroot is pointless.

mjcr99 (commented):
Thanks @pkoprda I have a minor comment.

rhc.spec
pkoprda: Make sense, I removed those lines.

rhc.spec
```diff
@@ -222,9 +224,14 @@ fi
 %{_libexecdir}/%{name}/collectors/com.redhat.minimal
 %{_prefix}/lib/%{name}/collectors/com.redhat.minimal.toml
 # Logs
-%dir %{_localstatedir}/log/%{name}/
+%ghost %dir %{_localstatedir}/log/%{name}/
 # Logrotate
 %config(noreplace) %{_sysconfdir}/logrotate.d/rhc
+# Runtime directories
+%ghost %dir /run/%{name}
+%ghost %dir %{_localstatedir}/tmp/%{name}
```
Peto-RH: I suggest removing
`install -m 0755 -vd                     %{buildroot}%{_localstatedir}/log/%{name}/`
and adding `%attr` to the ghost directories.

Keeping the old `%install` line without `%attr` would make `/var/log/rhc` appear correct because RPM reads its `0755 root:root` metadata from the directory in the buildroot. However, because it is `%ghost`, that directory is still not included in the RPM payload—`systemd-tmpfiles` creates it on the installed system. This makes the `%install` line only an indirect way to provide RPM metadata. Removing it and adding `%attr` is clearer. Without both the buildroot directory and `%attr`, RPM records the ghost directory with mode `000`, which can cause incorrect `rpm -V` results and for example make `rpm --restore` apply mode `000`.

After your changes, the RPM header contains:
```
/run/rhc       flags=g  mode=d---------   owner=root:root
/var/log/rhc   flags=g  mode=drwxr-xr-x   owner=root:root
/var/tmp/rhc   flags=g  mode=d---------   owner=root:root
```
the `g` flag means `%ghost`

After adding the proposed attributes, the locally rebuilt RPM records:
```
# Logs
%ghost %attr(0755,root,root) %dir %{_localstatedir}/log/%{name}/
# Logrotate
%config(noreplace) %{_sysconfdir}/logrotate.d/rhc
# Runtime directories
%ghost %attr(0755,root,root) %dir /run/%{name}
%ghost %attr(0700,root,root) %dir %{_localstatedir}/tmp/%{name}
```
```
/run/rhc       mode=drwxr-xr-x
/var/log/rhc   mode=drwxr-xr-x
/var/tmp/rhc   mode=drwx------
```

Peto-RH (commented):
Thank you @pkoprda for the changes. I left one comment.

rhc.spec
pkoprda: @Peto-RH I already removed line `install -m 0755 -vd                     %{buildroot}%{_localstatedir}/log/%{name}/`, but surely I can add `%attr` to the ghost directories. I didn't realize this, thanks.

Also insights-client uses `%attr` macros: https://github.com/RedHatInsights/insights-client/blob/db64b40294bd1bc79d037a30db421935204a4a09/insights-client.spec#L159

Peto-RH (approved):
LGTM

PR #548, opened 2026-08-10
feat(ui): support FORCE_COLOR

Peto-RH:
- CardID: CCT-2653

Honor FORCE_COLOR when stdout is redirected so commands such as `watch -c FORCE_COLOR=1 rhc status` retain ANSI colors.

Handle colors independently from terminal animations, keeping spinners disabled for non-interactive output. Reject FORCE_COLOR when combined with NO_COLOR or --no-color.

### DIFF
--- a/cmd/rhc/main.go
+++ b/cmd/rhc/main.go
@@ -6,6 +6,7 @@ import (
 	"log/slog"
 	"os"
 	"os/signal"
+	"strconv"
 	"strings"
 	"syscall"
 
@@ -23,10 +24,11 @@ import (
 )
 
 const (
-	cliLogLevel  = "log-level"
-	cliCertFile  = "cert-file"
-	cliKeyFile   = "key-file"
-	cliAPIServer = "base-url"
+	cliLogLevel   = "log-level"
+	cliCertFile   = "cert-file"
+	cliKeyFile    = "key-file"
+	envForceColor = "FORCE_COLOR"
+	envNoColor    = "NO_COLOR"
 )
 
 // mainAction is triggered in the case, when no sub-command is specified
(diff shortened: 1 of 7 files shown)


Peto-RH:
There will probably be a need to create a task for the unit/integration tests creation/modification. Because now after this feature is introduced, the tests seem incomplete.

cmd/rhc/main.go
```diff
@@ -23,10 +24,12 @@ import (
...
 
 const (
-	cliLogLevel  = "log-level"
-	cliCertFile  = "cert-file"
-	cliKeyFile   = "key-file"
-	cliAPIServer = "base-url"
+	cliLogLevel   = "log-level"
+	cliCertFile   = "cert-file"
+	cliKeyFile    = "key-file"
+	cliAPIServer  = "base-url"
```
stephenbourke: I don't see `cliAPIServer` used anywhere. It can probably be removed.

cmd/rhc/main.go
```diff
@@ -51,23 +54,43 @@ func mainAction(ctx context.Context, cmd *cli.Command) error {
...
-		// - we're printing in JSON or other parseable format.
-		cmd.IsSet("format"),
+		animationsEnabled,
+		colorsEnabled,
+		machineReadable,
 	)
 }
 
+// isEnvironmentVariableEnabled returns true for Boolean true values.
+func isEnvironmentVariableEnabled(name string) bool {
```
stephenbourke: I see there are some cases in other applications where FORCE_COLOR=2/3 is used. There's not really an official standard for what is used from what I can tell, so I don't think we need to support anything other than 1/true. But when using 2/3 in this code, it is essentially a no-op. We could regard anything greater than or equal to 1 as being switched on or document that only 1 and true will work and anything else being ignored (no-op).

Peto-RH: Okay, I can see that some users might be used to `FORCE_COLOR=2/3`. Let's keep it simple and support only positive integers.

I chose the standard `ParseBool` function because I thought that was the standard approach. However, I can see that we should accept more values as well.

The general `isEnvironmentVariableEnabled` function probably should not exist because it only applies to the `FORCE_COLOR` environment variable. Therefore, I will move the logic directly into the code where it is used.

Peto-RH: Oh the function is used in different 3 places. I will look for different approach

Peto-RH: see this comparison
https://github.com/RedHatInsights/rhc/compare/f7219ab..92e254

stephenbourke: Maybe it's be better to retain this as a general purpose util function and 
have a separate helper function?  Like `isNoColorEnabled`, `isForceColorEnabled` could be used for the logic you have below?

Peto-RH: I agree.
These are the new changes:
https://github.com/RedHatInsights/rhc/compare/f7219ab..9329c6a

stephenbourke (dismissed):
Thanks @Peto-RH . Looks great!

go.sum
```diff
@@ -25,14 +25,6 @@ github.com/henvic/httpretty v0.2.0 h1:U4pKgF9SV4uRrE/7PF85TVnCJxXA91zKFpYU0iFwFO
...
 github.com/jeandeaual/go-locale v0.0.0-20250612000132-0ef82f21eade h1:FmusiCI1wHw+XQbvL9M+1r/C3SPqKrmBaIOYwVfQoDE=
 github.com/jeandeaual/go-locale v0.0.0-20250612000132-0ef82f21eade/go.mod h1:ZDXo8KHryOWSIqnsb/CiDq7hQUYryCgdVnxbj8tDG7o=
-github.com/jirihnidek/rhsm2 v0.0.0-20260729132541-bd687fccd469 h1:3/3RmAOqNXc8V9m/+EGiG0k8m3xpX05WXgZS8GfclJk=
-github.com/jirihnidek/rhsm2 v0.0.0-20260729132541-bd687fccd469/go.mod h1:OhFF2Ju6ZLtmvh1uFiGUIsO9+4vXdXYIp08ZxjpFp3g=
-github.com/jirihnidek/rhsm2 v0.0.0-20260806054428-629538e7441d h1:1jpnEBDJ0ZXQjP8UlBJZZLU/sjD3tCaP7BGXNRdfyQ4=
-github.com/jirihnidek/rhsm2 v0.0.0-20260806054428-629538e7441d/go.mod h1:OhFF2Ju6ZLtmvh1uFiGUIsO9+4vXdXYIp08ZxjpFp3g=
-github.com/jirihnidek/rhsm2 v0.0.0-20260806125901-cf03098a4d1c h1:aUpj2aebLrKSazyOZvo+DP0GKNzfznijypZxDJGO4jg=
-github.com/jirihnidek/rhsm2 v0.0.0-20260806125901-cf03098a4d1c/go.mod h1:OhFF2Ju6ZLtmvh1uFiGUIsO9+4vXdXYIp08ZxjpFp3g=
-github.com/jirihnidek/rhsm2 v0.0.0-20260807102124-ac46e4945e02 h1:zVvIeouzIILYfkFs6vqKnnd6ENTK5nniisv/fbEp/04=
-github.com/jirihnidek/rhsm2 v0.0.0-20260807102124-ac46e4945e02/go.mod h1:OhFF2Ju6ZLtmvh1uFiGUIsO9+4vXdXYIp08ZxjpFp3g=
```
Peto-RH, line 35: I ran 
`go mod tidy` 
`go mod vendor`

stephenbourke: Cool, always a good idea to tidy up unused modules.

stephenbourke (approved):
LGTM

PR #552, opened 2026-08-12
feat: implement WriteContentOverrides method in content override varlink API

DuckBoss:
Add WriteContentOverrides to com.redhat.rhsm.testing.content.override,
enabling clients to write downloaded content override data to the local
DNF5 repo override file at /etc/dnf/repos.override.d/98-redhat.repo.

- Add WriteContentOverrides method to varlink interface definition
- Regenerate overrideapi go code with new method types and handler
- Add WriteLocalContentOverrides helper wrapping rhsm2.WriteDnf5RepoOverrides
- No registration check required (it's local-only file operation)
- Bump rhsm2 dependency to 648f6b60538a

This patch also includes Behave integration test scenarios covering the new
WriteContentOverrides method in com.redhat.rhsm.testing.content.override.

Resolves: [CCT-2814](https://redhat.atlassian.net/browse/CCT-2814)

[CCT-2814]: https://redhat.atlassian.net/browse/CCT-2814?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/cmd/rhc-server/rhc-server.go
+++ b/cmd/rhc-server/rhc-server.go
@@ -172,6 +172,27 @@ func (b *ContentOverrideBackend) Download(in *overrideapi.DownloadIn) (*override
 	return &overrideapi.DownloadOut{ContentOverrides: result}, nil
 }
 
+// WriteContentOverrides writes the given content overrides to the local DNF5 repo override file.
+func (b *ContentOverrideBackend) WriteContentOverrides(in *overrideapi.WriteContentOverridesIn) (*overrideapi.WriteContentOverridesOut, error) {
+	overrides := make([]rhsm2.ContentOverride, 0, len(in.ContentOverrides))
+	for _, co := range in.ContentOverrides {
+		overrides = append(overrides,
+			rhsm2.ContentOverride{
+				ContentLabel: co.ContentLabel,
+				Name:         co.Name,
+				Value:        co.Value,
+			},
+		)
+	}
+
+	if err := WriteLocalContentOverrides(overrides); err != nil {
+		slog.Error("Failed to write content overrides to local DNF5 repo override file", "error", err)
+		return nil, err
+	}
+
+	return &overrideapi.WriteContentOverridesOut{Success: true}, nil
+}
+
 // Upload reads local DNF5 repo overrides and sends them to the candlepin server.
 func (b *ContentOverrideBackend) Upload(in *overrideapi.UploadIn) (*overrideapi.UploadOut, error) {
 	registered, err := IsSystemRegistered()
--- a/cmd/rhc-server/rhsm.go
+++ b/cmd/rhc-server/rhsm.go
@@ -2,8 +2,11 @@ package main
 
 import (
 	"fmt"
+	"os"
+	"path/filepath"
 
 	"github.com/jirihnidek/rhsm2"
+	"github.com/redhatinsights/rhc/varlink/overrideapi"
 )
 
 type ClientError struct {
@@ -95,6 +98,18 @@ func UploadContentOverrides(ipcSender *string, locale *string, correlationID *st
 	return nil
 }
 
+// WriteLocalContentOverrides writes content overrides to the local DNF5 repo override file.
+func WriteLocalContentOverrides(overrides []rhsm2.ContentOverride) error {
+	dir := filepath.Dir(rhsm2.Dnf5RedHatReposOverrideFilePath)
+	if err := os.MkdirAll(dir, 0755); err != nil {
+		return &overrideapi.IOFailedError{Message: fmt.Sprintf("failed to create override directory %s: %s", dir, err)}
+	}
+	if err := rhsm2.WriteDnf5RepoOverrides(overrides, rhsm2.Dnf5RedHatReposOverrideFilePath); err != nil {
+		return &overrideapi.IOFailedError{Message: fmt.Sprintf("failed to write content overrides to local file: %s", err)}
+	}
+	return nil
+}
+
 // DownloadRelease downloads the current release information from the server. The release can be set
 // on the candlepin server for the given system. The client has to have chance to get this information
 // from the server.
@@ -135,4 +150,3 @@ func SetRelease(release string, ipcSender *string, locale *string, correlationID
 	clientInfo := rhsm2.RequestMetadata{IPCSender: ipcSender, Locale: locale, CorrelationId: correlationID}
 	return rhsmClient.SetRelease(release, &clientInfo)
 }
-
--- a/features/com.redhat.rhsm.testing.content.override.feature
+++ b/features/com.redhat.rhsm.testing.content.override.feature
@@ -1,7 +1,8 @@
 Feature: The Varlink interface com.redhat.rhsm.testing.content.override
   The Varlink interface com.redhat.rhsm.testing.content.override provides
-  methods for downloading content overrides from the candlepin server and
-  uploading local DNF5 repo overrides to the server.
+  methods for downloading content overrides from the candlepin server,
+  uploading local DNF5 repo overrides to the server, and writing content
+  overrides to the local DNF5 repo override file.
 
 
   Scenario: Download() method raises error on unregistered system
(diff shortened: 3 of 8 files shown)


cmd/rhc-server/rhc-server.go
```diff
@@ -172,6 +172,32 @@ func (b *ContentOverrideBackend) Download(in *overrideapi.DownloadIn) (*override
...
+			ContentLabel: co.ContentLabel,
+			Name:         co.Name,
+			Value:        co.Value,
+		}
+		if co.Created != nil {
+			o.Created = *co.Created
+		}
+		if co.Updated != nil {
+			o.Updated = *co.Updated
+		}
```
jirihnidek: This is useless to set, because `.repo` file does not need have concept of updated or created repository definition. The `ContentOverrides.Created` and `ContentOverrides.Updated` are there, because it is part of JSON document returned from candlepin. These items were useful, when we cached this information in external JSON file, but we write content overrides directly to drop-in repo file. Thus, this block could be skipped.

cmd/rhc-server/rhsm.go
```diff
@@ -95,6 +97,18 @@ func UploadContentOverrides(ipcSender *string, locale *string, correlationID *st
 	return nil
 }
 
+// WriteLocalContentOverrides writes content overrides to the local DNF5 repo override file.
+func WriteLocalContentOverrides(overrides []rhsm2.ContentOverride) error {
+	dir := filepath.Dir(rhsm2.Dnf5RedHatReposOverrideFilePath)
+	if err := os.MkdirAll(dir, 0755); err != nil {
+		return &ServerError{Message: fmt.Sprintf("failed to create override directory %s: %s", dir, err)}
```
jirihnidek: This is example of wrong usage of `ServerError{}`, because this error is not caused by server. The `ServerError{}` should not be used as generic error. We have two options here:
1. Introduce some generic error for IO errors
2. Introduce another error for this specific use case, when it is not possible to create `/etc/dnf/repo.override.d` directory.

cmd/rhc-server/rhsm.go
```diff
@@ -95,6 +97,18 @@ func UploadContentOverrides(ipcSender *string, locale *string, correlationID *st
...
+	}
+	if err := rhsm2.WriteDnf5RepoOverrides(overrides, rhsm2.Dnf5RedHatReposOverrideFilePath); err != nil {
+		return &ServerError{Message: fmt.Sprintf("failed to write content overrides to local file: %s", err)}
```
jirihnidek: Again, this error has nothing to do with any server. It is caused by the fact that the service providing Varlink API is not able to write a drop-in repo file to local file system. I believe that `rhsm2` should have defined special errors for these cases, and we should define related Varlink error in our .varlink file for content overrides.

features/com.redhat.rhsm.testing.content.override.feature
```diff
@@ -94,6 +95,69 @@ Feature: The Varlink interface com.redhat.rhsm.testing.content.override
...
+  @fixture.no_redhat_dnf5_override_installed
+  Scenario: WriteContentOverrides() writes overrides to local file
+    When varlink method is called
+      | method                | interface                                | arguments                                                                                |
+      | WriteContentOverrides | com.redhat.rhsm.testing.content.override | '{"content_overrides": [{"content_label": "test-write-repo", "name": "enabled", "value": "1"}]}' |
+    Then varlink method returns
+      """
+      {"success":true}
+      """
+    And local DNF5 repo override file contains section 'test-write-repo'
```
jirihnidek, line 107: I would also add here a test step checking if the `enabled` contains value `1`.

features/com.redhat.rhsm.testing.content.override.feature
```diff
@@ -94,6 +95,69 @@ Feature: The Varlink interface com.redhat.rhsm.testing.content.override
...
+
+  @fixture.no_redhat_dnf5_override_installed
+  Scenario: WriteContentOverrides() succeeds with empty array
+    When varlink method is called
+      | method                | interface                                | arguments                    |
+      | WriteContentOverrides | com.redhat.rhsm.testing.content.override | '{"content_overrides": []}' |
+    Then varlink method returns
+      """
+      {"success":true}
+      """
```
jirihnidek, line 124: I think that we also need another case. It would be interesting to check what happens, when there is already something in the `98-redhat.repo` file and we make this Varlink method call. My expectation is that the file should be empty after this call.

features/steps/test_content_override.py
```diff
@@ -47,3 +48,50 @@ def step_impl(context: behave.runner.Context, label: str):
...
+
+
+@when("varlink method is called with saved content overrides from '{name}'")
+def step_impl(context: behave.runner.Context, name: str):
+    """
+    Call WriteContentOverrides using the content_overrides array from a previously saved result.
+    :param context: behave context
+    :param name: key of the previously saved Download result containing content_overrides
+    :return: None
+    """
```
jirihnidek, line 78: Is this a new special method for calling Varlink method necessary? Couldn't we re-use existing methods?

features/com.redhat.rhsm.testing.content.override.feature
```diff
@@ -94,6 +95,69 @@ Feature: The Varlink interface com.redhat.rhsm.testing.content.override
...
+      | WriteContentOverrides | com.redhat.rhsm.testing.content.override | '{"content_overrides": [{"content_label": "test-unreg-repo", "name": "gpgcheck", "value": "0"}]}' |
+    Then varlink method returns
+      """
+      {"success":true}
+      """
+    And local DNF5 repo override file contains section 'test-unreg-repo'
+
+
+  @fixture.no_redhat_dnf5_override_installed
+  Scenario: Download() then WriteContentOverrides() persists overrides locally
```
jirihnidek: This scenario does not test anything special in our varlink methods, but it tries to replicate behavior of old `subscription-manager repos` command and subscription-manager dnf4 plugin. We will not implement a new dnf5 plugin in this way. This scenario requires a new special step that is specific for content overrride. I am not sure if it is really necessary.

features/com.redhat.rhsm.testing.content.override.feature
jirihnidek: We also need to test cases, when the `WriteContentOverrides` is called multiple times with different arguments.

jirihnidek (dismissed):
Thanks for the PR. I have several suggestions and comments.

cmd/rhc-server/rhsm.go
DuckBoss: True, this would be a much better approach.

cmd/rhc-server/rhc-server.go
DuckBoss: Good point, I thought I removed this change already earlier.

cmd/rhc-server/rhc-server.go
```diff
@@ -172,6 +172,27 @@ func (b *ContentOverrideBackend) Download(in *overrideapi.DownloadIn) (*override
...
+	for _, co := range in.ContentOverrides {
+		overrides = append(overrides,
+			rhsm2.ContentOverride{
+				ContentLabel: co.ContentLabel,
+				Name:         co.Name,
+				Value:        co.Value,
+			},
+		)
+	}
+
```
jlocash, line 187: Do we need the `Created` and `Updated` fields on `ContentOverride`? I noticed you added them and then removed them. They should probably be removed from `overrideapi.WriteContentOverridesIn` entirely, and controlled here by the server and returned in `overrideapi.WriteContentOverridesOut`

cmd/rhc-server/rhsm.go
DuckBoss: I've introduced an IOFailed generic error for this case now.

cmd/rhc-server/rhc-server.go
DuckBoss: Yeah those fields don't serve any purpose and it's safe to remove them. They carried over earlier from candlepin's api response, but we can remove these with no impact.

jlocash (approved):
LGTM

PR #553, opened 2026-08-13
test: cover rhc-server socket disabled and already-running paths

Archana-PandeyM:
Extend socket activation integration tests for unavailable socket errors and Varlink success when the service is already active.
Rename is_service_active to is_unit_active

| Acceptance criterion | Covered by |
|---|---|
| Stopped service + enabled socket → Varlink succeeds, service auto-starts | `test_fd3_socket_activation` (existing; fixture `fd3_socket_setup`) |
| Auto-boot → exit 0, valid `collectors` array, service active | Same test (`test_fd3_socket_activation`) |
| Socket disabled → Varlink and CLI fail with actionable error; socket restored in teardown | `test_socket_disabled_varlink_and_cli_fail` (fixture `socket_disabled`) |
| Service already running → Varlink works, service stays active | `test_service_already_running_varlink_succeeds` (fixture `service_running`) |

All cases live in `integration-tests/test_socket_activated_service.py`.

Card Id : CCT-2710

Assisted-by: Cursor


Archana-PandeyM:
Since infra to run the tests is CI is down, here is the Tests results from manual run https://artifacts.osci.redhat.com/testing-farm/be6c5182-8673-415d-9981-be13a71f6671/

integration-tests/test_rhc_server_socket.py
```diff
@@ -9,14 +9,16 @@
 import pytest
```
stephenbourke, line 9: Maybe this module should be renamed as it's testing more than "activated" socket states.

Something like `test_rhc_server_socket.py` ?

integration-tests/test_rhc_server_socket.py
```diff
@@ -34,19 +36,52 @@ def fd3_socket_setup():
...
+    yield
+
+    if not service_was_active:
+        stop_service(RHC_SERVER_SERVICE)
+    if not socket_was_enabled:
+        disable_and_stop_socket(RHC_SERVER_SOCKET)
+
+
 @pytest.mark.tier2
 def test_fd3_socket_activation(fd3_socket_setup):
```
stephenbourke, line 86: I think `fd3_socket_setup` is module scoped, but  only applied on function level within this module. Might be a good idea to remove the scope specification and move it back to default (function level).

stephenbourke (approved):
LGTM

PR #554, opened 2026-08-13
feat(CI): add RPM package verification

mjcr99:
* Card ID: CCT-2708

## Description

Verify that the rhc RPM manifest includes the expected collector directory layout, minimal collector files, and systemd units while excluding runtime-only paths. Also verify installed file permissions and collector config identity on disk.

### DIFF
--- a/.github/actions/verify-rpm-excludes-runtime/action.yml
+++ b/.github/actions/verify-rpm-excludes-runtime/action.yml
@@ -1,0 +1,35 @@
+name: "Verify RPM excludes runtime paths"
+description: "Check that the rhc RPM does not own runtime-generated paths"
+
+inputs:
+  manifest:
+    description: "Path to the pre-generated RPM manifest file (output of rpm -qlp)"
+    required: true
+
+runs:
+  using: composite
+  steps:
+    - name: "Verify runtime paths are excluded"
+      shell: bash
+      run: |
+        FAIL=0
+
+        CACHE_ENTRIES=$(grep '^/var/cache/rhc' "${{ inputs.manifest }}" || true)
+        if [ -n "${CACHE_ENTRIES}" ]; then
+          echo "FAIL  RPM must not own runtime cache paths under /var/cache/rhc/:"
+          echo "${CACHE_ENTRIES}"
+          FAIL=1
+        else
+          echo "  OK  /var/cache/rhc/ entries correctly absent"
+        fi
+
+        RUN_ENTRIES=$(grep '^/run/rhc/.' "${{ inputs.manifest }}" || true)
+        if [ -n "${RUN_ENTRIES}" ]; then
+          echo "FAIL  RPM must not own files under /run/rhc/:"
+          echo "${RUN_ENTRIES}"
+          FAIL=1
+        else
+          echo "  OK  /run/rhc/ entries correctly absent"
+        fi
+
+        exit "${FAIL}"
--- a/.github/actions/verify-rpm-unit-naming/action.yml
+++ b/.github/actions/verify-rpm-unit-naming/action.yml
@@ -1,0 +1,33 @@
+name: "Verify systemd unit naming"
+description: "Check that all rhc-collector-* units follow the rhc-collector-{id}.(service|timer) pattern"
+
+inputs:
+  manifest:
+    description: "Path to the pre-generated RPM manifest file (output of rpm -qlp)"
+    required: true
+
+runs:
+  using: composite
+  steps:
+    - name: "Verify unit naming convention"
+      shell: bash
+      run: |
+        COLLECTOR_UNITS=$(grep '^/usr/lib/systemd/system/rhc-collector-' "${{ inputs.manifest }}" || true)
+        if [ -z "${COLLECTOR_UNITS}" ]; then
+          echo "FAIL  No rhc-collector-* units found in manifest"
+          exit 1
+        fi
+
+        PATTERN='^rhc-collector-[a-z0-9]+\.[a-z0-9]+(\.[a-z0-9]+)*\.(service|timer)$'
+        FAIL=0
+        for UNIT_PATH in ${COLLECTOR_UNITS}; do
+          UNIT_NAME=$(basename "${UNIT_PATH}")
+          if echo "${UNIT_NAME}" | grep -qE "${PATTERN}"; then
+            echo "  OK  ${UNIT_NAME}"
+          else
+            echo "FAIL  ${UNIT_NAME} does not match rhc-collector-{id}.(service|timer)"
+            FAIL=1
+          fi
+        done
+
+        exit "${FAIL}"
--- a/.github/workflows/rpmbuild.yml
+++ b/.github/workflows/rpmbuild.yml
@@ -54,6 +54,53 @@ jobs:
           mkdir -p rpms
           cp ~/rpmbuild/RPMS/*/*.rpm rpms/
 
+      - name: "Identify rhc RPM"
+        id: find-rpm
+        run: |
+          shopt -s nullglob
+          MATCHES=(rpms/rhc-[0-9]*.rpm)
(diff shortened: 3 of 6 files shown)


stephenbourke (commented):
Thanks for the PR @mjcr99 .

Could you put the ticket number in the description?

integration-tests/test_package_verification.py
```diff
@@ -0,0 +1,286 @@
...
+    :steps:
+        1. Read the minimal collector config file
+        2. Verify it contains the expected name field
+    :expectedresults:
+        1. File is readable
+        2. Config contains ``name = "Minimal Host Inventory Collector"``
+    """
+    result = subprocess.run(
+        [
+            "grep", "-q",
```
stephenbourke: You could use tomllib (part of stdlib afaik) here to add a bit more deterministic parsing and not have to fork a grep process. 

For example: 
```
import tomllib

with open(MINIMAL_COLLECTOR_CONFIG_PATH, "rb") as f:
    collector_config = tomllib.load(f)

assert collector_config["meta"]["name"] == MINIMAL_COLLECTOR_NAME
```

integration-tests/test_package_verification.py
```diff
@@ -0,0 +1,286 @@
...
+    for unit_path in collector_units:
+        unit_name = os.path.basename(unit_path)
+        assert pattern.match(unit_name), (
+            f"Unit name {unit_name} does not match "
+            f"rhc-collector-{{id}}.(service|timer) pattern"
+        )
+
+    minimal_service = os.path.join(SYSTEMD_UNIT_DIR, MINIMAL_SERVICE_UNIT)
+    minimal_timer = os.path.join(SYSTEMD_UNIT_DIR, MINIMAL_TIMER_UNIT)
+    assert minimal_service in rpm_manifest, (
```
stephenbourke: Not a big deal, but i think you are already making this assertion in `test_rpm_manifest_includes_minimal_collector_files` above?

integration-tests/test_package_verification.py
```diff
@@ -0,0 +1,286 @@
...
+        naming pattern.  Validates the pattern specifically for
+        com.redhat.minimal.
+    :tags: Tier 1
+    :steps:
+        1. Filter for rhc-collector-* units under the systemd unit directory
+        2. Verify each matches the rhc-collector-{id}.(service|timer) pattern
+        3. Verify com.redhat.minimal has both a .service and .timer unit
+    :expectedresults:
+        1. Collector units are found
+        2. All collector unit names match the expected pattern
```
stephenbourke: Maybe just detail the expected pattern here to provide a context for the regex?

You could point to the section in this doc below or just write something similar.

https://github.com/RedHatInsights/rhc/blob/main/doc/COLLECTORS.md#naming-conventions-and-artifact-paths

integration-tests/test_package_verification.py
```diff
@@ -0,0 +1,286 @@
...
+def test_rpm_manifest_includes_collector_directories(rpm_manifest):
+    """
+    :id: f1a2b3c4-d5e6-7890-abcd-100000000001
+    :title: Verify RPM manifest includes collector directory layout
+    :description:
+        Test that the rhc RPM file list includes the collector config and
+        binary directories.
+    :tags: Tier 1
+    :steps:
+        1. Verify the manifest includes /usr/lib/rhc/collectors/
```
stephenbourke: I think the assertions are made on directories with non-trailing slashes. 
AFAIK `rpm -ql` omits any trailing slashes.

integration-tests/test_package_verification.py
mjcr99: Good idea!

stephenbourke (dismissed):
It's unfortunate that the CI is down to get an idea. 
But the changes look good!

Archana-PandeyM (changes requested):
These are mostly verifying  RPM contents, packaging. Lets move these to https://github.com/RedHatInsights/rhc/blob/main/.github/workflows/rpmbuild.yml

.github/workflows/rpmbuild.yml
```diff
@@ -54,6 +54,37 @@ jobs:
           mkdir -p rpms
           cp ~/rpmbuild/RPMS/*/*.rpm rpms/
 
+      - name: "Identify rhc RPM"
+        id: find-rpm
+        run: |
+          RHC_RPM=$(ls rpms/rhc-[0-9]*.rpm | head -1)
```
stephenbourke: Do we always expect a single RPM file in the directory? I wonder if the `head -1` confuses things a bit. If there was a case where we had multiple versions, we can't be sure the right one by taking the first from  the `ls` glob. 

Would it be better to ensure only a single rpm exists by failing the glob if more than one match is found as an unexpected state?

.github/actions/verify-rpm-excludes-runtime/action.yml
```diff
@@ -0,0 +1,35 @@
...
+    description: "Path to the rhc RPM file"
+    required: true
+
+runs:
+  using: composite
+  steps:
+    - name: "Verify runtime paths are excluded"
+      shell: bash
+      run: |
+        rpm -qlp "${{ inputs.rpm }}" > manifest.txt
```
stephenbourke: Minor nit. This is used across a few of the steps. Maybe it's worth consolidating the output from this to a single step for reuse?

.github/actions/verify-rpm-excludes-runtime/action.yml
```diff
@@ -0,0 +1,35 @@
...
+
+        FAIL=0
+
+        if grep -qx '/var/cache/rhc/collectors/com.redhat.minimal.json' manifest.txt; then
```
stephenbourke: I wonder if it should be expanded out to the whole cache directory tree?

```
grep '^/var/cache/rhc' manifest.txt || true
```

Archana-PandeyM:
Commit subject says refactor(Rpmbuild) but  PR title is feat(CI). I think we should change commit subject.
This is new  verification, not a refactor.

.github/actions/verify-rpm-unit-naming/action.yml
```diff
@@ -0,0 +1,35 @@
...
+      run: |
+        rpm -qlp "${{ inputs.rpm }}" > manifest.txt
+
+        COLLECTOR_UNITS=$(grep '^/usr/lib/systemd/system/rhc-collector-' manifest.txt || true)
+        if [ -z "${COLLECTOR_UNITS}" ]; then
+          echo "FAIL  No rhc-collector-* units found in manifest"
+          exit 1
+        fi
+
+        PATTERN='^rhc-collector-[a-z][a-z0-9]*(\.[a-z][a-z0-9]*)+(\.service|\.timer)$'
```
Archana-PandeyM: nit:  This regex looks stricter than the collector ID rule in ValidateID / [COLLECTORS.md](https://github.com/RedHatInsights/rhc/blob/main/doc/COLLECTORS.md#naming-conventions-and-artifact-paths).

Product IDs use ^[a-z0-9]+\.[a-z0-9]+(\.[a-z0-9]+)*$, which allows a digit or number prefix. This pattern requires [a-z] at the start of every segment, so some valid IDs would fail like "com.redhat.2" or "2.example.collector"

Archana-PandeyM (dismissed):

.github/actions/verify-rpm-unit-naming/action.yml
mjcr99: Addressed!

.github/actions/verify-rpm-excludes-runtime/action.yml
mjcr99: Addressed!

.github/workflows/rpmbuild.yml
mjcr99: Addressed!

Archana-PandeyM (approved):

PR #555, opened 2026-08-14
test: Integration tests for RHC CLI error conditions and help/man pages

stephenbourke:
Card Id: CCT-2751

- Collector info/enable command - Missing/Invalid IDs
- Collector help and man pages

### DIFF
--- a/integration-tests/test_collector_cli.py
+++ b/integration-tests/test_collector_cli.py
@@ -375,6 +375,88 @@ def test_collector_cli_info_nonexistent_id(rhc):
 
     assert result.returncode != 0
     assert "failed to get collector info" in result.stderr
+    assert "varlink call failed: com.redhat.rhc.collector.NoSuchCollector" in result.stderr
+
+
+@pytest.mark.tier2
+def test_collector_cli_info_invalid_id(rhc):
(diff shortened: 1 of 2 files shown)


integration-tests/utils/systemctl.py
```diff
@@ -23,6 +23,20 @@ def is_service_active(service_name: str) -> bool:
     return result.returncode == 0
 
 
+def is_service_inactive(service_name: str) -> bool:
```
Archana-PandeyM: can we use existing method `if not is_service_active()`

Archana-PandeyM (changes requested):

integration-tests/utils/systemctl.py
stephenbourke: Thanks @Archana-PandeyM . Since you're covering it in https://redhat.atlassian.net/browse/CCT-2710, I'm going to remove the socket tests and the related fixture elements from this PR.

Archana-PandeyM (approved):
nit: Commit type feat(integration-tests) is a bit odd for test-only, the repo usually uses test:.

Everything else LGTM!

PR #556, opened 2026-08-14
fix(integration-tests): add Fedora 46 product cert for Rawhide CI

DuckBoss:
This patch aims to fix a versioning mismatch in the behave integration tests where the 
Fedora Rawhide version ID was pointing to an outdated release value. 
Fedora Rawhide [cut their F45 branch recently,](https://fedorapeople.org/groups/schedule/f-45/f-45-key-tasks.html) so this was most likely the cause.
Fedora Rawhide now reports VERSION_ID=46, which was missing from PRODUCT_CERTS_TABLE causing a KeyError in the default_product_cert_is_installed fixture.


jlocash (approved):
LGTM

PR #558, opened 2026-08-18
test: integration tests for minimal collector

stephenbourke:
* Card Id: CCT-2711

 - Successful run
 - Unregistered host
 - Failing minimal collector executable
 - Missing minimal collector executable


integration-tests/test_collector_cli.py
```diff
@@ -756,3 +759,174 @@ def test_collector_cli_help(rhc):
...
+@pytest.mark.tier1
+def test_minimal_collector_unregistered(rhc):
+    """
+    :id: 8d2f5b3c-0e41-5f91-c2b7-1a3d9e6f6b22
+    :title: Unregistered com.redhat.minimal run writes cache with matching non-zero exit_code
+    :description:
+        After disconnect, com.redhat.minimal collect fails because
+        /etc/pki/consumer/cert.pem is missing. rhc-collector must still write
+        the timer cache, and last_finished.exit_code must match the collector
+        binary's exit code.
```
zpetrace: The docstring says collect fails because cert.pem is missing. In cmd/rhc-collector/main.go, executeCollector() runs first and writes the timer cache; upload happens later in uploadArchive() and fails when the consumer cert cannot be loaded.

Suggest updating the description to: unregistered host → collect succeeds → upload fails.

integration-tests/test_collector_cli.py
```diff
@@ -756,3 +759,174 @@ def test_collector_cli_help(rhc):
...
+    )
+
+    combined_output = result.stdout + result.stderr
+    assert result.returncode != 0, combined_output
+    assert "failed to execute collector" in combined_output
+    assert "Successfully uploaded archive" not in combined_output
+    assert os.path.exists(cache_path)
+    with open(cache_path) as cache_file:
+        cache = json.load(cache_file)
+    assert cache["last_finished"]["exit_code"] == result.returncode
```
zpetrace: Likely causes RHEL TF failures. Timer cache last_finished.exit_code is the collector subprocess exit code (often 0 when collect succeeds). result.returncode is rhc-collector's exit code (exitcode.Err = 1 on upload failure). These are not the same on this path.

maybe:
```
assert result.returncode == 1
assert cache["last_finished"]["exit_code"] == 0
```

integration-tests/test_collector_cli.py
```diff
@@ -756,3 +759,174 @@ def test_collector_cli_help(rhc):
...
+    with open(cache_path) as cache_file:
+        cache = json.load(cache_file)
+    assert result.returncode != 0
+    assert cache["last_finished"]["exit_code"] == EXIT_CODE_MOCK_MINIMAL_COLLECTOR_EXECUTABLE
+
+
+@pytest.mark.tier1
+def test_minimal_collector_missing_executable(missing_minimal_collector_executable):
+    """
+    :id: 8d2f5b3c-0e41-5f91-c2b7-1a3d9e6f6b22
```
zpetrace, line 922: Duplicate Polarion :id: — same UUID as test_minimal_collector_unregistered (line 820). Each test needs a unique :id:

integration-tests/test_collector_cli.py
```diff
@@ -756,3 +759,174 @@ def test_collector_cli_help(rhc):
...
+    )
+
+    combined_output = result.stdout + result.stderr 
+    assert "failed to execute collector" in combined_output
+    assert "Successfully uploaded archive" not in combined_output
+    assert os.path.exists(cache_path)
+    with open(cache_path) as cache_file:
+        cache = json.load(cache_file)
+    assert result.returncode != 0
+    assert cache["last_finished"]["exit_code"] == EXIT_CODE_MOCK_MINIMAL_COLLECTOR_EXECUTABLE
```
zpetrace: Maybe trying `result.returncode != 0` before parsing the cache (same order as the other failure tests)? Also consider asserting `result.returncode == 1` explicitly because rhc-collector uses `exitcode.Err`, not the collector's exit code 42, no?

integration-tests/test_collector_cli.py
```diff
@@ -6,13 +6,15 @@
...
 
+import contextlib
 import json
 import os
 import subprocess
 import time
 
 import pytest
 
+from utils import prepare_args_for_connect
```
mjcr99: Could we consolidate the imports from `ùtils` in the same lane?

integration-tests/test_collector_cli.py
```diff
@@ -756,3 +562,174 @@ def test_collector_cli_help(rhc):
...
+        text=True,
+    )
+
+    combined_output = result.stdout + result.stderr 
+    assert "failed to execute collector" in combined_output
+    assert "Successfully uploaded archive" not in combined_output
+    assert os.path.exists(cache_path)
+    with open(cache_path) as cache_file:
+        cache = json.load(cache_file)
+    assert result.returncode != 0
```
mjcr99: Wouldn't it make more sense to first assert this condition and then move on with the rest?

mjcr99 (changes requested):
Thanks @stephenbourke! I have some comments on the PR 

`test_minimal_collector_unregistered`, `test_minimal_collector_executable_failure`, and `test_minimal_collector_missing_executable` all remove a stale cache before
 running, then assert the new cache was written, but never clean up the cache afterwards while `test_collector_enable_now_triggers_immediate_run` is there a reason for this difference?

integration-tests/test_collector_cli.py
```diff
@@ -756,3 +759,174 @@ def test_collector_cli_help(rhc):
...
+    """
+    :id: 8d2f5b3c-0e41-5f91-c2b7-1a3d9e6f6b22
+    :title: Unregistered com.redhat.minimal run writes cache with matching non-zero exit_code
+    :description:
+        After disconnect, com.redhat.minimal collect fails because
+        /etc/pki/consumer/cert.pem is missing. rhc-collector must still write
+        the timer cache, and last_finished.exit_code must match the collector
+        binary's exit code.
+    :tags: Tier 1
+    :steps:
```
zpetrace, line 832: The number of expected results must be the same as number of steps - each step should have its own expected results.

integration-tests/test_collector_cli.py
```diff
@@ -756,3 +759,174 @@ def test_collector_cli_help(rhc):
...
+    :id: 8d2f5b3c-0e41-5f91-c2b7-1a3d9e6f6b22
+    :title: Missing com.redhat.minimal executable runs writes cache with matching non-zero exit_code
+    :description:
+        com.redhat.minimal executable is missing, so collect fails.
+        rhc-collector must still write the timer cache, and last_finished.exit_code
+        must match the collector executable's exit code.
+    :tags: Tier 1
+    :steps:
+        1. Run rhc-collector run com.redhat.minimal
+        3. Verify rhc-collector exits non-zero, cache exists, and exit_code matches
```
zpetrace: Wrong numbering

integration-tests/conftest.py
```diff
@@ -356,3 +357,36 @@ def _configure_proxy(proxy_url):
...
+        echo "usage: $0 collect" >&2
+        exit 64
+    """)
+    os.rename(bin_path, backup_path)
+    with open(bin_path, "w") as f:
+        f.write(script)
+    os.chmod(bin_path, 0o755)
+    yield
+    os.remove(bin_path)
+    os.rename(backup_path, bin_path)
```
zpetrace: Maybe I am wrong here, but doesn't this mutates the installed `com.redhat.minimal` binary under /usr/libexec/...? Teardown is fine when everything goes right, but if setup fails after os.rename (line [376](https://github.com/RedHatInsights/rhc/pull/558/changes#diff-f99db5aabf9de5120c24409d7be1b79e635a807f5dc3c12de0116b66f5528be2R376)) or a prior run left a stale .bak, the packaged binary may not be restored
I'd suggest considering `try/finally` around setup/teardown, and guard with assert `os.path.exists(bin_path)` before rename, so the same pattern as `minimal_collector_slow_service` uses with `suppress(OSError)`

zpetrace (changes requested):

integration-tests/test_collector_cli.py
stephenbourke: Hmm good point. I think the mistake is trying to encapsulate the whole process flow as collector execution. I'll update this.

integration-tests/test_collector_cli.py
stephenbourke: Just noticed this. I've updated to `!= 0`. It seems to follow conventions in pre-existing tests.

stephenbourke:
> Thanks @stephenbourke! I have some comments on the PR
> 
> `test_minimal_collector_unregistered`, `test_minimal_collector_executable_failure`, and `test_minimal_collector_missing_executable` all remove a stale cache before running, then assert the new cache was written, but never clean up the cache afterwards while `test_collector_enable_now_triggers_immediate_run` is there a reason for this difference?

No. Just something I missed. Thanks for spotting.  I see `try/finally` blocks used to handle this in the other tests, so I'll add this.

integration-tests/test_collector_cli.py
stephenbourke: We can't have everything on the same line as there are imports from a sub-packages. But we can move them together to make it a bit tidier...

```
from utils import prepare_args_for_connect, poll_until
from utils.systemctl import is_unit_active, is_unit_enabled
from utils.constants import (
...
)
```

integration-tests/conftest.py
stephenbourke: You're right. I'll put this in a `try/finally block` and add the extra check before rename.

integration-tests/test_collector_cli.py
stephenbourke: Actually this is incorrect. The minimal collector process fails on `GetCanonicalFacts` (https://github.com/RedHatInsights/rhc/blob/main/cmd/minimal-collector/minimal-collector.go#L257) which depends on the cert to be present. So com.redhat.minimal binary fails before `rhc-collector` gets to upload.

zpetrace (approved):
LGTM, thanks for the changes:)

mjcr99 (approved):
LGTM, GJ!

PR #561, opened 2026-08-24
refactor(status): move checks to operations

Peto-RH:
* CardID: CCT-2407

Move status collection behind a clean GetStatus API while keeping
dependency injection private for tests. Keep formatting, spinner, and
exit-code handling in the CLI, rendering output after collection completes.

### DIFF
--- a/cmd/rhc/status_cmd.go
+++ b/cmd/rhc/status_cmd.go
@@ -2,182 +2,14 @@ package main
 
 import (
 	"context"
-	"encoding/json"
 	"fmt"
-	"log/slog"
-	"os"
 
 	"github.com/urfave/cli/v3"
 
-	"github.com/redhatinsights/rhc/internal/datacollection"
-	"github.com/redhatinsights/rhc/internal/remotemanagement"
-	"github.com/redhatinsights/rhc/internal/subman"
 	"github.com/redhatinsights/rhc/internal/ui"
 	"github.com/redhatinsights/rhc/pkg/exitcode"
+	"github.com/redhatinsights/rhc/pkg/operations"
 )
-
-// rhsmStatus tries to print status provided by RHSM D-Bus API. If we provide
-// output in machine-readable format, then we only set files in SystemStatus
-// structure and content of this structure will be printed later
-func rhsmStatus(systemStatus *SystemStatus) error {
-	slog.Info("Checking status of Red Hat Subscription Management")
-
-	client, err := subman.NewRHSMClient()
-	if err != nil {
-		systemStatus.returnCode += 1
-		systemStatus.RHSMError = err.Error()
-		return fmt.Errorf("unable to check registration status: %s", err)
-	}
-	registered, err := client.IsRegistered()
-	if err != nil {
-		systemStatus.returnCode += 1
-		systemStatus.RHSMError = err.Error()
-		return fmt.Errorf("unable to check registration status: %s", err)
-	}
-	if !registered {
-		systemStatus.returnCode += 1
-		systemStatus.RHSMConnected = false
-		infoMsg := "Not connected to Red Hat Subscription Management"
-		slog.Info(infoMsg)
-		ui.Printf("%s[ ] %v\n", ui.Indent.Small, infoMsg)
-	} else {
-		systemStatus.RHSMConnected = true
-		infoMsg := "Connected to Red Hat Subscription Management"
-		slog.Info(infoMsg)
-		ui.Printf("%s[%v] %v\n", ui.Indent.Small, ui.Icons.Ok, infoMsg)
-	}
-	return nil
-}
-
-// isContentEnabled reports whether the system has access to RHSM content.
-// It relies on systemStatus.RHSMConnected already being populated by rhsmStatus.
-func isContentEnabled(systemStatus *SystemStatus) error {
-	slog.Info("Checking content status")
-
-	client, err := subman.NewRHSMClient()
-	if err != nil {
-		systemStatus.returnCode += 1
-		systemStatus.ContentError = err.Error()
-		return fmt.Errorf("unable to check content management: %w", err)
-	}
-	contentEnabled, err := client.IsContentManagementEnabled()
-	if err != nil {
-		systemStatus.returnCode += 1
-		systemStatus.ContentError = err.Error()
-		return fmt.Errorf("unable to check content management: %w", err)
-	}
-
-	if contentEnabled && systemStatus.RHSMConnected {
-		systemStatus.ContentEnabled = true
-		infoMsg := "System has access to content"
-		slog.Info(infoMsg)
-		ui.Printf("%s[%v] Content ... %v\n", ui.Indent.Medium, ui.Icons.Ok, infoMsg)
-	} else {
-		systemStatus.ContentEnabled = false
-		infoMsg := "System has no access to content"
-		slog.Info(infoMsg)
-		ui.Printf("%s[ ] Content ... %v\n", ui.Indent.Medium, infoMsg)
-	}
-	return nil
-}
-
-// insightStatus tries to print status of insights client
-func insightStatus(systemStatus *SystemStatus) error {
-	slog.Info("Checking status of Red Hat Lightspeed")
-
-	var isRegistered bool
-	var err error
-	spinErr := ui.Spinner(func() error {
-		isRegistered, err = datacollection.InsightsClientIsRegistered()
-		return nil
-	}, ui.Indent.Medium, "Checking Red Hat Lightspeed (formerly Insights)...")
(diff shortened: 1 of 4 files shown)


cmd/rhc/status_cmd.go
```diff
@@ -2,183 +2,15 @@ package main
...
-// It relies on systemStatus.RHSMConnected already being populated by rhsmStatus.
-func isContentEnabled(systemStatus *SystemStatus) error {
-	slog.Info("Checking content status")
-
-	client, err := subman.NewRHSMClient()
-	if err != nil {
-		systemStatus.returnCode += 1
-		systemStatus.ContentError = err.Error()
-		return fmt.Errorf("unable to check content management: %w", err)
-	}
```
zpetrace, line 62: Formatters are well tested, but `statusAction` itself isn’t (spinner, human path, JSON branch). Even a small test asserting exit code + stdout for `--format json` would be good IMO

pkg/operations/status.go
```diff
@@ -0,0 +1,189 @@
...
+	}
+}
+
+func contentIsEnabled() (bool, error) {
+	service, err := subman.NewRHSMClient()
+	if err != nil {
+		return false, err
+	}
+	return service.IsContentManagementEnabled()
+}
```
zpetrace, line 40: Doesn't `contentIsEnabled` open a second `subman.NewRHSMClient()` while registration uses `IsRegistered()`? Wouldn't it be better share one client per `getStatus()` call?

cmd/rhc/status_cmd_test.go
```diff
@@ -0,0 +1,116 @@
...
+func TestFormatStatusLines(t *testing.T) {
+	tests := []struct {
+		name   string
+		render func()
+		want   string
+	}{
+		{
+			name:   "rhsm connected",
+			render: func() { formatRHSMStatus(operations.StatusReport{RHSMConnected: true}) },
+			want:   " [✓] Connected to Red Hat Subscription Management\n",
```
zpetrace: just a suggestion: `want` strings pin literal `✓` / `𐄂` but IMO it would be better to use  `ui.Icons.Ok` / `ui.Icons.Error` in expected output, or assert on message text with `strings.Contains`

zpetrace: The same goes for lines 57,72

cmd/rhc/status_cmd_test.go
```diff
@@ -0,0 +1,116 @@
...
+	fn()
+	_ = writer.Close()
+	out, _ := io.ReadAll(reader)
+	_ = reader.Close()
+	return string(out)
+}
+
+// TestFormatStatusLines pins the exact human-readable line each formatter prints
+// for its success, negative, and error states.
+func TestFormatStatusLines(t *testing.T) {
```
zpetrace, line 79: Only `format*` helpers are tested here — see `status_cmd.go:57` for `statusAction` / JSON coverage.

pkg/operations/status.go
```diff
@@ -0,0 +1,189 @@
...
+		return
+	}
+
+	if contentEnabled && report.RHSMConnected {
+		report.ContentEnabled = true
+		slog.Info("System has access to content")
+	} else {
+		report.ContentEnabled = false
+		slog.Info("System has no access to content")
+	}
```
zpetrace, line 134: `ContentEnabled` requires `contentEnabled && RHSMConnected`. Content disabled alone doesn’t increment `failedChecks` (parity with old `returnCode`). Maybe worth a test like RHSM not registered + content enabled → `ContentEnabled == false`, `failedChecks == 1`?

zpetrace (changes requested):
Mainly looking good, I am not that oriented in the code itself so my points are directed mainly towards testing but apart from few minor suggestions it looks good, thanks!:)

pkg/operations/status.go
Peto-RH: I did not want to include this change in the PR because it is outside the scope of the refactor; the old implementation also called `subman.NewRHSMClient()` twice. The diff is already large, and I did not want to make it even larger. In my opinion, this would be a good candidate for a separate PR. What do you think?

cmd/rhc/status_cmd_test.go
Peto-RH: Good idea to use the constants. I also updated the other lines that used the literal `✓` and `𐄂` symbols. Would you also like me to use `strings.Contains` in this test? I think exact equality is valuable here. What do you think?

cmd/rhc/status_cmd_test.go
Peto-RH: I added a test for the JSON output, while the existing tests cover the human-readable status lines. Do you think this is enough, or do you think of any other test that would be useful to add?

pkg/operations/status.go
zpetrace: yeah, that sounds reasonable:)

cmd/rhc/status_cmd_test.go
zpetrace: Yeah I think that should be enough, thanks.

cmd/rhc/status_cmd_test.go
zpetrace: Yeah I think that could also be useful, good idea:)

zpetrace: for clarity - I'd keep exact equality here; the test is explicitly pinning the full formatted line, and strings.Contains would be too loose for that. Looks good from my side:)

zpetrace (dismissed):

pkg/operations/status.go
```diff
@@ -0,0 +1,189 @@
...
+	report *StatusReport,
+	getUnitState func(string) (*remotemanagement.UnitState, error),
+) {
+	slog.Info("Checking status of yggdrasil service")
+
+	state, err := getUnitState("yggdrasil.service")
+	if err != nil {
+		report.YggdrasilRunning = false
+		report.YggdrasilError = err.Error()
+		return
```
stephenbourke, line 168: We likely get this error when Systemd is unreachable from what I understand. I think it's still necessary to produce a non-zero exit. So maybe `report.failedChecks` needs to be incremented here?

pkg/operations/status.go
```diff
@@ -0,0 +1,189 @@
...
+
+	collectRHSMStatus(report, dependencies.RHSMIsRegistered)
+	collectContentStatus(report, dependencies.ContentIsEnabled)
+	collectInsightsStatus(report, dependencies.InsightsClientIsRegistered)
+	collectYggdrasilStatus(report, dependencies.GetUnitState)
+
+	return report
+}
+
+func collectHostname(report *StatusReport, hostname func() (string, error)) {
```
stephenbourke, line 89: Should this also increment `report.failedChecks`? If we can't find a hostname, for the machine readable output (`ui.IsOutputMachineReadable()`), it looks like it will not exit with non-zero as `report.hasFailures` will be false.

pkg/operations/status_test.go
```diff
@@ -0,0 +1,142 @@
+package operations
```
stephenbourke, line 1: Maybe another negative test can be added here to test the third logical branch of `collectYggdrasilStatus`, where the yggdrasil service activeState is "inactive"  and the loadState is "not loaded" or "not found".

pkg/operations/status.go
Peto-RH: This makes sense, but the previous implementation also returned zero here, so I intentionally kept the behavior unchanged. Would you prefer to fix it in this PR or address it separately?

pkg/operations/status_test.go
Peto-RH: Yes, this makes sense

pkg/operations/status.go
Peto-RH: Good point. A non-zero exit makes sense when the systemd status cannot be determined. However, the previous implementation also returned zero in this case, so I kept the behavior unchanged. Would you prefer to fix it in this PR or address it separately?

stephenbourke: You're right. It's probably worth keeping existing behavior and having a separate ticket.

pkg/operations/status.go
stephenbourke: Same comment as above. I think keeping existing behavior and having a separate issue makes sense.

Peto-RH:
Thank you @stephenbourke for the review! Is there anything more I could change?

stephenbourke:
> Thank you @stephenbourke for the review! Is there anything more I could change?

Hi @Peto-RH . The refactor looks good. The comments I had on the exit code returns  can be addressed in another ticket.

stephenbourke (approved):
LGTM

PR #564, opened 2026-08-31
test: pin compose integration tests to the installed rhc release tag

Archana-PandeyM:
* Card ID: CCT-981

Compose / CTC / RC clones a moving git ref (`main`) while the compose still has an older `rhc` RPM. Pytest then asserts current-upstream behavior against a package that does not have it yet. PR/Packit stay green because tests and the Copr RPM share the same git.

With this PR When `PIN_TESTS_TO_RPM=1`, `systemtest/tests/integration/test.sh` reads `rpm -q --qf '%{VERSION}' rhc`, fetches tag `v$VERSION` (or unprefixed `$VERSION` for 0.2.x), **replaces** `integration-tests/` with that tree, then runs pytest. `test.sh` itself stays on `TEST_BRANCH` so the pin logic can run on tags that never had it.

 PIN_TESTS_TO_RPM is implemented in  downstream jenkins.

### DIFF
--- a/systemtest/tests/integration/test.sh
+++ b/systemtest/tests/integration/test.sh
@@ -1,8 +1,10 @@
 #!/bin/bash
 set -ux
 
-# get to project root
-cd ../../../
+# Resolve repo root from this script so it works both when tmt runs it from
+# systemtest/tests/integration/ and when invoked from the repository root.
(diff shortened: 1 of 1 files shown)


pkoprda (approved):
LGTM, this is great improvement!

PR #565, opened 2026-09-02
test: cover RHSM connect spinner Stop() ordering on a TTY

Archana-PandeyM:
Add a pexpect-based PTY helper (utils/tty.py) so tests can exercise rhc's spinner UI, which only activates when stdout is a real terminal and is invisible to the pipe-based rhc.run(). Cover both the success and failure paths of TryRegisterRHSM: the spinner must Stop() (cursor restore + line erase) before the final "Connected"/"Cannot connect" line is printed, not after.

Card Id : RHEL-244670

Assisted-by: Cursor

### DIFF
--- a/integration-tests/requirements.txt
+++ b/integration-tests/requirements.txt
@@ -1,3 +1,4 @@
 git+https://github.com/RedHatInsights/pytest-client-tools@main
+pexpect
 pyyaml
 sh
--- a/integration-tests/test_connect.py
+++ b/integration-tests/test_connect.py
@@ -9,6 +9,7 @@
 import contextlib
 import json
 import logging
+import re
 
 import pytest
 from pytest_client_tools.restclient import RestClient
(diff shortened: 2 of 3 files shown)


mjcr99 (commented):
Thanks @Archana-PandeyM it looks good to me, just a couple of minnor questions: 
 
* Should the `test_connect_tty_spinner_stops_on_success` case implement a final teardown/cleanup after PTY-based connect?

* This may be a very edge case, but I think the spinner frame assertion is timing-dependent, in a case where `rhc connect` completes extremely fast the spinner might not paint any frames before `Stop()`. This would cause the assert frames assertion to fail with a misleading error about "expected RHSM spinner frames" when the real issue is just speed. I know this is probably very unlikely to happen, but I was wondering if it's worth to handle.

Archana-PandeyM:
> * Should the `test_connect_tty_spinner_stops_on_success` case implement a final teardown/cleanup after PTY-based connect?

@mjcr99  the test takes rhc as a fixture argument, pytest_client_tools will call rhc.disconnect() for you once the test function returns, whether it passed or failed.

Archana-PandeyM:
> * This may be a very edge case, but I think the spinner frame assertion is timing-dependent, in a case where `rhc connect` completes extremely fast the spinner might not paint any frames before `Stop()`. This would cause the assert frames assertion to fail with a misleading error about "expected RHSM spinner frames" when the real issue is just speed. I know this is probably very unlikely to happen, but I was wondering if it's worth to handle.

Frame painting only needs one goroutine scheduling opportunity, not a full 100ms — the time.Sleep(100ms) in briandowns/spinner's loop happens after each frame write, not before the first one. And TryRegisterRHSM's registration call always blocks on a D-Bus→HTTPS round-trip to Candlepin before it can return, which guarantees the scheduler yields (and the spinner goroutine gets to paint) well before that. For zero frames to actually occur, registration would have to fail synchronously without touching D-Bus/network at all — neither of our two test cases (valid activation-key, invalid-key-rejected-by-server) can hit that. Given the margin (µs scheduling vs. ms+ network round-trip), I'll leave the stricter assertion for now .

That said I would be happy to revisit if it ever actually flakes in CI.

mjcr99 (approved):
Thanks for the clarifications! Approved!

PR #567, opened 2026-09-03
fix: Refresh varlink method should work properly now

jirihnidek:
* When Refresh() Varlink method was called, then rhsm2
  wrote only first block of entitlement certificate.
  This issue was fixed in upstream of rhsm2:
  https://github.com/jirihnidek/rhsm2/pull/111
* Updated rhsm2 to the latest version
* Added `test_content.py` with more testing steps
* Extended integration tests for the case, when entitlement
  is really updated on the candlepin server
* Added integration test for the issue fixed in:
  https://github.com/jirihnidek/rhsm2/pull/111

### DIFF
--- a/features/com.redhat.rhsm.testing.content.feature
+++ b/features/com.redhat.rhsm.testing.content.feature
@@ -30,8 +30,31 @@ Feature: The Varlink interface com.redhat.rhsm.testing.content
       """
 
 
+  Scenario: Refresh() method succeeds on registered system after regenerating of entitlement certificate
+    Given system is registered against candlepin server
+    Given entitlement certificate is installed
+    Given entitlement certificate is regenerated on candlepin server
+    # We have to wait one second here, because If-Modified-Since HTTP header can be sent only in
+    # time format defined in RFC 822, which has only second precision. If Refresh() method was
+    # called immediately, then it would lead to 304 HTTP code (not modified), because the time
+    # of creating old and new certificate with second precision would be the same.
+    Then wait '1' seconds
+    When varlink method is called
+      | method  | interface                       | arguments          |
+      | Refresh | com.redhat.rhsm.testing.content | '{"force": false}' |
+    Then varlink method returns
+      """
+      {"success":true}
+      """
+    And entitlement certificate has different name and i-node
+    And installed entitlement certificate contains 'BEGIN CERTIFICATE'
+    And installed entitlement certificate contains 'BEGIN ENTITLEMENT DATA'
+    And installed entitlement certificate contains 'BEGIN SIGNATURE'
+
+
   Scenario: Refresh() method succeeds on registered system with metadata
     Given system is registered against candlepin server
+    Given entitlement certificate is installed
     When varlink method is called
       | method  | interface                       | arguments                                             |
       | Refresh | com.redhat.rhsm.testing.content | '{"force": false, "metadata": {"user_agent": "foo"}}' |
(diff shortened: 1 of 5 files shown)


features/com.redhat.rhsm.testing.content.feature
```diff
@@ -30,41 +30,87 @@ Feature: The Varlink interface com.redhat.rhsm.testing.content
...
   Scenario: Refresh() method succeeds on registered system with metadata
     Given system is registered against candlepin server
+    Given entitlement certificate is installed
     When varlink method is called
       | method  | interface                       | arguments                                             |
       | Refresh | com.redhat.rhsm.testing.content | '{"force": false, "metadata": {"user_agent": "foo"}}' |
     Then varlink method returns
       """
       {"success":true}
       """
```
jirihnidek, line 64: @DuckBoss I was thinking what we are returning from this method and the returned data are useless and misleading. The method returns `{"success":true}` despite the entitlement certificate is still the same. Shouldn't we rather return JSON document with more details about updated entitlement certificate and key? Something like:

```json
{
  "certificate_updated": true,
  "certificate_path": "/etc/pki/entitlement/3445391557400405333.pem",
  "key_updated": true,
  "key_path": "/etc/pki/entitlement/3445391557400405333-key.pem"
}
```

features/com.redhat.rhsm.testing.content.feature
```diff
@@ -30,41 +30,87 @@ Feature: The Varlink interface com.redhat.rhsm.testing.content
...
 
 
+  Scenario: Refresh() method succeeds on registered system after regenerating of entitlement certificate
+    Given system is registered against candlepin server
+    Given entitlement certificate is installed
+    Given entitlement certificate is regenerated on candlepin server
+    Then wait '1' seconds
```
jirihnidek, line 41: We have to wait 1 second after calling REST API call that is implemented in `test_content.py`:

```python
url = f"https://{hostname}:{port}{prefix}/consumers/{consumer_uuid}/certificates?lazy_regen=false"
response = requests.put(url, auth=('admin', 'admin'), verify=False)
response.raise_for_status()
```

Otherwise, the entitlement certificate is still the same. I am wondering why?

It is also interesting that when the system is already registered and you call this scenario using `behave ./featurs/com.redhat.rhsm.testing.content.feature:33`, then it is not necessary to wait one second.

features/com.redhat.rhsm.testing.content.feature
cnsnyder: I'm inclined to are with you. Reporting "success":true, doesn't convey much and it seems likely that any caller of refresh would want to know at least the details you've suggested.

jirihnidek: OK. If we will have consent about this, then we should implement it in another PR. Agreed?

pbartiko: I’m fine with another PR for that. I’d keep this one on the rhsm2 bump and the regen tests.

pbartiko (approved):
Looks good. The rhsm2 bump is the actual fix, and the new regen tests cover the case where the cert really changes.

mjcr99 (approved):
ACK

Archana-PandeyM (approved):
go.mod and go.sum these don't have a specific rule, so they fall under defaults, rest of the changes are reviewed by Void team.

PR #568, opened 2026-09-07
ci: drop ARM64 from upstream PR tests

Peto-RH:
* Card ID: CCT-2729

Run Packit integration tests only on x86_64 while retaining ARM64 targets for RPM builds.


pkoprda (approved):
LGTM, thanks!

PR #569, opened 2026-09-07
fix(tests): make collector tests work on image mode

pkoprda:
* Card ID: CCT-2785

Collector integration tests fail on image mode because fixtures write to `/usr`, which is read-only there. Fix by replacing fake collectors with the shipped `com.redhat.minimal` where possible. Adding fixture `writable_collector_dirs` that mounts an overlayfs over the read-only `/usr` directories using `is_bootc_system()` to detect image mode. Skipping `test_collector_list_with_multiple_collectors` on image mode as it requires a second synthetic collector in `/usr`.

### DIFF
--- a/integration-tests/conftest.py
+++ b/integration-tests/conftest.py
@@ -3,9 +3,12 @@ import subprocess
 import logging
 import os
 import json
+import shutil
+import tempfile
 import time
 import textwrap
 from contextlib import suppress
+from functools import lru_cache
 
 from utils.systemctl import is_unit_active, is_unit_enabled
 from utils.constants import (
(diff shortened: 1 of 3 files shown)


pkoprda:
Image mode pipeline: https://jenkins-csb-rhsm-prod.dno.corp.redhat.com/view/CCT/job/rhc/job/imagemode/232/

integration-tests/test_collector_cli.py
```diff
@@ -70,7 +71,7 @@ def test_rhc_collector_writes_timer_cache(collector_config):
 
         assert "last_started" in cache
         assert "last_finished" in cache
-        assert cache["last_finished"]["exit_code"] == 0
+        assert "exit_code" in cache["last_finished"]
```
mjcr99: A real regression that causes rhc-collector run to fail could now pass silently; I would keep checking the value.

integration-tests/conftest.py
```diff
@@ -25,6 +27,73 @@
...
+    any files the test created.
+    """
+    if not check_is_bootc_system():
+        yield
+        return
+
+    mounted = []
+    tmpdirs = []
+
+    for target in [COLLECTOR_CONFIG_DIR, COLLECTOR_BIN_DIR]:
```
Archana-PandeyM: just in case if mounting the second directory fails, the code after yield won't execute, which could leave the first overlay mount and temporary directory on the test system and pollute other tests.
I suggest we wrap the setup and teardown in a try/finally?

integration-tests/conftest.py
```diff
@@ -25,6 +27,73 @@
 logger = logging.getLogger(__name__)
 
 
+def check_is_bootc_system():
```
mjcr99: This is a nit, but since the result of this computation will not change, we could somehow cache it.

integration-tests/test_collector_api.py
```diff
@@ -362,6 +363,7 @@ def test_collector_list_includes_minimal_collector():
 
 
 @pytest.mark.tier2
+@pytest.mark.skipif(check_is_bootc_system(), reason="/usr is not writable on bootc systems")
```
mjcr99: Shouldn't this test be able to work using `collector_minimal` which already uses `writable_collector_dirs` ?

mjcr99 (changes requested):
Thanks @pkoprda! A very minor comment and a couple of questions.

integration-tests/conftest.py
Archana-PandeyM: NIT: can we rename check_is_bootc_system() because the function returns a boolean rather than performing a check with side effects. we can call it is_bootc_system() or is_bootc()

Archana-PandeyM: Since is_bootc_system() may be called multiple times during the test session, would it make sense to cache the result with functools.lru_cache?

integration-tests/test_collector_api.py
pkoprda: It can use that, but what are the user stories for that? Isn't that just bypassing the actual implementation of image mode system?

Archana-PandeyM (dismissed):
LGTM!

mjcr99 (dismissed):

integration-tests/test_collector_cli.py
pkoprda: Since this test now uses the shipped minimal collector, it would need to be registered to rhc. But this test is already being tested in [test_collector_cli.py::test_minimal_collector_upload_production](https://github.com/RedHatInsights/rhc/blob/b12678926e09383f4e301088befb1a45eff767da/integration-tests/test_collector_cli.py#L816), so I think it's not a problem to keep here this assert: `assert "exit_code" in cache["last_finished"]`

integration-tests/test_collector_api.py
pkoprda: Let's skip this test for now on image mode, we can come back to it later.

mjcr99 (approved):

PR #570, opened 2026-09-11
feat(selinux): add in-tree module skeleton for collector domains

zpetrace:
Declare rhc_collector_t and rhc_collector_minimal_t plus the collector file types in selinux/rhc.{te,fc,if}. Export domain-transition interfaces for init_t -> rhc_collector_t and rhc_collector_t -> rhc_collector_minimal_t. Leave the full allow set for later work.

Resolves: CCT-2894

### DIFF
--- a/selinux/rhc.fc
+++ b/selinux/rhc.fc
@@ -1,0 +1,14 @@
+# rhc-collector systemd oneshot binary
+/usr/libexec/rhc/rhc-collector			--	gen_context(system_u:object_r:rhc_collector_exec_t,s0)
+
+# com.redhat.minimal collector binary
+/usr/libexec/rhc/collectors/com\.redhat\.minimal	--	gen_context(system_u:object_r:rhc_collector_minimal_exec_t,s0)
+
+# Collector configuration (TOML) shipped under /usr/lib/rhc/collectors
+/usr/lib/rhc/collectors(/.*)?				gen_context(system_u:object_r:rhc_collector_conf_t,s0)
+
+# Collector timer cache
+/var/cache/rhc/collectors(/.*)?				gen_context(system_u:object_r:rhc_cache_t,s0)
+
+# Collector working directory
+/var/tmp/rhc(/.*)?					gen_context(system_u:object_r:rhc_tmp_t,s0)
--- a/selinux/rhc.if
+++ b/selinux/rhc.if
@@ -1,0 +1,54 @@
+## <summary>SELinux policy for rhc-collector domains</summary>
+
+########################################
+## <summary>
+##	Execute rhc-collector in the rhc_collector_t domain.
+## </summary>
+## <desc>
+##	<p>
+##	Public API for transitioning into rhc_collector_t when executing
+##	rhc_collector_exec_t. Not invoked from rhc.te: init_daemon_domain()
+##	already wires init_t (systemd) for the oneshot runner. Exported for
+##	CCT-2894 and for other policy modules that need this transition.
+##	</p>
+## </desc>
+## <param name="domain">
+##	<summary>
+##	Domain allowed to transition.
+##	</summary>
+## </param>
+#
+interface(`rhc_collector_domtrans',`
+	gen_require(`
+		type rhc_collector_t, rhc_collector_exec_t;
+	')
+
+	corecmd_search_bin($1)
+	domtrans_pattern($1, rhc_collector_exec_t, rhc_collector_t)
+')
+
+########################################
+## <summary>
+##	Execute com.redhat.minimal in the rhc_collector_minimal_t domain.
+## </summary>
+## <desc>
+##	<p>
+##	Automatic domain transition to rhc_collector_minimal_t when executing
+##	rhc_collector_minimal_exec_t. Sufficient for rhc_collector_t to start
+##	the child collector confined.
+##	</p>
+## </desc>
+## <param name="domain">
+##	<summary>
+##	Domain allowed to transition.
+##	</summary>
+## </param>
+#
+interface(`rhc_collector_minimal_domtrans',`
+	gen_require(`
+		type rhc_collector_minimal_t, rhc_collector_minimal_exec_t;
+	')
+
+	corecmd_search_bin($1)
+	domtrans_pattern($1, rhc_collector_minimal_exec_t, rhc_collector_minimal_t)
+')
--- a/selinux/rhc.te
+++ b/selinux/rhc.te
@@ -1,0 +1,47 @@
+policy_module(rhc, 1.0.0)
+
+########################################
+#
+# Declarations
+#
+# Skeleton only: process domains and file types. Do not encode the full
+# allow set here; that is later work.
+#
+
+#
+# rhc_collector_t — systemd oneshot runner (/usr/libexec/rhc/rhc-collector)
+#
+type rhc_collector_t;
+type rhc_collector_exec_t;
+init_daemon_domain(rhc_collector_t, rhc_collector_exec_t)
+
+#
+# rhc_collector_minimal_t — com.redhat.minimal child collector
+#
+type rhc_collector_minimal_t;
+type rhc_collector_minimal_exec_t;
+application_domain(rhc_collector_minimal_t, rhc_collector_minimal_exec_t)
+role system_r types rhc_collector_minimal_t;
+
+#
+# File types
+#
+type rhc_collector_conf_t;
+files_config_file(rhc_collector_conf_t)
+
+type rhc_cache_t;
+files_type(rhc_cache_t)
+
+type rhc_tmp_t;
+files_tmp_file(rhc_tmp_t)
+
+########################################
+#
+# Domain transitions
+#
+# init_t -> rhc_collector_t is established by init_daemon_domain() above.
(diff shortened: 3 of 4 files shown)


selinux/rhc.te
```diff
@@ -0,0 +1,49 @@
...
+# Skeleton only: process domains and file types. Do not encode the full
+# allow set here; that is later work.
+#
+
+#
+# rhc_collector_t — systemd oneshot runner (/usr/libexec/rhc/rhc-collector)
+#
+type rhc_collector_t;
+type rhc_collector_exec_t;
+corecmd_executable_file(rhc_collector_exec_t)
```
stephenbourke: From what I understand, `init_daemon_domain` and `application_domain` will run the `corecmd_executable_file` path. So the usage of `corecmd_executable_file` is probably be redundant here. Though I don't think there are any averse effects having it explicitly declared.

selinux/rhc.fc
```diff
@@ -0,0 +1,14 @@
+# rhc-collector systemd oneshot binary
+/usr/libexec/rhc/rhc-collector			--	gen_context(system_u:object_r:rhc_collector_exec_t,s0)
+
+# com.redhat.minimal collector binary
+/usr/libexec/rhc/collectors/com\.redhat\.minimal	--	gen_context(system_u:object_r:rhc_collector_minimal_exec_t,s0)
+
+# Collector configuration (TOML) shipped under /usr/lib/rhc/collectors
+/usr/lib/rhc/collectors(/.*)?				gen_context(system_u:object_r:rhc_conf_t,s0)
```
stephenbourke: Should we name this `rhc_collector_conf_t` to avoid ambiguity with other configs? Thinking of the upcoming `rhc.conf` work.

selinux/rhc.if
```diff
@@ -0,0 +1,53 @@
...
+##	the oneshot runner confined.
+##	</p>
+## </desc>
+## <param name="domain">
+##	<summary>
+##	Domain allowed to transition.
+##	</summary>
+## </param>
+#
+interface(`rhc_domtrans_collector',`
```
stephenbourke: Would `rhc_collector_domtrans` be a more intuitive name here? Following the `rhc_collector_*` conventions set elsewhere in this PR.

selinux/rhc.fc
zpetrace: Yeah, that's actually a very good point, will do.

selinux/rhc.if
zpetrace: Sure, makes sense.

selinux/rhc.te
zpetrace: Yeah you're right, they are redundant. `init_daemon_domain()` and `application_domain()` already go through domain_entry_file(), which calls `corecmd_executable_file()`. I thought that keeping them there would be more explicit but it's true that the usual style is to omit them. Deleted:)

selinux/rhc.if
```diff
@@ -0,0 +1,53 @@
...
+##	the oneshot runner confined.
+##	</p>
+## </desc>
+## <param name="domain">
+##	<summary>
+##	Domain allowed to transition.
+##	</summary>
+## </param>
+#
+interface(`rhc_collector_domtrans_collector',`
```
stephenbourke: I think maybe drop the `_collector` suffix and have `rhc_collector_domtrans`

selinux/rhc.if
```diff
@@ -0,0 +1,53 @@
...
+##	the child collector confined.
+##	</p>
+## </desc>
+## <param name="domain">
+##	<summary>
+##	Domain allowed to transition.
+##	</summary>
+## </param>
+#
+interface(`rhc_domtrans_collector_minimal',`
```
stephenbourke: Following from above, I suggest `rhc_collector_minimal_domtrans`.

selinux/rhc.te
```diff
@@ -0,0 +1,47 @@
...
+########################################
+#
+# Domain transitions
+#
+# init_t -> rhc_collector_t is established by init_daemon_domain() above
+# (and equivalently by rhc_domtrans_collector(init_t) from rhc.if).
+# rhc_collector_t -> rhc_collector_minimal_t uses the exported interface.
+#
+
+rhc_domtrans_collector_minimal(rhc_collector_t)
```
stephenbourke: If the change if the `.fc` seems reasonable, it will be `rhc_collector_minimal_domtrans` here.

selinux/rhc.te
```diff
@@ -0,0 +1,47 @@
...
+type rhc_collector_minimal_t;
+type rhc_collector_minimal_exec_t;
+application_domain(rhc_collector_minimal_t, rhc_collector_minimal_exec_t)
+role system_r types rhc_collector_minimal_t;
+
+#
+# File types
+#
+type rhc_conf_t;
+files_config_file(rhc_conf_t)
```
mjcr99: I think we should use the updated `rhc_collector_conf_t` here as well.

selinux/rhc.te
```diff
@@ -0,0 +1,47 @@
...
+
+type rhc_tmp_t;
+files_tmp_file(rhc_tmp_t)
+
+########################################
+#
+# Domain transitions
+#
+# init_t -> rhc_collector_t is established by init_daemon_domain() above
+# (and equivalently by rhc_domtrans_collector(init_t) from rhc.if).
```
mjcr99: The interface `rhc_domtrans_collector` doesn't exist. In the updated .if, the interface that transitions into `rhc_collector_t` is named `rhc_collector_domtrans`. So, the comment should say:
```console
  # (and equivalently by rhc_collector_domtrans(init_t) from rhc.if).
```

mjcr99 (changes requested):
Thanks @zpetrace just a couple of comments.

Also a suggestion, even with the error in the line 30 on the `.te` file, if the policy compiles:

```console
macano@macano-thinkpadt14gen5:~/Desktop/lv_vagrant/share/repos/rhc/selinux$ make -f /usr/share/selinux/devel/Makefile rhc.pp
Compiling targeted rhc module
Creating targeted rhc.pp policy package
rm tmp/rhc.mod.fc tmp/rhc.mod
```

It's when it's tried to be loaded when it fails, maybe it's not worth to implement in this very first card, but I think we could implement a CI check to build and load the policy, to quicky smoke test it.

```console
macano@macano-thinkpadt14gen5:~/Desktop/lv_vagrant/share/repos/rhc/selinux$ sudo semodule -i rhc.pp
Failed to resolve filecon statement at /var/lib/selinux/targeted/tmp/modules/400/rhc/cil:75
Failed to resolve AST
semodule:  Failed!
```

zpetrace:
@mjcr99 `semodule -i` on GitHub Actions is a poor fit. rhc PR CI is ubuntu-latest plus Fedora containers. Those jobs have no SELinux kernel (/sys/fs/selinux is not there), so semodule cannot do the check that actually failed. A nested VM would work but IMO is not worth the cost now for a skeleton module.

selinux/rhc.if
```diff
@@ -0,0 +1,53 @@
...
+##	the oneshot runner confined.
+##	</p>
+## </desc>
+## <param name="domain">
+##	<summary>
+##	Domain allowed to transition.
+##	</summary>
+## </param>
+#
+interface(`rhc_collector_domtrans',`
```
stephenbourke, line 21: Sorry @zpetrace, just one more query from me. Do we need this domain transition API? I'm not sure where it will be used. 

From what I understand.

`init_daemon_domain(rhc_collector_t, rhc_collector_exec_t)` sets up `init_t` -> `rhc_collector_t`

zpetrace: Fair question, we don't call rhc_collector_domtrans() from rhc.te. init_daemon_domain() already wires the init_t → rhc_collector_t transition for the systemd oneshot. The .if export is the public API (CCT-2894 AC + usual refpolicy pattern) for anything outside this module that needs to transition into rhc_collector_t. The interface we actually invoke in .te is rhc_collector_minimal_domtrans(rhc_collector_t) for the child collector handoff. Added a note in the .if desc to make that clearer

selinux/rhc.if
```diff
@@ -0,0 +1,53 @@
+## <summary>SELinux policy for rhc-collector domains</summary>
+
+########################################
+## <summary>
+##	Execute rhc-collector in the rhc_collector_t domain.
+## </summary>
+## <desc>
+##	<p>
+##	Automatic domain transition to rhc_collector_t when executing
```
stephenbourke: I think based on my comment below this interface does not set up the oneshot runner confinement, rather `init_daemon_domain` does.

zpetrace: See my comment above:)

selinux/rhc.if
stephenbourke: Ah sweet. I put a comment on the description there right after. I think what you have now makes sense.

selinux/rhc.if
stephenbourke: Ignore this. I wrote it just as it was updated.

stephenbourke (approved):

mjcr99 (approved):
LGTM, GJ @zpetrace!

Peto-RH (approved):
I couldn’t find anything to improve/change. This change looks good. Nice work!

PR #571, opened 2026-09-11
refactor(connect): move connection business logic to pkg/operations

stephenbourke:
* Card Id: CCT-2409

    - Move RHSM, Lightspeed, yggdrasil connect logic and connect report
      data into pkg/operations
    - Move connect formating operations to a separate util file
    - Move `internal/ui` helper functions to `cmd/rhc`


cmd/rhc/connect_cmd.go
```diff
@@ -3,275 +3,21 @@ package main
...
-	"time"
 
-	"github.com/briandowns/spinner"
-	"github.com/urfave/cli/v3"
-	"golang.org/x/term"
-
-	"github.com/redhatinsights/rhc/internal/datacollection"
-	"github.com/redhatinsights/rhc/internal/remotemanagement"
-	"github.com/redhatinsights/rhc/internal/subman"
 	"github.com/redhatinsights/rhc/internal/ui"
```
mjcr99: Should we also try to refactor this import to avoid it?

cmd/rhc/connect_cmd.go
```diff
@@ -454,40 +193,40 @@ func beforeConnectAction(ctx context.Context, cmd *cli.Command) (context.Context
...
 	if uid != 0 {
 		errMsg := "non-root user cannot connect system"
 		slog.Error(errMsg)
+		report.UID = uid
+		report.UIDError = errMsg
 		if ui.IsOutputMachineReadable() {
-			connectResult.UID = uid
-			connectResult.UIDError = errMsg
-			return cli.Exit(connectResult, exitcode.NoPerm)
+			_ = formatConnectJSON(report)
```
mjcr99: I would keep consistency with the way we are dealing with errors [here](https://github.com/RedHatInsights/rhc/pull/571/changes#diff-5b353d60254729a61eb38c7960b0137f1535d5d3aba7d410e2dda641dc831dbeR296):
 
```go
		if printErr := formatConnectJSON(report); printErr != nil {
			return cli.Exit(
				fmt.Errorf("unable to print connect result as %s document: %s", cmd.String("format"), printErr.Error()),
				exitcode.IOErr)
		}
```

cmd/rhc/connect_cmd.go
```diff
@@ -454,40 +193,40 @@ func beforeConnectAction(ctx context.Context, cmd *cli.Command) (context.Context
...
 
-	// Gather hostname
 	hostname, err := os.Hostname()
 	if err != nil {
 		slog.Error(fmt.Sprintf("Error retrieving system hostname: %v", err))
+		report.HostnameError = err.Error()
 		if ui.IsOutputMachineReadable() {
-			connectResult.HostnameError = err.Error()
-			return cli.Exit(connectResult, exitcode.Err)
+			_ = formatConnectJSON(report)
```
mjcr99: Same as lane 213.

cmd/rhc/connect_cmd.go
```diff
@@ -506,86 +245,172 @@ func connectAction(ctx context.Context, cmd *cli.Command) error {
...
+				report, err = connectWithIdentity(opts, hostname, uid)
+			}
 		}
-	} else {
-		ui.Printf("%s[%v] Remote Management ... Skipped\n", ui.Indent.Medium, ui.Icons.Info)
 	}
 
-	if connectResult.RHSMConnected {
-		ui.Printf("\nSuccessfully connected to Red Hat!\n")
+	if errors.Is(err, operations.ErrOrganizationRequired) {
```
mjcr99, line 289: Why do we duplicate this conditional with the previous one?

cmd/rhc/connect_cmd_fmt.go
```diff
@@ -0,0 +1,159 @@
...
+		RHSMConnected:    report.RHSMConnected,
+		RHSMConnectError: report.RHSMError,
+	}
+	doc.Features.Content = connectJSONFeatureFrom(report.Content)
+	doc.Features.Analytics = connectJSONFeatureFrom(report.Analytics)
+	doc.Features.RemoteManagement = connectJSONFeatureFrom(report.RemoteManagement)
+	return ui.PrintJSON(doc)
+}
+
+func setConnectFeatureStatus(report *operations.ConnectReport) {
```
mjcr99: Since this function calls `operations.FeatureStatus()` which is business logic (querying system state), not formatting, I think it would fit better in `connect_cmd.go` near `connectAction`.

cmd/rhc/connect_cmd_fmt.go
```diff
@@ -0,0 +1,159 @@
...
+func connectJSONFeatureFrom(result operations.FeatureResult) connectJSONFeature {
+	return connectJSONFeature{
+		Enabled:    result.Enabled,
+		Successful: result.Successful,
+		Error:      result.Error,
+		Skipped:    result.Skipped,
+	}
+}
+
+func connectErrorMessages(report operations.ConnectReport) map[string]string {
```
mjcr99: I have the same feeling about this function; it does not really format anything. For me, it should be closer to the action logic.

mjcr99 (changes requested):
Thanks @stephenbourke I have some comments.

cmd/rhc/connect_cmd.go
stephenbourke: The block above this is to handle the non-machine readable format case where we have an org error, so we get the prompt to enter the org.

The case here is the fall through for that where we have the org error but we are in machine machine readable format but also to also handle the case where we get an org error again on Connect (line 273).

cmd/rhc/connect_cmd_fmt.go
stephenbourke: Yeah. I was debating that too. I'm kind of thinking this formatting specific file might not be necessary as there is there may not be enough formatting specific logic to warrant it. Might leave it for now, as the disconnect refactor ticket (CCT-2410) might be able to expand it.

cmd/rhc/connect_cmd.go
stephenbourke: I think it's a good idea. I'll consolidate all the ui calls (internal/ui import) into the connect_cmd_fmt.go file.

mjcr99: I think we could try to directly move the `ui.go` file into `rhc/cmd/rhc`. For me, it does not really make sense to have it in the `internal/` folder; importing it breaks the layered code organization, and it's used in many `rhc/cmd/rhc` pending files.
What do you think @stephenbourke ?

pkg/operations/connect_test.go
```diff
@@ -0,0 +1,454 @@
...
+	opts := ConnectOptions{
+		OnStep: func(step ConnectStep, fn func() error) error {
+			onSteps = append(onSteps, step)
+			return fn()
+		},
+		AfterStep: func(step ConnectStep, _ ConnectReport) {
+			afterSteps = append(afterSteps, step)
+		},
+	}
+	return opts, &onSteps, &afterSteps
```
Peto-RH: Would it make sense to remove `OnStep` and `AfterStep` from `ConnectOptions` and let `operations.Connect` run each step directly? The CLI could wrap the entire call in a single spinner and format the returned report, while the existing private fake dependencies would keep the business flow testable.

I ran into the same spinner complications when refactoring the `status` subcommand and discussed this with @m-horky. He confirmed that we can drop the individual spinner wrappers to simplify the code, since per-step progress indicators are not a business priority. Using one overall spinner would remove the need for these per-step callbacks.

cmd/rhc/connect_cmd_fmt_test.go
```diff
@@ -0,0 +1,243 @@
...
+	if got != want {
+		t.Errorf("output = %q, want %q", got, want)
+	}
+}
+
+func TestConnectErrorMessages(t *testing.T) {
+	report := operations.ConnectReport{
+		RHSMError: "rhsm failed",
+		Analytics: operations.FeatureResult{Error: "insights failed"},
+		RemoteManagement: operations.FeatureResult{
```
Peto-RH: Could we rename `FeatureResult` to `ConnectFeatureResult` and document the exported connect types? The current name is broad and easy to confuse with the existing enable, disable, and status result types. What do you think?

cmd/rhc/connect_cmd.go
```diff
@@ -454,138 +191,237 @@ func beforeConnectAction(ctx context.Context, cmd *cli.Command) (context.Context
...
+	return opts, nil
+}
+
+func failOrganizationRequired(report *operations.ConnectReport, msg string) {
+	report.RHSMError = msg
+	slog.Error(report.RHSMError)
+	formatConnectStepsReport(*report)
+}
+
+func promptOrganization(username, password string) (string, error) {
```
Peto-RH, line 394: Could we add the cannot retrieve organizations context only around GetOrganizations and pass prompt-input errors through unchanged? Right now EOF, read failures, and an empty selection are all described as organization-retrieval failures. What do you think?

Peto-RH (commented):
Hi @stephenbourke, excellent work!

I left a few comments below. Since each comment concerns several lines of code, I also created a branch with AI assistance and pushed the suggested changes there to make it clearer what I meant.

Branch: https://github.com/RedHatInsights/rhc/tree/pschrimp/branch-to-be-deleted

There are 4 additional commits

cmd/rhc/connect_cmd.go
stephenbourke: Yeah i think you're right. Based on the architecture guidelines we shouldn't be importing `/internal` anywhere in `cmd/rhc`

"Presentation code must not import from `internal/`, it should always go through `pkg/`."

However `pkg` does not really make sense either as that is largely geared towards business logic and these are just presentation helpers.

cmd/rhc/connect_cmd_fmt_test.go
stephenbourke: That's a good point `operations.FeatureResult` is very ambiguous. I'll update this and document the exported types.

cmd/rhc/connect_cmd.go
stephenbourke: I think that makes perfect sense. I'll update this.

pkg/operations/connect_test.go
stephenbourke: Okay that's good to know. I assumed there was some real time requirement. That definitely simplifies the logic! I'll make those updates based off the commit on your branch.

Peto-RH (dismissed):
LGTM!

mjcr99 (dismissed):
Thanks @stephenbourke! It LGTM.

cmd/rhc/connect_cmd.go
pkoprda, line 1: Some methods in this file doesn't have any docstring, can you please add it?

cmd/rhc/connect_cmd.go
```diff
@@ -454,138 +191,242 @@ func beforeConnectAction(ctx context.Context, cmd *cli.Command) (context.Context
...
+	if report.RemoteManagement.Error != "" && !report.RemoteManagement.Skipped {
+		errorMessages["yggdrasil"] = report.RemoteManagement.Error
+	}
+	return errorMessages
+}
+
+// connectAction is the CLI handler for 'rhc connect'.
+// It checks identity, collects credentials, runs operations.Connect
+// with a spinner, handles an organization prompt
+// if RHSM requires one, and prints the human readable or JSON result.
```
pkoprda: Nit: I think that this is spelled with a hyphen
```suggestion
// if RHSM requires one, and prints the human-readable or JSON result.
```

pkoprda (commented):
I have two notes, just about documentation, nothing blocking.

Can you also please squash the commits into one? Or move it out from this PR? Because the squashed commit may not look great.

cmd/rhc/connect_cmd.go
stephenbourke: No problem. Added in latest push.

pkoprda (dismissed):
LGTM, thanks!

pkoprda (approved):

PR #576, opened 2026-09-15
docs: add com.redhat.rhc(7) man page for the Varlink API

mjcr99:
* Card ID: CCT-1758

Documents the com.redhat.rhc.collector Varlink interface served by rhc-server on the /run/rhc/com.redhat.rhc Unix socket. It Covers types, methods, errors, socket activation, and example varlinkctl invocations.

Assisted-by: Claude Code


doc/source/markdown/com.redhat.rhc.7.md
```diff
@@ -0,0 +1,176 @@
...
+
+# SYNOPSIS
+
+```
+varlinkctl call unix:/run/rhc/com.redhat.rhc INTERFACE.METHOD '{ARGUMENTS}'
+```
+
+# DESCRIPTION
+
+**rhc-server** exposes a set of Varlink interfaces on a Unix domain socket at **/run/rhc/com.redhat.rhc**. These interfaces allow local programs to list data collectors without requiring direct access to the **rhc** CLI.
```
stephenbourke: We can list and gather info on the registered collectors. Maybe change the sentence to just "query data collectors".

```suggestion
**rhc-server** exposes a set of Varlink interfaces on a Unix domain socket at **/run/rhc/com.redhat.rhc**. These interfaces allow local programs to query data collectors without requiring direct access to the **rhc** CLI.
```

stephenbourke: Or maybe we could be a bit more general and not just single out data collectors. This API interface is likely to expand further in time. For example when the "testing" suffix is dropped from the other APIs. 

"These interfaces allow local programs to interact with rhc-server without requiring the rhc CLI. For example to query data collectors."

doc/source/markdown/com.redhat.rhc.7.md
```diff
@@ -0,0 +1,176 @@
...
+$ systemctl status rhc-server.socket
+```
+
+To check if the service is running:
+
+```
+$ systemctl status rhc-server.service
+```
+
+If the socket is disabled, Varlink calls and **rhc** CLI commands that depend on the service will fail with an error suggesting to restart the socket:
```
stephenbourke: I think only the `rhc` CLI outputs an error suggesting to restart?

doc/source/markdown/com.redhat.rhc.7.md
```diff
@@ -0,0 +1,176 @@
...
+        "id" : "com.redhat.minimal",
+        "name" : "Minimal Host Inventory Collector",
+        "service_name" : "rhc-collector-com.redhat.minimal.service",
+        "timer_name" : "rhc-collector-com.redhat.minimal.timer"
+    }
+}
+```
+
+## Errors
+
```
stephenbourke, line 130: Should we tie the errors specifically to the method calls that return them? I think these two only apply to Info().

doc/source/markdown/com.redhat.rhc.7.md
```diff
@@ -0,0 +1,176 @@
...
+$ systemctl status rhc-server.service
+```
+
+If the socket is disabled, Varlink calls and **rhc** CLI commands that depend on the service will fail with an error suggesting to restart the socket:
+
+```
+# systemctl restart rhc-server.socket
+```
+
+# FILES
```
stephenbourke, line 163: Might also be worth mentioning the systemd unit files for `rhc-server`.

```
ls /usr/lib/systemd/system/rhc-server.s*
/usr/lib/systemd/system/rhc-server.service  
/usr/lib/systemd/system/rhc-server.socket
```

doc/source/markdown/com.redhat.rhc.7.md
```diff
@@ -0,0 +1,182 @@
...
+# SYNOPSIS
+
+```
+varlinkctl call unix:/run/rhc/com.redhat.rhc INTERFACE.METHOD '{ARGUMENTS}'
+```
+
+# DESCRIPTION
+
+**rhc-server** exposes a set of Varlink interfaces on a Unix domain socket at **/run/rhc/com.redhat.rhc**. These interfaces allow local programs to query data collectors without requiring direct access to the **rhc** CLI.
+
```
stephenbourke, line 16: Maybe we should mention that that the socket is owned by root (0600) and only root can connect?

doc/source/markdown/com.redhat.rhc.7.md
```diff
@@ -0,0 +1,182 @@
...
+The service is managed by systemd. The socket unit **rhc-server.socket** supports socket activation: the first Varlink call to the socket starts **rhc-server.service** automatically if it is not already running.
```
stephenbourke: I think this should say the "first Varlink call from the CLI"?

stephenbourke (approved):
LGTM

PR #579, opened 2026-09-17
refactor: move disconnect logic to operations

Peto-RH:
* CardID: CCT-2410

Extract the disconnect workflow into pkg/operations and return a data-only report. Keep formatting and exit codes in the CLI.

Assisted-by: claude-opus-5


cmd/rhc/disconnect_cmd.go
```diff
@@ -182,72 +23,80 @@ func beforeDisconnectAction(ctx context.Context, cmd *cli.Command) (context.Cont
...
+				fmt.Errorf("unable to print disconnect as json document: %s", err.Error()),
+				exitcode.IOErr)
+		}
+		// Preserve the successful JSON exit code even when individual steps fail.
+		return nil
+	}
 
-	if !isOutputMachineReadable() {
-		showTimeDuration(durations)
+	printf("Disconnecting %v from Red Hat.\nThis might take a few seconds.\n\n", report.Hostname)
```
stephenbourke: I think the disconnect execution has already run been this point and this function is the report afterwards. I don't think this line is accurate. Previously the printout occurred before disconnect. Maybe rephrase this print out to say "host <hostname> disconnected" or maybe some means to print this during the disconnect execution.

pkg/operations/disconnect.go
```diff
@@ -0,0 +1,180 @@
...
+		InsightsClientIsRegistered:  datacollection.InsightsClientIsRegistered,
+		UnregisterInsightsClient:    datacollection.UnregisterInsightsClient,
+		NewRHSMClient: func() (disconnectRHSMClient, error) {
+			return subman.NewRHSMClient()
+		},
+	}
+}
+
+// DisconnectReport describes the result of disconnecting the system from Red Hat.
+type DisconnectReport struct {
```
stephenbourke: Just a suggestion. It might make sense to remove the tags from this struct and use a separate one to marshal the report to (CLI only)? Similar to how `operations.ConnectReport` and `connectJSONDocument` do it. This would separate out presentation data from the in process report data.

Disconnect’s JSON is a closer one to one than connect's is though, so I understand if you want to keep it this way. But a separate CLI struct could a useful pattern to use here also if the report grew and we don't want internal only fields to be JSON be default.

Peto-RH: I think this makes absolute sense. Will do :)

cmd/rhc/disconnect_cmd.go
```diff
@@ -182,72 +53,80 @@ func beforeDisconnectAction(ctx context.Context, cmd *cli.Command) (context.Cont
...
+				fmt.Errorf("unable to print disconnect as json document: %s", err.Error()),
+				exitcode.IOErr)
+		}
+		// Preserve the successful JSON exit code even when individual steps fail.
+		return nil
+	}
 
-	if !isOutputMachineReadable() {
-		showTimeDuration(durations)
+	printf("Disconnected %v from Red Hat.\n\n", report.Hostname)
```
stephenbourke: Sorry my last comment was probably a bit too simplistic with how this print should be executed. It can only realistically be printed when each step succeeds. We don't want to have to have it report a disconnected state when it didn't succeed in disconnecting.

pkoprda: +1, I don't think we should change the output message. Be aware of CCT-2410 acceptance criteria: "External behavior has not changed (same output, same exit codes)".

pkg/operations/disconnect.go
```diff
@@ -0,0 +1,180 @@
...
+	return report, nil
+}
+
+func deactivateRemoteManagement(report *DisconnectReport, deps disconnectDependencies) {
+	slog.Info("Deactivating the yggdrasil service")
+
+	isInactive, err := deps.AssertYggdrasilServiceState("inactive")
+	if err != nil {
+		return
+	}
```
pkoprda: Shouldn't we at least log the error? I think `slog.Debug()` would be sufficient here, the error is not really user-oriented

pkg/operations/disconnect.go
```diff
@@ -0,0 +1,180 @@
...
+	slog.Info("Deactivated the yggdrasil service")
+}
+
+func disconnectInsights(report *DisconnectReport, deps disconnectDependencies) {
+	slog.Info("Disconnecting from Red Hat Lightspeed")
+
+	isRegistered, err := deps.InsightsClientIsRegistered()
+	if err != nil {
+		return
+	}
```
pkoprda: Ditto here

pkg/operations/disconnect.go
```diff
@@ -0,0 +1,180 @@
...
+	slog.Info("Unregistering system from Red Hat Subscription Management")
+
+	client, err := deps.NewRHSMClient()
+	if err != nil {
+		return
+	}
+	isRegistered, err := client.IsRegistered()
+	if err != nil {
+		return
+	}
```
pkoprda: Ditto here

pkg/operations/disconnect.go
```diff
@@ -0,0 +1,180 @@
...
+func Disconnect() (*DisconnectReport, error) {
+	return disconnect(defaultDisconnectDependencies())
+}
+
+func disconnect(deps disconnectDependencies) (*DisconnectReport, error) {
+	report := &DisconnectReport{UID: deps.UID()}
+	if report.UID != 0 {
+		report.UIDError = "non-root user cannot disconnect system"
+		slog.Error(report.UIDError)
+		return report, fmt.Errorf("%s", report.UIDError)
```
pkoprda: `fmt.Errorf()` is not really useful here, I would like to see here `errors.New(report.UIDError)`

pkoprda (commented):
Thanks for the PR, I have a few requests.

Also please add `Assisted-by: <AI model>` to the commit description if you used some AI model during creation of this PR.

cmd/rhc/disconnect_cmd.go
```diff
@@ -182,72 +53,80 @@ func beforeDisconnectAction(ctx context.Context, cmd *cli.Command) (context.Cont
...
-// and finally it unregisters system from Red Hat Subscription Management
-func disconnectAction(ctx context.Context, cmd *cli.Command) error {
+// disconnectAction disconnects the system and formats the resulting report.
+func disconnectAction(_ context.Context, cmd *cli.Command) error {
 	logCommandStart(cmd)
 
-	var disconnectResult DisconnectResult
-	disconnectResult.format = cmd.String("format")
+	var report *operations.DisconnectReport
+	err := withSpinner(func() error {
```
stephenbourke: There seems to be a few consequences from moving uid and hostname out of the CLI code. The connectAction code didn't move these which may not be ideal layering, as these elements do lean more towards `pkg/operation` logic. 

https://github.com/RedHatInsights/rhc/blob/main/cmd/rhc/connect_cmd.go#L241

However by not having CLI checks there seems to be two issues:

1. We can't keep the original header hostname printout before running disconnect.

```
Disconnecting {hostname} from Red Hat.
```

2.  We will present a spinner for non-root users (who can't disconnect a host). That would be a bit confusing and breaks regression.  

So maybe if we read the UID/Hostname and validate the returns (same as connect), pass the values for each through to Disconnect, but still maintaining the guards it already has.

What do you think?

stephenbourke (dismissed):
LGTM

pkoprda (approved):
LGTM

PR #580, opened 2026-09-18
feat(selinux): package in-tree policy as rhc-selinux subpackage

Archana-PandeyM:
Compile selinux/rhc.{te,fc,if} into a policy module in %build via selinux-policy-devel's Makefile and ship it as a new rhc-selinux subpackage.
- %package selinux carries the usual selinux-policy floor (42.1.8,
  matching CCT-2850 / insights-core.spec on RHEL 10+) and
  selinux_modules_install/uninstall scriptlets mirrored from
  insights-core.spec.
- Main rhc package gets a weak/conditional Requires so rhc-selinux
  is only pulled in when selinux-policy-targeted is installed.
- %files selinux ghosts the active-module path so rpm -V doesn't
  flag the semodule-managed policy store entry.
The existing rhcd_t permissive shim in %post/%postun is untouched.

* Card ID: CCT-2895


Archana-PandeyM:
Sharing some verification evidence for failed test for collector(**its failing because we haven't granted any allow rules yet**) - 

```
[root@1fdb1db0-68f3-45f5-b026-4cfd29b19721 ~]# rhc collector enable --now com.redhat.minimal
failed to start service rhc-collector-com.redhat.minimal.service: failed to start unit with reason: failed
[root@1fdb1db0-68f3-45f5-b026-4cfd29b19721 ~]# 
[root@1fdb1db0-68f3-45f5-b026-4cfd29b19721 ~]# ausearch -m AVC -ts recent
----
time->Fri Sep 18 11:04:49 2026
type=PROCTITLE msg=audit(1789729489.556:932): proctitle=2F7573722F6C6962657865632F7268632F7268632D636F6C6C6563746F720072756E00636F6D2E7265646861742E6D696E696D616C
type=SYSCALL msg=audit(1789729489.556:932): arch=c000003e syscall=257 success=no exit=-13 a0=ffffff9c a1=556fa8bd70c0 a2=0 a3=0 items=0 ppid=1 pid=10609 auid=4294967295 uid=0 gid=0 euid=0 suid=0 fsuid=0 egid=0 sgid=0 fsgid=0 tty=(none) ses=4294967295 comm="rhc-collector" exe="/usr/libexec/rhc/rhc-collector" subj=system_u:system_r:rhc_collector_t:s0 key=(null)
type=AVC msg=audit(1789729489.556:932): avc:  denied  { read } for  pid=10609 comm="rhc-collector" name="hpage_pmd_size" dev="sysfs" ino=1205 scontext=system_u:system_r:rhc_collector_t:s0 tcontext=system_u:object_r:sysfs_t:s0 tclass=file permissive=0
----
time->Fri Sep 18 11:04:49 2026
type=PROCTITLE msg=audit(1789729489.557:933): proctitle=2F7573722F6C6962657865632F7268632F7268632D636F6C6C6563746F720072756E00636F6D2E7265646861742E6D696E696D616C
type=SYSCALL msg=audit(1789729489.557:933): arch=c000003e syscall=257 success=no exit=-13 a0=ffffffffffffff9c a1=556fa8c4fc40 a2=80000 a3=0 items=0 ppid=1 pid=10609 auid=4294967295 uid=0 gid=0 euid=0 suid=0 fsuid=0 egid=0 sgid=0 fsgid=0 tty=(none) ses=4294967295 comm="rhc-collector" exe="/usr/libexec/rhc/rhc-collector" subj=system_u:system_r:rhc_collector_t:s0 key=(null)
type=AVC msg=audit(1789729489.557:933): avc:  denied  { search } for  pid=10609 comm="rhc-collector" name="/" dev="cgroup2" ino=1 scontext=system_u:system_r:rhc_collector_t:s0 tcontext=system_u:object_r:cgroup_t:s0 tclass=dir permissive=0
----
time->Fri Sep 18 11:04:49 2026
type=PROCTITLE msg=audit(1789729489.558:934): proctitle=2F7573722F6C6962657865632F7268632F7268632D636F6C6C6563746F720072756E00636F6D2E7265646861742E6D696E696D616C
type=SYSCALL msg=audit(1789729489.558:934): arch=c000003e syscall=257 success=no exit=-13 a0=ffffffffffffff9c a1=12af8cbd65a0 a2=80000 a3=0 items=0 ppid=1 pid=10609 auid=4294967295 uid=0 gid=0 euid=0 suid=0 fsuid=0 egid=0 sgid=0 fsgid=0 tty=(none) ses=4294967295 comm="rhc-collector" exe="/usr/libexec/rhc/rhc-collector" subj=system_u:system_r:rhc_collector_t:s0 key=(null)
type=AVC msg=audit(1789729489.558:934): avc:  denied  { search } for  pid=10609 comm="rhc-collector" name="collectors" dev="nvme0n1p3" ino=109054212 scontext=system_u:system_r:rhc_collector_t:s0 tcontext=system_u:object_r:rhc_collector_conf_t:s0 tclass=dir permissive=0
[root@1fdb1db0-68f3-45f5-b026-4cfd29b19721 ~]# 
```

rhc.spec
```diff
@@ -50,11 +57,30 @@ Requires: yggdrasil >= 0.4
...
 Requires: yggdrasil-worker-package-manager
 %endif
+# Pull rhc-selinux when the targeted SELinux policy is installed.
+Requires: ((%{name}-selinux >= %{version}-%{release}) if selinux-policy-%{selinuxtype})
 
 %description
 Client tool to register Fedora, CentOS Stream or Red Hat Enterprise Linux
 to Red Hat Subscription Management and Red Hat Lightspeed.
 
+%package selinux
```
m-horky, line 65: Should this be `rhc-selinux` instead?

rhc.spec
```diff
@@ -193,6 +227,19 @@ if [ $1 -eq 0 ]; then
...
+
+%post selinux
+%selinux_modules_install -s %{selinuxtype} %{_datadir}/selinux/packages/%{selinuxtype}/%{selinux_modulename}.pp.bz2
+%selinux_relabel_post -s %{selinuxtype}
+
+%postun selinux
+if [ $1 -eq 0 ]; then
+	%selinux_modules_uninstall -s %{selinuxtype} %{selinux_modulename}
+	%selinux_relabel_post -s %{selinuxtype}
+fi
```
m-horky, line 239: How does this handle bootc use-cases, where this code never runs? Solving it as such feels fragile.

rhc.spec
```diff
@@ -7,6 +7,12 @@
 %global with_rhcd_compat 1
 %endif
 
+# SELinux policy module shipped as the rhc-selinux subpackage.
+# Built from in-tree selinux/; not a Source tarball.
```
m-horky: Nit: This feels like a note required during the work on this patch, rather than something that would be useful to later readers.

rhc.spec
Archana-PandeyM: from my understanding so far %package selinux is correct, not %package rhc-selinux.
In RPM spec syntax: %package selinux automatically creates: Name: rhc-selinux because the main package is rhc

rhc.spec
```diff
@@ -50,11 +57,30 @@ Requires: yggdrasil >= 0.4
...
 %endif
+# Pull rhc-selinux when the targeted SELinux policy is installed.
+Requires: ((%{name}-selinux >= %{version}-%{release}) if selinux-policy-%{selinuxtype})
 
 %description
 Client tool to register Fedora, CentOS Stream or Red Hat Enterprise Linux
 to Red Hat Subscription Management and Red Hat Lightspeed.
 
+%package selinux
+Summary:        SELinux policy for rhc-collector
```
m-horky, line 66: For `rhc-collector` or for `rhc`, of which rhc-collector is part of?

Archana-PandeyM: For rhc-collector, which is shipped as part of the rhc package. The policy itself is scoped to the collector domains (rhc_collector_t and rhc_collector_minimal_t), not the rhc client as a whole.

Are you suggesting to reword it?

rhc.spec
m-horky: TIL!

rhc.spec
m-horky: In the future the `rhc-selinux` package will contain policy for everything, not just for the `rhc-collector` binary, so I thought we could be more generic. If we described the subpackage as "a policy for rhc", we wouldn't have to deal with updating it (and tracking differences against downstream .spec files). But in the current state it is actually more correct summary. I'll leave it up to you, feel free to close this thread without changing anything.

rhc.spec
Archana-PandeyM: Tested this end-to-end on a real bootc image today (centos-bootc:stream10, layered via podman build + dnf install, deployed via bcvk libvirt run .  here's what I observed:

At dnf install rhc-selinux inside the Containerfile build: %post  does execute (I hit the known libsemanage/overlayfs EXDEV warning during the build, which is a documented, non-fatal fallback to a copy-based commit: [SELinuxProject/selinux#343](https://github.com/SELinuxProject/selinux/issues/343)). It's only the selinuxenabled-gated bits - the fixfiles-diff relabel in %pre/%post and the explicit load_policy call  that no-op during the build, because there's no live target-matching SELinux state in a container build root. That's standard for every SELinux-policy-shipping RPM built this way, not specific to us is what I understood so far.

At real boot, the journal shows systemd itself does the actual policy load unconditionally on every boot (Successfully loaded SELinux policy in ), independent of our scriptlets. After that:

`semodule -l `shows rhc registered.
The module store on disk (/var/lib/selinux/targeted/active/modules/200/rhc/) has real compiled content (cil/hll), correctly labeled semanage_store_t.
Running rhc collector enable --now actually transitions into rhc_collector_t and gets real AVC denials (permissive=0) against rhc_collector_t/rhc_collector_conf_t. That's proof the type and module are genuinely loaded and enforced by the kernel. 

Do you have any suggestions on what else I could look for?

rhc.spec
Archana-PandeyM: I think we should let it be, because changing it to rhc risks implying the whole package is confined (CLI, rhc-server, etc.), which isn't true today.
If confinement ever expands to rhc-server or the CLI itself(which eventually will), we'll revisit then because that's a big enough change to warrant a fresh look at the wording anyway.

m-horky: Okay, sounds good to me!

rhc.spec
m-horky: Thanks for the validation, this is enough. Resolved!

pkoprda (approved):
LGTM

PR #581, opened 2026-09-21
feat(selinux): add least-privilege allow rules for rhc-collector domains

mjcr99:
## Description

Populate the rhc.te skeleton with allow rules so rhc_collector_t,
rhc_collector_minimal_t, and the new rhc_collector_plugin_t can run in
SELinux enforcing mode.

rhc_collector_t (systemd oneshot runner):
- fork/exec + setuid/setgid for launching child collectors and tar
- read collector configs (rhc_collector_conf_t)
- manage timer caches (rhc_cache_t) and temp dirs (rhc_tmp_t)
- exec /usr/bin/tar via corecmd_exec_bin (execute_no_trans, no domtrans)
- read mTLS certs and outbound HTTPS to console.redhat.com / Satellite
- Go runtime: sysfs (THP), /dev/urandom, NSS, locale, cgroup, journal

rhc_collector_minimal_t (com.redhat.minimal child collector):
- inherit FDs/pipes from parent, manage working dir (rhc_tmp_t)
- read /etc/machine-id, insights-client machine-id (optional_policy for
  insights_client_etc_t/insights_client_etc_rw_t), consumer cert, BIOS
  UUID from sysfs, network interfaces via netlink
- Go runtime: same as above plus kernel_read_network_state for iface enum

rhc_collector_plugin_t (third-party collector fallback):
- label /usr/libexec/rhc/collectors/.* as rhc_collector_plugin_exec_t;
  longest-stem matching preserves the com.redhat.minimal override
- domtrans from rhc_collector_t enforces confinement automatically
- grants only: working directory (rhc_tmp_t) + Go runtime basics
- deliberately omits: corenet, certs, setuid/setgid, DNS — enforces
  the "collectors must not do network" contract from rhc-collector(5)
- collectors needing more access should ship their own domain

Domain transition plumbing:
- siginh/rlimitinh/noatsecure for both minimal and plugin transitions
  (hidden by base policy dontaudit, required for clean exec handoff)

tmpfiles.d/rhc.conf:
- add /var/cache/rhc and /var/cache/rhc/collectors so the cache path
  exists at boot; avoids granting write on var_t from the policy

## Testing
To test the policy, it's required to manually compile it:
```console
[vagrant@lv-rhel101-go-lab selinux]$ pwd
/vagrant/repos/rhc/selinux

[vagrant@lv-rhel101-go-lab selinux]$ make -f /usr/share/selinux/devel/Makefile rhc.pp
```

Then we apply the policy:
```console
[vagrant@lv-rhel101-go-lab selinux]$ sudo semodule -i rhc.pp
```

A quick first view to check if it has been properly applied is to check if the new types has been added:
```console
[vagrant@lv-rhel101-go-lab selinux]$ seinfo -t | grep rhc_
   rhc_cache_t
   rhc_collector_conf_t
   rhc_collector_exec_t
   rhc_collector_minimal_exec_t
   rhc_collector_minimal_t
   rhc_collector_t
   rhc_tmp_t

[vagrant@lv-rhel101-go-lab selinux]$ getenforce 
Enforcing
```

Then we can manually trigger the minimal collector to look for denials and check if everything runs smoothly:
```console
[vagrant@lv-rhel101-go-lab selinux]$ sudo systemctl start rhc-collector-com.redhat.minimal.service
[vagrant@lv-rhel101-go-lab selinux]$ sudo systemctl status rhc-collector-com.redhat.minimal.service
○ rhc-collector-com.redhat.minimal.service - Minimal Host Inventory data collector
     Loaded: loaded (/usr/lib/systemd/system/rhc-collector-com.redhat.minimal.service; disabled; preset: disabled)
     Active: inactive (dead)
       Docs: https://github.com/RedHatInsights/rhc

Sep 21 11:11:11 lv-rhel101-go-lab rhc-collector[70668]: 2026/09/21 11:11:11 INFO starting rhc-collector id=com.redhat.minimal
Sep 21 11:11:11 lv-rhel101-go-lab rhc-collector[70668]: 2026/09/21 11:11:11 INFO configuration of the collector config="{ID:com.redhat.minimal Name:Minimal Host Inventory Collector IsAnalyticsFeature:true User:root Group:root ContentType:application/vnd.redhat.advisor.minimal}"
Sep 21 11:11:11 lv-rhel101-go-lab rhc-collector[70668]: 2026/09/21 11:11:11 INFO created temporary directory dir=/var/tmp/rhc/collector-1780751460
Sep 21 11:11:11 lv-rhel101-go-lab rhc-collector[70668]: 2026/09/21 11:11:11 INFO executing collector as configured user/group collector=com.redhat.minimal user=root group=root
Sep 21 11:11:11 lv-rhel101-go-lab rhc-collector[70668]: 2026/09/21 11:11:11 INFO collector has ran successfully output="2026/09/21 11:11:11 INFO starting minimal collector outputDir=/var/tmp/rhc/collector-1780751460\n2026/09/21 11:11:11 INFO minimal collector completed successfull>
Sep 21 11:11:11 lv-rhel101-go-lab rhc-collector[70668]: 2026/09/21 11:11:11 INFO archive created path=/var/tmp/rhc/rhc-collector-20260921111111125.tar.xz
Sep 21 11:11:11 lv-rhel101-go-lab rhc-collector[70668]: 2026/09/21 11:11:11 INFO Uploading archive archive=/var/tmp/rhc/rhc-collector-20260921111111125.tar.xz url=https://cert.console.redhat.com/api/ingress/v1/upload
Sep 21 11:11:11 lv-rhel101-go-lab rhc-collector[70668]: 2026/09/21 11:11:11 INFO Successfully uploaded archive archive=/var/tmp/rhc/rhc-collector-20260921111111125.tar.xz
Sep 21 11:11:11 lv-rhel101-go-lab systemd[1]: rhc-collector-com.redhat.minimal.service: Deactivated successfully.
Sep 21 11:11:11 lv-rhel101-go-lab systemd[1]: Finished rhc-collector-com.redhat.minimal.service - Minimal Host Inventory data collector.
``` 

And check for AVC denials:
```console
[vagrant@lv-rhel101-go-lab ~]$ sudo ausearch -m AVC -ts recent
<no matches>
```

So it seems to work properly.

We can also check the domain transition behavior, to do so, the best way to track it is to instruct audit to watch for system calls as follows:

```console
[vagrant@lv-rhel101-go-lab ~]$ sudo auditctl -a always,exit -F arch=b64 -S execve -k trace_exec
```

This basically adds an audit rule that logs every 64-bit execve system call, i.e., whenever a process launches a new binary. The `trace_exec` key tags these events, making them easier to filter.

The we can launch this command in another terminal to filter in realtime for the generated logs when `sudo systemctl start rhc-collector-com.redhat.minimal.service` is executed. 

```console
[vagrant@lv-rhel101-go-lab ~]$ sudo tail -f /var/log/audit/audit.log | awk '
/type=SYSCALL.*key="trace_exec"/ {
    pid = ""; subj = ""; exe = "";
    
    for (i = 1; i <= NF; i++) {
        if ($i ~ /^pid=/) {
            split($i, a, "="); pid = a[2];
        }
        if ($i ~ /^subj=/) {
            split($i, a, "="); subj = a[2];
        }
        if ($i ~ /^exe=/) {
            split($i, a, "="); exe = a[2]; gsub(/"/, "", exe);
        }
    }
    
    if (pid != "" && exe != "") {
        printf "[%s] PID: %-6s | Executing: %-45s | Label: %s\n", strftime("%H:%M:%S"), pid, exe, subj;
        fflush();
    }
}'
```

This command continuously monitors `/var/log/audit/audit.log` for `execve` events tagged with `trace_exec` and extracts the process `PID`, executable path (`exe`), and SELinux context (`subj`). It then displays them in a formatted, real-time output with timestamps.

This generates the following output:
```console
[11:23:48] PID: 70724  | Executing: /usr/bin/tail                                 | Label: unconfined_u:unconfined_r:unconfined_t:s0-s0:c0.c1023
[11:23:52] PID: 70725  | Executing: /usr/bin/sudo                                 | Label: unconfined_u:unconfined_r:unconfined_t:s0-s0:c0.c1023
[11:23:52] PID: 70726  | Executing: /usr/sbin/unix_chkpwd                         | Label: unconfined_u:unconfined_r:chkpwd_t:s0-s0:c0.c1023
[11:23:52] PID: 70728  | Executing: /usr/bin/systemctl                            | Label: unconfined_u:unconfined_r:unconfined_t:s0-s0:c0.c1023
[11:23:52] PID: 70729  | Executing: /usr/bin/systemd-tty-ask-password-agent       | Label: unconfined_u:unconfined_r:unconfined_t:s0-s0:c0.c1023
[11:23:52] PID: 70730  | Executing: /usr/lib/systemd/systemd-executor             | Label: system_u:system_r:init_t:s0
[11:23:52] PID: 70730  | Executing: /usr/libexec/rhc/rhc-collector                | Label: system_u:system_r:rhc_collector_t:s0
[11:23:52] PID: 70736  | Executing: /usr/libexec/rhc/collectors/com.redhat.minimal | Label: system_u:system_r:rhc_collector_minimal_t:s0
[11:23:52] PID: 70741  | Executing: /usr/bin/tar                                  | Label: system_u:system_r:rhc_collector_t:s0
[11:23:52] PID: 70742  | Executing: /usr/bin/bash                                 | Label: system_u:system_r:rhc_collector_t:s0
[11:23:52] PID: 70742  | Executing: /usr/bin/xz                                   | Label: system_u:system_r:rhc_collector_t:s0
```

It's observed that the minimal collector is executed under its own tag, which indicates that the domain transition works as expected.


selinux/rhc.te
```diff
@@ -40,8 +37,149 @@ files_tmp_file(rhc_tmp_t)
...
+allow rhc_collector_t self:fifo_file rw_fifo_file_perms;
+
+# --- Collector configuration: /usr/lib/rhc/collectors/*.toml ---------------
+# GetConfig -> loadConfigFromFile -> toml.DecodeFile
+list_dirs_pattern(rhc_collector_t, rhc_collector_conf_t, rhc_collector_conf_t)
+read_files_pattern(rhc_collector_t, rhc_collector_conf_t, rhc_collector_conf_t)
+
+# --- Timer cache: /var/cache/rhc/collectors/*.json -------------------------
+# WriteTimerCache -> writeTimerToFile
+manage_dirs_pattern(rhc_collector_t, rhc_cache_t, rhc_cache_t)
```
stephenbourke: I think we may to run into an AVC denial here? Maybe worth a quick test to see.

The collector directory under `/var/cache` is only created at runtime.

https://github.com/RedHatInsights/rhc/blob/main/internal/collector/collector.go#L187

`manage_dirs_pattern(...)` only applies after the path exists and labels are applied. A file
context won't be applied on `mkdir` AFAIK.

I don't think we want to expand write permissions across `/var`. It's probably better
to leave this file as is and create a tmpfiles (`data/tmpfiles.d/rhc.conf`) entry to resolve this.

zpetrace: Agree with @stephenbourke as this only covers objects already labeled `rhc_cache_t`. `writeTimerToFile` always `MkdirAlls` `/var/cache/rhc/collectors/` ([collector.go](https://github.com/RedHatInsights/rhc/blob/main/internal/collector/collector.go#L187)). That path is not in `data/tmpfiles.d/rhc.conf` today (only `/var/tmp/rhc, /run/rhc`, `/var/log/rhc`). File contexts from rhc.fc are not applied at mkdir, so new dirs stay `var_t` and `manage_*_pattern(..., rhc_cache_t, ...)` never matches.

selinux/rhc.te
```diff
@@ -40,8 +37,149 @@ files_tmp_file(rhc_tmp_t)
...
+manage_files_pattern(rhc_collector_t, rhc_tmp_t, rhc_tmp_t)
+
+# Search /var/tmp (tmp_t) to reach /var/tmp/rhc (rhc_tmp_t).
+# tmpfiles.d/rhc.conf pre-creates /var/tmp/rhc at boot; this grants
+# traversal only, not write to tmp_t.
+files_search_tmp(rhc_collector_t)
+
+# --- Archive creation: tar --create --xz ----------------------------------
+# createArchive -> exec.Command("tar", …)
+corecmd_exec_bin(rhc_collector_t)
```
stephenbourke: I know we are not specifically thinking about third party collectors. But think we might open up a bit of a gap here.

From what I understand this line translates to `can_exec(rhc_collector_t, bin_t)` and uses `execute_no_trans` where a specific domain transition is not defined. The `execute_no_trans` will allow a process in one security domain to execute a file in a different domain without triggering a proper domain transition. For third party collectors (ones not in-tree), this seems like it might pose a bit of a problem as the child process will stay in `rhc_collector_t` (with certs, setuid/setgid, sockets for outbound HTTPS etc).

For the usage of `/usr/bin/tar` specifically, this is okay. For third party collectors, I don't think it is. When running `com.redhat.minimal` it is `rhc_collector_minimal_exec_t` and it has `rhc_collector_minimal_domtrans`. For any collectors installed at` /usr/libexec/rhc/collectors/{ID}`, it is an ordinary `libexec (bin_t)` and the collector execution runs with `execute_no_trans`. This then inherits the runner's domain (stays `rhc_collector_t`). The “collectors must not do network” contract in **rhc-collector(5)** (https://github.com/RedHatInsights/rhc/blob/main/doc/source/markdown/rhc-collector.5.md#executable) is then unenforced.

There are two potential solutions to this:

1) We label `/usr/libexec/rhc/collectors(/.*)?` with an exec type, keep the more specific `com.redhat.minimal` type after the glob and only ever domtrans into a no-network domain. We then never `execute_no_trans` on third party binaries.

 2) We don't fork a `tar` process. Instead we use an in-process archive mechanism provided by the Go library. We could drop `corecmd_exec_bin` from here afterward.

Let me know what you think.

zpetrace: This works for `tar/xz`, but it also lets the domain run anything under `/usr/bin`. Probably fine for now if that's intentional, maybe just a short comment?

Archana-PandeyM (commented):
this test continues to fail - test_collector_enable_now_triggers_immediate_run 
type=AVC msg=audit(1790068136.343:857): avc:  denied  { write } for  pid=7538 comm="rhc-collector" name="cache" dev="nvme0n1p3" ino=8389433 scontext=system_u:system_r:rhc_collector_t:s0 tcontext=system_u:object_r:var_t:s0 tclass=dir permissive=0

zpetrace (changes requested):
Looks good overall, the allows map cleanly to the runner vs minimal-collector paths, and the enforcing + domain-transition testing notes are really helpful, thanks:) I left just 2 comments, so once that tmpfiles piece is in and re-checked, this should be IMO in a good shape.

selinux/rhc.te
stephenbourke: I had a similar more rambling comment above. :)
I think we do need to keep an eye on it for when third party collectors get built.

selinux/rhc.te
mjcr99: Addressed, I've updated the `data/tmpfiles.d/rhc.conf` file.

selinux/rhc.te
mjcr99: You are right, and it's very well spotted. I've added the `rhc_collector_plugin_t` to deal with the rest of the upcoming collectors. It may not be perfect so far because we can not test it yet, but the foundation is there.

Regarding option 2, I think we could also address it; it would reduce the dependency of `rhc` on other packages, but it would belong to another PR, I think.

mjcr99: Addressed with [this](https://github.com/RedHatInsights/rhc/pull/581#discussion_r4070973095) approach.

zpetrace (dismissed):
LGTM after the changes, let's wait for other reviewers though:)

selinux/rhc.te
stephenbourke: Looks good!

Just one more thing that i think is worth noting.

I think we are closing the door on interpreted collector scripts running. For example scripts using `#!/usr/bin/python` shebang.

`rhc_collector_plugin_exec_t` -> `plugin_t` -> `bin_t` (`/usr/bin/python`) -> denied

`plugin_t` has no `corecmd_exec_bin` or `corecmd_exec_shell`.

But we do have a note that interpreted scripts are supported as collectors:

https://github.com/RedHatInsights/rhc/blob/main/doc/COLLECTORS.md#collector-executable

It might be worth adding `corecmd_exec_shell` which will allow `plugin_t` executions to use `execute_no_trans`. It won't be the same problem as before as we it will be confined within `plugin_t` as opposed to `rhc_collector_t` (allowing network access etc).

`corecmd_exec_shell(rhc_collector_plugin_t)`

The only thing is, if a script is more involved it could lead to further AVC denials. Maybe that is not for us to solve.

mjcr99: I think it's worth mentioning in the `.te` file; I agree.

stephenbourke (approved):
Nice work @mjcr99! LGTM

Archana-PandeyM (approved):
Thanks for the changes Manu! LGTM!

